PR #100, opened 2014-12-24
CURL handles not properly initialized to use DNS or SSL session caching.

adobos:
When initializing CURL handles, the is_dns_cache and is_ssl_session_cache flags were not checked properly. As a result, the handles were not set-up properly for DNS or SSL session caching.


ggtakec:
I'm sorry for replying late.
Thanks very much, this is my fault.

Regards,

PR #101, opened 2014-12-24
Optimized function "bool directory_empty()"

adobos:
Function "bool directory_empty()" downloads full directory listing to find out whether the directory is empty or not. This requires both a larger download from S3, and possibly multiple round trips to S3 (S3 API returns at most 1,000 entries per call). For directories with many thousands of entries, this comes with a significant performance penalty.

The modified version of the function checks for at most one entry to make a decision whether the directory is empty or not.


ggtakec:
Thank you.
As you pointed out, it was unnecessary cost, and your codes is very good.

PR #104, opened 2015-01-12
fix rename before close

kahing:
nautilus does this when you drag and drop to overwrite a file:

1) create .goutputstream-XXXXXX to write to
2) fsync the fd for .goutputstream-XXXXXX
3) rename .goutputstream-XXXXXX to target file
4) close the fd for .goutputstream-XXXXXX

previously, doing this on s3fs would result in an empty target file
because after the rename, s3fs would not flush the content of
.goutputstream-XXXXXX to target file.

this change moves the FdEntity from the old path to the new path
whenever rename happens. On flush s3fs would now flush the correct
content to the rename target.


kahing:
@andrewgaul this fixes the issue I was talking to you about

src/fdcache.cpp
```diff
@@ -1177,6 +1177,19 @@ FdEntity* FdManager::Open(const char* path, off_t size, time_t time, bool force_
...
 }
 
+void FdManager::Rename(const std::string from, const std::string to)
+{
+  fdent_map_t::iterator iter = fent.find(from);
+  if(fent.end() != iter){
+    // found
+    FPRNINFO("[from=%s][to=%s]", from.c_str(), to.c_str());
+    FdEntity* ent = (*iter).second;
+    fent.erase(from);
```
andrewgaul: Call `erase(iter)` to avoid duplicated lookup.

src/fdcache.h
```diff
@@ -169,6 +170,7 @@ class FdManager
     FdEntity* GetFdEntity(const char* path);
     FdEntity* Open(const char* path, off_t size = -1, time_t time = -1, bool force_tmpfile = false, bool is_create = true);
     FdEntity* ExistOpen(const char* path) { return Open(path, -1, -1, false, false); }
+    void Rename(const std::string from, const std::string to);
```
andrewgaul: Pass by const reference?

andrewgaul:
:+1:

kahing:
addressed comments

ggtakec:
Hi, all
Thanks very much, and merged.

Issue #107, opened 2015-01-21
chroot catch-22


consultantRR:
I am trying to implement a chroot jail backed by s3fs.

But, chroot requires that the parent directory be owned by root:root and not world writable.

If I use `-o allow_other`, then the permissions on the s3fs mnt directory is root:root 777.

If I don't use `-o allow_other` then the permissions on the s3fs mnt directory is root:root 700.

What I actually need in order to make chroot not reject the path is root:root 775.

Manually trying to chmod on the s3fs mnt directory doesn't work either:

```
# chmod 775 /s3fsmnt
chmod: changing permissions of ‘.’: Input/output error
```

Any way to do this?

consultantRR:
I patched line 3386 of s3fs.cpp from:
```
 mp_mode = S_IFDIR | (allow_other ? (S_IRWXU | S_IRWXG | S_IRWXO) : S_IRWXU);
```
to:
```
mp_mode = S_IFDIR | (allow_other ? (S_IRWXU | S_IRGRP | S_IXGRP | S_IROTH | S_IXOTH) : S_IRWXU);
```

...and I believe it is working fine. Perhaps this could be added as an option?

ggtakec:
Hi, consultantRR

Thanks for pull request.
After I see your code(idea), but I think that we should make new option like unmask for only mount point.
So that, s3fs can set any mode for mount point directory.

Then I'm going to update codes as soon as possible by your source code to reference.

Thanks in advance for your help.

ggtakec:
I merged #119.
It is added mp_umask option like umask, for mount point permission.
Thanks very much.
Regards,

PR #110, opened 2015-01-21
support for facilitate_chroot

consultantRR:
Used in tandem with `-o allow_other`, `-o facilitate_chroot` sets the permissions on the mount dir to 755 instead of 777 in order to provide support for backing a chroot jail.


ggtakec:
Hi, 
I merged #119, it is included from your codes.
Please see #107 comment.

I added new "mp_umask" option, it likes your facilitate_chroot option.
But mp_umask is as same as unmask option, so that it makes any permission for mount point directory.

Thanks very much.

PR #114, opened 2015-01-26
Disable pages merge

fcand-anevia:
I need to have some softwares that do random access to parts of objects. I cannot wait for the whole object to be downloaded, when I just need to access the first megabyte of data.

The current implementation merges available pages and download the object to the end. If a software needs to read small piece of data it will need to wait for the download to complete.

It would be convenient to have an option that disables it. This is what this pull requests is about.

Note : in order to see the benefit of this, I had to lower the pages sizes but the unit of multipart size was megabytes, so I took the liberty of changing the unit of the multipart_size option from mbytes to bytes. This can be rejected and be the purpose of another pull request.


fcand-anevia:
Actually, only commits c3ea03b  7b30736 and  b56d7ff are needed. I pushed other modifications but I should have used another branch ...

fcand-anevia:
Closing my branch unexpectedly closed the pull request. Reopening since still valid. The needed commits are still the same (c3ea03b 7b30736 and b56d7ff)

PR #122, opened 2015-02-12
Disable pages merge

fcand-anevia:
Sequel of pull request #114. Work was merged from branch to master then branch was deleted thus original pull request was closed. But it is still a valid pull request. Reopening here.

Needed commits are c3ea03b 7b30736 and b56d7ff


src/curl.cpp
```diff
@@ -188,7 +188,8 @@ const char* BodyData::str(void) const
 //-------------------------------------------------------------------
 // Class S3fsCurl
 //-------------------------------------------------------------------
-#define MULTIPART_SIZE              10485760          // 10MB
+#define MIN_MULTIPART_SIZE          (64*1024)           // 64k
```
andrewgaul: Minimum multipart size should be 5 MB expect for the last part which can be any size.  Do we need this?

andrewgaul:
@fcand-anevia Can you explain this pull request more completely? Also can you squash the commits, remove the spurious reformatting, and remove the Debian junk (submit as a separate PR)?

ggtakec:
@fcand-anevia 
This PR could not be merged, because your modified files(codes) is changed many.
So if you need to merge codes, please explain what to do and retry to push PR.

PR #124, opened 2015-02-17
Fallback to v2 signatures correctly.

timuralp:
Missing parameter to SetSignatureV4() call in the fallback code path
results in not actually falling back.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -3068,7 +3068,7 @@ static int s3fs_check_service(void)
       // retry to use sigv2
       LOWSYSLOGPRINT(LOG_ERR, "Could not connect, so retry to connect by signature version 2.");
       FPRN("Could not connect, so retry to connect by signature version 2.");
-      S3fsCurl::SetSignatureV4();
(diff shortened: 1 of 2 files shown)


src/s3fs.cpp
```diff
@@ -3068,7 +3068,7 @@ static int s3fs_check_service(void)
       // retry to use sigv2
       LOWSYSLOGPRINT(LOG_ERR, "Could not connect, so retry to connect by signature version 2.");
       FPRN("Could not connect, so retry to connect by signature version 2.");
-      S3fsCurl::SetSignatureV4();
+      S3fsCurl::SetSignatureV4(false);
```
kahing: remove the default argument from curl.h to prevent future errors?

timuralp: Done

andrewgaul:
:+1:

kahing:
:+1:

ggtakec:
Sorry for replying too late.
I merged this PR.

Thank you for your help.

PR #130, opened 2015-02-23
refactor integration tests create/cleanup file

kahing:


test/integration-test-main.sh
```diff
@@ -60,12 +89,7 @@ then
 fi
 
 # create the test file again
-echo $TEST_TEXT > $TEST_TEXT_FILE
-if [ ! -e $TEST_TEXT_FILE ]
-then
-   echo "Could not create file ${TEST_TEXT_FILE}, it does not exist"
-   exit 1
-fi
+mk_test_file
```
andrewgaul: Better to pass `$TEST_TEXT` explicitly instead of having a default behavior?

andrewgaul:
:+1:

test/integration-test-main.sh
kahing: I don't think the custom text is actually needed in the test that uses it, but I didn't want to change the actual tests in this pr. IMO it's better to just change the test to always use $TEST_TEXT and have mk_test_file to always write the same text.

Issue #129, opened 2015-02-23
run integration tests against S3Proxy


andrewgaul:
S3Proxy provides an S3 endpoint which does not need public Internet access or AWS credentials to run.  s3fs could bring this in as a submodule and run tests in Travis to prevent regressions.

ggtakec:
@andrewgaul I merged #132.
Thank you for your help.

PR #132, opened 2015-02-24
Use S3Proxy to run integration tests

andrewgaul:
References #129.


andrewgaul:
This needs S3Proxy 1.4.0 (not yet released) to run successfully.  Longer term I would like s3fs to use Travis or similar to automate this per-pull request.

Issue #141, opened 2015-03-05
Can create/remove objects but content is lost


benkehoe:
Starting with commit https://github.com/s3fs-fuse/s3fs-fuse/commit/114966e7c0371eed17e2898cdf505ba4dbfb9b59 (I have narrowed it to this commit) I observe the following:
```
$ s3fs my_bucket ~/s3data && cd ~/s3data
$ echo "foo" > bar
bash: echo: write error: Input/output error
$ ls
bar
$ cat bar
$ rm bar
$ ls
$
```
The object can be seen to be present in the S3 console, but with a size of 0 bytes.

andrewgaul:
This fails in `small-integration-test.sh` as well.  We should really automate these tests per-pull request via Travis #129!

admorphit:
+1

Am using centos 7 and getting error also
```
s3fs mybucket -f -d -o allow_other,del_cache,endpoint=us-west-1,url=https://s3-us-west-1.amazonaws.com /mnt/s3data/
```
Copy command fails
```
cp s3data/bigfile.mp4 /home
cp: error reading ‘s3data/bigfile.mp4’: Input/output error
cp: failed to extend ‘/home/bigfile.mp4’: Input/output error
```
and here's the log
```
s3fs_getattr(721): [path=/bigfile.mp4]
    HeadRequest(2112): [tpath=/bigfile.mp4]
    RequestPerform(1572): connecting to URL https://mybucket.s3-us-west-1.amazonaws.com/bigfile.mp4
    RequestPerform(1588): HTTP response code 200
    AddStat(248): add stat cache entry[path=/bigfile.mp4]
    GetStat(171): stat cache hit [path=/bigfile.mp4][time=1425723772][hit count=0]
s3fs_open(1911): [path=/bigfile.mp4][flags=32768]
    DelStat(370): delete stat cache entry[path=/bigfile.mp4]
    HeadRequest(2112): [tpath=/bigfile.mp4]
    RequestPerform(1572): connecting to URL https://mybucket.s3-us-west-1.amazonaws.com/bigfile.mp4
    RequestPerform(1588): HTTP response code 200
    AddStat(248): add stat cache entry[path=/bigfile.mp4]
    GetStat(171): stat cache hit [path=/bigfile.mp4][time=1425723772][hit count=0]
    ParallelGetObjectRequest(1105): [tpath=/bigfile.mp4][fd=5]
[...]

PR #143, opened 2015-03-08
Fixed a bug no use_cache case about fixed #138 - issue#141

ggtakec:
Fixed a bug about #141 
This bug was made at #138 commit.


admorphit:
looks good, thanks for the quick fix

PR #146, opened 2015-03-09
add exit handler to cleanup on failures

kahing:


test/small-integration-test.sh
```diff
@@ -1,10 +1,16 @@
-#!/bin/bash -e
+#!/bin/bash -ex
```
andrewgaul: Intentional change?  If so can you use `set -o xtrace` instead (and `set -o errexit`).

test/small-integration-test.sh
```diff
@@ -32,10 +38,10 @@ $S3FS $TEST_BUCKET_1 $TEST_BUCKET_MOUNT_POINT_1 \
     -o url=http://127.0.0.1:8080 \
     -o use_path_request_style
 
-./integration-test-main.sh $TEST_BUCKET_MOUNT_POINT_1
-
-fusermount -u $TEST_BUCKET_MOUNT_POINT_1
-
-kill $S3PROXY_PID
+if [ $? == 0 ]; then
```
andrewgaul: How does this ever false if `errexit` is set as above?

test/small-integration-test.sh
```diff
@@ -1,10 +1,16 @@
-#!/bin/bash -e
+#!/bin/bash -ex
 
 # Require root
 REQUIRE_ROOT=require-root.sh
 #source $REQUIRE_ROOT
 source integration-test-common.sh
 
+function exit_handler {
+    kill $S3PROXY_PID
```
andrewgaul: Call `set +o errexit` here in case S3Proxy failed to start?

test/small-integration-test.sh
kahing: done

test/small-integration-test.sh
kahing: removed

test/small-integration-test.sh
kahing: done

kahing:
also amended commit to dump s3fs output to log file to ease debugging

test/small-integration-test.sh
```diff
@@ -20,6 +48,8 @@ do
     sleep 1
 done
 
+S3PROXY_PID=$(netstat -lpnt | grep :8080 | awk '{ print $7 }' | sed -u 's|/java||')
```
andrewgaul: How about `lsof -i:8080 -t`?

andrewgaul:
:+1:

PR #149, opened 2015-03-10
URL-encode ampersand

andrewgaul:
Previously s3fs could not list a directory if it contained a file with
an ampersand character when using AWSv4 authentication.

### DIFF
--- a/src/string_util.cpp
+++ b/src/string_util.cpp
@@ -127,8 +127,6 @@ string urlEncode(const string &s)
       result += s[i];
     }else if (s[i] == '=') { // Note- special case for s3...
       result += s[i];
-    }else if (s[i] == '&') { // Note- special case for s3...
-      result += s[i];
     } else if (isalnum(s[i])) {
       result += s[i];
     } else if (s[i] == '.' || s[i] == '-' || s[i] == '*' || s[i] == '_') {


andrewgaul:
@ggtakec The URL encoding functions are rather wonky; can you take a broader look at these, especially the changes made since V4 signing merged?

ggtakec:
This code has been mixed at the time of the merge for sigv4, then I merged manually PR.
Because I think this function changing needs to check the entire code, meybe I take a little time.
Thanks a lot.

ggtakec:
Hi @andrewgaul 

I checked codes, and some problem about this fixing codes.
It can not remove the directory(or file?) with special character('&' and '=').
So I try to change those function as following.
```
string urlEncode(const string &s)
{
  string result;
  for (unsigned i = 0; i < s.length(); ++i) {
    if (s[i] == '/') { // Note- special case for fuse paths...
      result += s[i];
/*
    }else if (s[i] == '=') { // Note- special case for s3...
      result += s[i];
    }else if (s[i] == '&') { // Note- special case for s3...
      result += s[i];
*/
    } else if (isalnum(s[i])) {
      result += s[i];
    } else if (s[i] == '.' || s[i] == '-' || s[i] == '*' || s[i] == '_') {
      result += s[i];
    } else if (s[i] == ' ') {
      result += '%';
      result += '2';
      result += '0';
    } else {
      result += "%";
      result += hexAlphabet[static_cast<unsigned char>(s[i]) / 16];
      result += hexAlphabet[static_cast<unsigned char>(s[i]) % 16];
    }
  }
  return result;
}

string urlEncode2(const string &s)
{
  string result;
  for (unsigned i = 0; i < s.length(); ++i) {
    if (s[i] == '=') { // Note- special case for fuse paths...
      result += s[i];
    }else if (s[i] == '&') { // Note- special case for s3...
      result += s[i];
    }else if(s[i] == '%'){
      result += s[i];
    } else if (isalnum(s[i])) {
      result += s[i];
    } else if (s[i] == '.' || s[i] == '-' || s[i] == '*' || s[i] == '_') {
      result += s[i];
    } else if (s[i] == ' ') {
      result += '%';
      result += '2';
      result += '0';
    } else {
      result += "%";
      result += hexAlphabet[static_cast<unsigned char>(s[i]) / 16];
      result += hexAlphabet[static_cast<unsigned char>(s[i]) % 16];
    }
  }
  return result;
}
```
it seems this changed codes work good.
If you do not have problem about this codes, I will update codes soon.

ggtakec:
@andrewgaul I merged #154 instead of this #149.
Please check it.

ggtakec:
@andrewgaul I closed this pull request, because #154 instead of this changes.
Thanks a lot.

Issue #94, opened 2014-12-09
s3fs: failed to read - randomly occuring


mknwebsolutions:
I've got s3fs mounted and working, every once in a while I'll see a handful of errors i.e. below:

```
Dec  9 05:40:23 ip-10-0-0-16 s3fs: failed to read(remaining: 0 code: 28  msg: Timeout was reached), so retry this.
Dec  9 05:40:34 ip-10-0-0-16 s3fs: failed to read(remaining: 4 code: 28  msg: Timeout was reached), so retry this.
Dec  9 05:40:34 ip-10-0-0-16 s3fs: failed to read(remaining: 3 code: 28  msg: Timeout was reached), so retry this.
Dec  9 05:40:34 ip-10-0-0-16 s3fs: failed to read(remaining: 2 code: 28  msg: Timeout was reached), so retry this.
Dec  9 05:40:34 ip-10-0-0-16 s3fs: failed to read(remaining: 1 code: 28  msg: Timeout was reached), so retry this.
Dec  9 05:40:34 ip-10-0-0-16 s3fs: failed to read(remaining: 0 code: 28  msg: Timeout was reached), so retry this.
Dec  9 05:40:44 ip-10-0-0-16 s3fs: failed to read(remaining: 4 code: 28  msg: Timeout was reached), so retry this.
Dec  9 05:40:44 ip-10-0-0-16 s3fs: failed to read(remaining: 3 code: 28  msg: Timeout was reached), so retry this.
Dec  9 05:40:44 ip-10-0-0-16 s3fs: failed to read(remaining: 2 code: 28  msg: Timeout was reached), so retry this.
Dec  9 05:40:44 ip-10-0-0-16 s3fs: failed to read(remaining: 1 code: 28  msg: Timeout was reached), so retry this.
Dec  9 05:40:44 ip-10-0-0-16 s3fs: failed to read(remaining: 0 code: 28  msg: Timeout was reached), so retry this.
```

Eventually the file retry will reach limit and not pull file over s3fs.  Not sure why this is happening, it's been pretty random... I want to say I see it occurring more often when files are 20mb+ 

I'm able to switch into the s3fs directory and view the actual files / touch new files / etc.

chrislovecnm:
I am getting the exact same problem.  How can we help you debug this??  I have compiled the master

mknwebsolutions:
So I was actually able to get everything working after rebooting the server.  It just worked (and still working) since then.

[...]

PR #167, opened 2015-04-12
Increased default connecting/reading/writing timeout value

ggtakec:
Changed the default imeout value for connect and read/write for #94.
Because the timeout default value for connecting is 10 seconds as default, but it seems too small.
And read/write timeout is small too.
So the connect default value is changed 300 seconds which is as same as curl default.
And the read/write timeout is set 60 seconds.


ml019:
-help in 1.79 still shows default values as 10s for connect and 30s for read/write. Presume this is static text and not linked to actual values used by the code?

Thanks

ggtakec:
@ml019 @andrewgaul Thanks.

PR #171, opened 2015-04-19
Support buckets with mixed-case names

pabigot:
Buckets with mixed-case names can't be accessed with the virtual-hosted style API due to DNS limitations.  S3FS has an option for pathrequeststyle which is used for the URL, but it was not applied when building the endpoint passed through the Host header.  Fix this, and relax the validation on bucket names when using this style.

See: http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#access-bucket-intro

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -121,7 +121,10 @@ static string url_to_host(const std::string &url)
 
 static string get_bucket_host()
 {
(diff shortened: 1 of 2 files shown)


ggtakec:
@pabigot Thanks a lot.

PR #176, opened 2015-04-26
configure.ac: detect target, if target is darwin (OSX), then

mooredan:
Builds and operates on OSX Yosemite using MacPorts osxfuse

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -2741,6 +2741,7 @@ static void* s3fs_init(struct fuse_conn_info* conn)
   }
 
   // Investigate system capabilities
+  #ifndef __APPLE__
   if((unsigned int)conn->capable & FUSE_CAP_ATOMIC_O_TRUNC){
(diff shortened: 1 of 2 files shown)


src/s3fs.cpp
```diff
@@ -2748,6 +2749,7 @@ static void* s3fs_init(struct fuse_conn_info* conn)
   if(is_remove_cache && !FdManager::DeleteCacheDirectory()){
     DPRNINFO("Could not inilialize cache directory.");
   }
+  #endif
```
ggtakec: For __APPLE__ case, does FdManager::DeleteCacheDirectory() occur something error?

mooredan:
Oops, that #endif is in the wrong spot, it should be a few lines up. I'll correct this shortly.  Good catch!

mooredan:
Fixed pull request

ggtakec:
@mooredan Thanks for fixing. I merged this.

PR #181, opened 2015-05-01
fix v4 fallback to work correctly for GCS

### LINKED ISSUE #179: Invalid Security Credentials - Google Cloud
Hello,

I'm trying to connect to my google cloud storage bucket and I'm using the following format in the /etc/passwd-s3fs
Your documentation states to use accessKeyId:secretAccessKey as the format for the credentials file.  However, when I try to connect using the Client ID and Client Secret provided by the google cloud api, it gives me this as the result:
`  $s3fs storage1 /mnt/user_storage -oallow_other -o url=https://storage.googleapis.com -f -d
    set_moutpoint_attribute(3539): PROC(uid=0, gid=0) - MountPoint(uid=0, gid=0, mode=40755)
s3fs_init(2713): init
s3fs_check_service(3072): check services.
    CheckBucket(2525): check a bucket.
    insertV4Headers(1961): computing signature [GET] [/] [] []
    url_to_host(99): url is https://storage.googleapis.com
    RequestPerform(1619): connecting to URL https://storage1.storage.googleapis.com/
    RequestPerform(1653): HTTP response code 403 was returned, returning EPERM
CheckBucket(2563): Check bucket failed, S3 response: <?xml version='1.0' encoding='UTF-8'?><Error><Code>InvalidSecurity</Code><Message>The provided security credentials are not valid.</Message><Details>Incorrect Authorization header</Details></Error>`

Any help would be appreciated.

kahing:
fixes #179

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -3078,28 +3078,17 @@ static int s3fs_check_service(void)
   }
 
   S3fsCurl s3fscurl;
-  if(-1 == s3fscurl.CheckBucket()){
+  if(1 == s3fscurl.CheckBucket()){
     fprintf(stderr, "%s: Failed to access bucket.\n", program_name.c_str());
     return EXIT_FAILURE;
   }
   long responseCode = s3fscurl.GetLastResponseCode();
 
-  if(responseCode == 400){
-    if(!S3fsCurl::IsSignatureV4()){
-      // signature version 2
-      fprintf(stderr, "%s: Bad Request\n", program_name.c_str());
-      return EXIT_FAILURE;
-    }
-    if(is_specified_endpoint){
-      // if specifies endpoint, do not retry to connect.
-      fprintf(stderr, "%s: Bad Request\n", program_name.c_str());
-      return EXIT_FAILURE;
-    }
-
+  if(responseCode == 400) {
     // check region error for signature version 4
     BodyData* body = s3fscurl.GetBodyData();
     string    expectregion;
-    if(check_region_error(body->str(), expectregion)){
+    if(check_region_error(body->str(), expectregion) && !is_specified_endpoint){
       // not specified endpoint, so try to connect to expected region.
       LOWSYSLOGPRINT(LOG_ERR, "Could not connect wrong region %s, so retry to connect region %s.", endpoint.c_str(), expectregion.c_str());
       FPRN("Could not connect wrong region %s, so retry to connect region %s.", endpoint.c_str(), expectregion.c_str());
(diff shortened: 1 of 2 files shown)


andrewgaul:
:+1:

src/s3fs.cpp
```diff
@@ -3078,28 +3078,17 @@ static int s3fs_check_service(void)
   }
 
   S3fsCurl s3fscurl;
-  if(-1 == s3fscurl.CheckBucket()){
+  if(1 == s3fscurl.CheckBucket()){
```
ggtakec: We should not change codes in curl.cpp, so I think this if-steatment codes(from L:3081 to L3084) should be removed.

src/s3fs.cpp
```diff
@@ -3078,28 +3078,17 @@ static int s3fs_check_service(void)
...
-      fprintf(stderr, "%s: Bad Request\n", program_name.c_str());
-      return EXIT_FAILURE;
-    }
-
+  if(responseCode == 400) {
     // check region error for signature version 4
     BodyData* body = s3fscurl.GetBodyData();
     string    expectregion;
-    if(check_region_error(body->str(), expectregion)){
+    if(check_region_error(body->str(), expectregion) && !is_specified_endpoint){
```
ggtakec: Checking "!is_specified_endpoint" should be first. because the checking cost is is_specified_endpoint < check_region_error flag.

src/s3fs.cpp
```diff
@@ -3120,26 +3109,27 @@ static int s3fs_check_service(void)
...
-        return EXIT_FAILURE;
-      }
-      responseCode = s3fscurl.GetLastResponseCode();
-      if(responseCode == 400){
-        fprintf(stderr, "%s: Bad Request\n", program_name.c_str());
-        return EXIT_FAILURE;
-      }
+    // retry to check
+    s3fscurl.DestroyCurlHandle();
+    if(-1 == s3fscurl.CheckBucket()){
```
ggtakec: this line should be changed one line.("s3fscurl.CheckBucket();")

ggtakec:
@kahing thanks you for your help.

I think that we should not change the return code in RequestPerform() method.
(If need to change the return code, it should change the return code in CheckBucket().)

More than anything, I'm sorry this issue(bug) is my fault in s3fs_check_service() function.
In s3fs_check_service, s3fs checks the result from calling s3fscurl.CheckBucket(), it should be following.
```
if(0 > s3fscurl.CheckBucket()){
  or
No checking the result code from CheckBucket, and check only respons code.
```
When any http error(over 400) is occurred at checking bucket, s3fs should try to signature v2.

So, we should change only the code in s3fs_check_service() function s3fs.cpp.
Your fixed code for s3fs.cpp seems ok for me, and I commented some.

Please check my comment.

ggtakec:
@kahing  
I'm sorry for waiting your response, and merged new codes(#200) which is based your this code.
I closed this PR, and if you found bugs please let me know.
Regards,

PR #190, opened 2015-05-20
Add a no_check_certificate option.

Rotwang:
Which allows to ignore issues with self signed certs. For example in the company I work for we use our own s3 deployment and our certificates are self signed, so by default s3fs can't be used because of:

s3fs_check_service(3072): check services.
    CheckBucket(2537): check a bucket.
    RequestPerform(1631): connecting to URL https://mahbucket.mahs3.net/
RequestPerform(1733): ### CURLE_SSL_CACERT
    RequestPerform(1790): ### retrying...
    RemakeHandle(1453): Retry request. [type=5][url=https://mahbucket.mahs3.net//][path=/]
RequestPerform(1733): ### CURLE_SSL_CACERT
RequestPerform(1744): curlCode: 60  msg: Peer certificate cannot be authenticated with given CA certificates

This pull request adds non secure option to ignore this kind of errors.


ggtakec:
Thank you.

PR #192, opened 2015-06-06
Simplify URL encoding

### LINKED ISSUE #188: Attempts to access non-existent files may result in bogus EPERM instead of ENOENT
If the attempted name contains asterisk, the error-code will be `EPERM`, which is incorrect. Otherwise, the failure is `ENOENT`. Here, for example, we have an S3-bucket mounted under `/data/ftp`. Trying to list three non-existent names, one gets different results:
```
% ls -l '/data/ftp/*ogus' '/data/ftp/?ogus' '/data/ftp/bogus'
ls: cannot access /data/ftp/*ogus: Operation not permitted
ls: cannot access /data/ftp/?ogus: No such file or directory
ls: cannot access /data/ftp/bogus: No such file or directory
```
Note, how using ?-template still works properly, but asterisk must have some special meaning, which causes erroneous behavior.

### LINKED ISSUE #194: Tilda in a file-name breaks things (EPERM)
Trying to `touch` a file `test~` or `mkdir` a directory with the same name results in `EPERM`:

    test~: Operation not permitted

Seems like Issue #188, but the tilda-symbol does _not_ have a special handling in the `urlEncode`-function (whereas asterisk used to). Maybe, unlike `'*'`, the `'~'` _does_ need to be passed-through unescaped?

andrewgaul:
This also encodes asterisk and tilde correctly when listing a file
with a V4 auth endpoint.  Also add tests for special characters
although s3proxy does not yet support V4 auth.
Fixes #188.  Fixes #194.

### DIFF
--- a/src/string_util.cpp
+++ b/src/string_util.cpp
@@ -123,20 +123,20 @@ string urlEncode(const string &s)
 {
   string result;
   for (unsigned i = 0; i < s.length(); ++i) {
-    if (s[i] == '/') { // Note- special case for fuse paths...
-      result += s[i];
-    } else if (isalnum(s[i])) {
-      result += s[i];
-    } else if (s[i] == '.' || s[i] == '-' || s[i] == '*' || s[i] == '_') {
-      result += s[i];
-    } else if (s[i] == ' ') {
-      result += '%';
-      result += '2';
-      result += '0';
+    char c = s[i];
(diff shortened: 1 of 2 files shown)


andrewgaul:
Not sure what is going on with Travis -- this passes local tests for me.

ggtakec:
Merged this PR(travisci error is skipped.)

Issue #169, opened 2015-04-16
Problem with encfs


delx:
I executed s3fs-fuse (latest master) with the following command:
```
/opt/s3fs-fuse/bin/s3fs -f -d bucket /mnt/s3/ -o allow_other,uid=33,noatime,max_stat_cache_size=0
```

Then on top of s3fs-fuse I mounted a encfs file system:
```
encfs -f -d --public --extpass="cat /root/password" -o uid=33 /mnt/s3/ /mnt/encfs/
```

There is a file with path `/mnt/encfs/test/file` that has 30 KB. I executed the following command:
```
echo CONTENT > /mnt/encfs/test/file
```

**After this command the file still has 30 KB although it should only have 8 bytes!**
```
root# ls -l /mnt/encfs/test/file
-rw-r--r-- 1 www-data root 29395 Apr 16 15:38 /mnt/encfs/test/file
```

The output of s3fs-fuse:
```
set_moutpoint_attribute(3530): PROC(uid=0, gid=0) - MountPoint(uid=0, gid=0, mode=40755)
s3fs_init(2713): init
s3fs_check_service(3070): check services.
    CheckBucket(2518): check a bucket.
    insertV4Headers(1954): computing signature [GET] [/] [] []
    url_to_host(99): url is http://s3.amazonaws.com
    RequestPerform(1616): connecting to URL http://bucket.s3.amazonaws.com/
    RequestPerform(1632): HTTP response code 400
    RequestPerform(1646): HTTP response code 400 was returned, returing EIO.
CheckBucket(2556): Check bucket failed, S3 response: <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AuthorizationHeaderMalformed</Code><Message>The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'eu-central-1'</Message><Region>eu-central-1</Region><RequestId>38128C004E63105C</RequestId><HostId>[cut]</HostId></Error>
s3fs_check_service(3103): Could not connect wrong region us-east-1, so retry to connect region eu-central-1.
    CheckBucket(2518): check a bucket.
    insertV4Headers(1954): computing signature [GET] [/] [] []
    url_to_host(99): url is http://s3.amazonaws.com
    RequestPerform(1616): connecting to URL http://bucket.s3.amazonaws.com/
    RequestPerform(1632): HTTP response code 200

[encfs starts now]

s3fs_getattr(722): [path=/]
[...]

PR #196, opened 2015-06-13
Supported extended attributes

ggtakec:
Supports extended attributes.
s3fs saves extended attributes to object's header "x-amz-meta-xattr" which is formatted as  encoded json.
```
x-amz-meta-xattr: url encoded string({"key":"base64(value)","key":"base64(value)",....})
```
This change solves issues #166 and #169.

### DIFF
--- a/src/common_auth.cpp
+++ b/src/common_auth.cpp
@@ -30,7 +30,7 @@ using namespace std;
 //-------------------------------------------------------------------
 // Utility Function
 //-------------------------------------------------------------------
-char* s3fs_base64(unsigned char* input, size_t length)
+char* s3fs_base64(const unsigned char* input, size_t length)
 {
   static const char* base = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=";
   char* result;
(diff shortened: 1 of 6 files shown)


andrewgaul:
It might be nice to HTTP percent encode these instead of base64 so that other tools can introspect one them in the common case.

andrewgaul:
Could we store each EA as its own user metadata key as well, e.g., `x-amz-meta-s3fs-ea-KEYNAME`

src/s3fs.cpp
```diff
@@ -2708,6 +2721,472 @@ static int remote_mountpath_exists(const char* path)
...
+    strpath  = path;
+    nowcache = strpath;
+    result   = get_object_attribute(strpath.c_str(), NULL, &meta);
+  }
+  if(0 != result){
+    return result;
+  }
+
+  // get xattrs
+  if(meta.end() == meta.find("x-amz-meta-xattr")){
```
andrewgaul: Reuse the iterator, e.g.,

```c++
headers_t::iterator it = meta.find("x-amz-meta-xattr");
if (it == meta.end()) { ... }
string strxattrs = *it;
```

src/s3fs.cpp
```diff
@@ -2708,6 +2721,472 @@ static int remote_mountpath_exists(const char* path)
...
+  if(xattrs.end() == xattrs.find(strname)){
+    free_xattrs(xattrs);
+    return -ENOATTR;
+  }
+
+  // make new header_t after deleting name xattr
+  if(xattrs[strname]){
+    delete xattrs[strname];
+  }
+  xattrs.erase(strname);
```
andrewgaul: Same reuse iterator instead of three lookups.

src/s3fs.cpp
```diff
@@ -2708,6 +2721,472 @@ static int remote_mountpath_exists(const char* path)
...
+  if(xattrs.end() == xattrs.find(strname)){
+    // not found name in xattrs
+    free_xattrs(xattrs);
+    return -ENOATTR;
+  }
+
+  // decode
+  size_t         length = 0;
+  unsigned char* pvalue = NULL;
+  if(NULL != xattrs[strname]){
```
andrewgaul: This should always be true -- `xattrs[strname]` will call the default constructor here.

src/s3fs.cpp
```diff
@@ -2708,6 +2721,472 @@ static int remote_mountpath_exists(const char* path)
...
+    }
+
+    // need to set xattr header for directory.
+    strpath  = newpath;
+    nowcache = strpath;
+  }
+
+  // set xattr all object
+  meta["x-amz-copy-source"]        = urlEncode(service_path + bucket + get_realpath(strpath.c_str()));
+  meta["x-amz-metadata-directive"] = "REPLACE";
```
andrewgaul: Call `meta.insert("x-amz-metadata-directive", "REPLACE")` to avoid an unneeded default constructor.

src/s3fs.cpp
```diff
@@ -2708,6 +2721,472 @@ static int remote_mountpath_exists(const char* path)
...
+  }
+
+  // make new header_t after deleting name xattr
+  if(xattrs[strname]){
+    delete xattrs[strname];
+  }
+  xattrs.erase(strname);
+
+  // build new xattr
+  if(0 < xattrs.size()){
```
andrewgaul: Call `xattrs.empty()`

src/s3fs.cpp
ggtakec: Here want to give priority to the people of the readability of the source.

ggtakec:
@andrewgaul 
Do you know the reason about travis error about small-integration-test.sh?

src/common_auth.cpp
```diff
@@ -61,6 +61,61 @@ char* s3fs_base64(unsigned char* input, size_t length)
...
+  }else if('0' <= ch && ch <= '9'){             // 0 - 9
+    by = static_cast<unsigned char>(ch - '0' + 52);
+  }else if('+' == ch){                         // +
+    by = 62;
+  }else if('/' == ch){                         // /
+    by = 63;
+  }else if('=' == ch){                         // =
+    by = 64;
+  }else{                                       // something wrong
+    by = 64;
```
andrewgaul: Call `std::abort`?

ggtakec: I think that we should not terminate s3fs as a character has been detected illegal.
Because it is base64 string that has been written in the http header, there is a possibility that illegal character exists.
It will skip the character instead, and it will the act is another error(ex. not found extended attribute)

Thanks.

ggtakec:
I closed this PR, and will post new PR.
Because .travis.yml is changed for passing tavis.ci.

PR #202, opened 2015-06-25
Specialize {set,get}xattr for OS X

flandr:
These system calls take an extra 'position' parameter on OS X. A
non-zero position value is only valid for resource forks (the Darwin
VFS layer will reject anything else with EINVAL); this patch simply
adds and ignores the parameter on Apple platforms.

Allows building against OSXFUSE.


ggtakec:
@flandr Many thanks for this PR.

PR #213, opened 2015-07-27
Parse ETag from copy multipart correctly

andrewgaul:
Previously s3fs misparsed this, preventing renames of files larger
than 5 GB.  Integration test disabled until S3Proxy 1.5.0 is released.


src/curl.cpp
```diff
@@ -3127,10 +3127,10 @@ int S3fsCurl::CopyMultipartPostRequest(const char* from, const char* to, int par
 
   int result = RequestPerform();
   if(0 == result){
-    const char* start_etag= strstr(bodydata->str(), "ETag");
-    const char* end_etag  = strstr(bodydata->str(), "/ETag>");
-
-    partdata.etag.assign((start_etag + 11), (size_t)(end_etag - (start_etag + 11) - 7));
+    const char* start_etag = strstr(bodydata->str(), "<ETag>\"");
```
kahing: add a comment example of what you are parsing

src/s3fs_util.cpp
```diff
@@ -971,6 +971,10 @@ void show_help (void)
     "   multipart_size (default=\"10\")\n"
     "      - part size, in MB, for each multipart request.\n"
     "\n"
+    "   singlepart_copy_limit (default=\"5120\")\n"
```
kahing: maybe not document this since it's only used for testing?

src/curl.cpp
andrewgaul: Added a comment and converted this to proper XML parsing.

PR #215, opened 2015-07-30
Fix mem leak in openssl_auth.cpp:s3fs_sha256hexsum

RobbKistler:
Destroy sha256ctx on all function returns. Otherwise memory leaks
every time a new file is created.

### DIFF
--- a/src/openssl_auth.cpp
+++ b/src/openssl_auth.cpp
@@ -297,10 +297,8 @@ bool s3fs_sha256(const unsigned char* data, unsigned int datalen, unsigned char*
 
 unsigned char* s3fs_sha256hexsum(int fd, off_t start, ssize_t size)
 {
-  const EVP_MD* md        = EVP_get_digestbyname("sha256");
-  EVP_MD_CTX*   sha256ctx = EVP_MD_CTX_create();
-  EVP_DigestInit_ex(sha256ctx, md, NULL);
-
+  const EVP_MD*  md = EVP_get_digestbyname("sha256");
+  EVP_MD_CTX*    sha256ctx;
   char           buf[512];
   ssize_t        bytes;
   unsigned char* result;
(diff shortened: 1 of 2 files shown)


src/openssl_auth.cpp
```diff
@@ -315,6 +316,7 @@ unsigned char* s3fs_sha256hexsum(int fd, off_t start, ssize_t size)
 
   // seek to top of file.
   if(-1 == lseek(fd, start, SEEK_SET)){
+    EVP_MD_CTX_destroy(sha256ctx);
```
andrewgaul: Can we sink `sha26ctx` initialization closer to its first use to avoid the first two calls to `EVP_MD_CTX_destroy`?  This will more closely match `nss_auth.cpp:s3fs_sha256hexsum`.

andrewgaul:
Does this leak every time a new file is created or only on failure paths?  Also can you add the missing call to `PK11_DestroyContext` in `nss_auth.cpp:s3fs_sha256hexsum`?

RobbKistler:
It leaks every time a new file is created, per valgrind. 
Latest commits should address your feedback.  Thanks!

src/openssl_auth.cpp
```diff
@@ -330,10 +328,15 @@ unsigned char* s3fs_sha256hexsum(int fd, off_t start, ssize_t size)
       DPRNNN("file read error(%d)", errno);
       return NULL;
     }
+
+    const EVP_MD* md        = EVP_get_digestbyname("sha256");
+    sha256ctx = EVP_MD_CTX_create();
+    EVP_DigestInit_ex(sha256ctx, md, NULL);
```
andrewgaul: You need to declare `sha256ctx` outside the for loop since it accumulates results in every iteration.

RobbKistler:
Oops!  Sorry about that.  I just amended that commit with that fix and also put the declaration of `md` back where it was originally to be consistent with the rest of the code.

andrewgaul:
:+1: Please squash commits.

RobbKistler:
Squashed.  Thanks!

andrewgaul:
:+1:

ggtakec:
@RobbKistler Thanks.

PR #219, opened 2015-08-06
Address Coverity errors

andrewgaul:
Fixed an uninitialized member, misordered NULL check, resource leak,
and unconsumed return value.


kahing:
:+1: not everything is impactful but they look correct.

PR #224, opened 2015-08-08
Configure cppcheck

andrewgaul:


andrewgaul:
Some overlap with #219.

kahing:
:+1: make release depend on cpp_check

kahing:
or to be more aggressive, make the s3fs target depend on it

PR #231, opened 2015-08-15
Rewrite AutoLock

andrewgaul:
Previously `AutoLock::Lock` allowed subsequent callers to proceed
without the lock.  Further `is_locked` was not always protected by
`auto_mutex`.  Finally `AutoLock` eagerly released `auto_mutex` when
recursively unlocking.  s3fs does not need recursive locks so we
rewrite and simplify `AutoLock`.  Partially surfaced by Coverity.


ggtakec:
You are right, now we do not need recursive mutex(AutoLock).
If we need it, we will use mutex attribute as recursive.
Thanks,

PR #240, opened 2015-08-17
Enable Content-MD5 during multipart upload part

andrewgaul:
This allows retries of multi-part uploads instead of discovering a
fatal error during complete multipart upload.  Also enable Content-MD5
for integration tests and refactor hexadecimal code.


ggtakec:
Thanks, :+1:

PR #251, opened 2015-08-21
Skip xattr tests if utilities are missing

flandr:
The `attr` package isn't installed by default on many distributions.


ggtakec:
Thanks.

PR #252, opened 2015-08-21
This fixes an issue with caching when the creation of a subdirectory …

oliverfriedmann:
…within the cache is aborted because a common cached parent directory already exists.


ggtakec:
We should check directory stat before mkdir, and result of mkdirp in CacheFileStat::MakeCacheFileStatPath.
(If mkdirp returns not zero, this method should return false.)
I try to modify this after merging this pr.
Thanks,

ggtakec:
#253, Added checking cache directory permission at starting s3fs.
If you have a problem, please let me know.
Thanks,

PR #256, opened 2015-09-01
Add no atomic rename to limitations

andrewgaul:


akiradeveloper:
could you please explain the theoretical background of this commit?

andrewgaul:
@akiradeveloper POSIX specifies that rename is an *atomic* operation, i.e., if it succeeds you see the new renamed file/directory and if it fails you only see the original.  s3fs cannot guarantee this even for renaming a file; instead it server-side copies the original to the new name then removes the original.  If the operation fails you can have two files, the original and the new.  This is most notable when renaming a directory with many files underneath it.

akiradeveloper:
So you explained what s3fs does for rename operation and I am convinced that way can't guarantee atomicity.

I think the limitation is because backend object storages in general don't provide atomic rename operation.

Do you know other atomic operations in POSIX are guaranteed atomic in s3fs?

andrewgaul:
@akiradeveloper Some of the atomicity depends on the underlying object store.  For example, AWS S3 has eventual consistency for read-after-overwrite so mutating a file is not atomic to clients.  DreamObjects guarantees strong consistency for the same operation.  Further discussion:

https://github.com/andrewgaul/are-we-consistent-yet

PR #257, opened 2015-09-05
Update README.md: Bugfix password file permissions errors

jesselsteele:
received "should not have others permissions" when mounting
did chmod 640 and received "should not have group permissions"
used chmod 600 after creating the password file and had no problems


andrewgaul:
:+1:

PR #258, opened 2015-09-07
Update README.md to better explain mount upon boot

juandiegogonzales:
As a novice Linux user, I didn't know I had to add a line into /etc/fstab for automatic mount upon boot. It took me some minutes to research and notice the right process (at first, I even entered s3fs#mybucket into the command line).

This change will (hopefully) save time to unseasoned users.


andrewgaul:
:+1:

PR #265, opened 2015-09-11
Fix integration tests

RobbKistler:
Update test_append_file to keep file open during tests

Update remote_nonepty_directory to prevent passing test from causing exit

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -69,10 +69,8 @@ function test_append_file {
     # Write a small test file
     for x in `seq 1 $TEST_TEXT_FILE_LENGTH`
     do
-       echo "echo ${TEST_TEXT} ${x} to ${TEST_TEXT_FILE}"
-       echo "$TEST_TEXT ${x}" >&3
-    done
-    exec 3>&-  # close
+       echo "echo ${TEST_TEXT} to ${TEST_TEXT_FILE}"
+    done > ${TEST_TEXT_FILE}
 
     # Verify contents of file
     echo "Verifying length of test file"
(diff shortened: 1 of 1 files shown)


test/integration-test-main.sh
```diff
@@ -63,13 +63,16 @@ function rm_test_dir {
...
+    exec 3<> ${TEST_TEXT_FILE}  # open
     # Write a small test file
     for x in `seq 1 $TEST_TEXT_FILE_LENGTH`
     do
-       echo "echo ${TEST_TEXT} to ${TEST_TEXT_FILE}"
-       echo $TEST_TEXT >> $TEST_TEXT_FILE
+       echo "echo ${TEST_TEXT} ${x} to ${TEST_TEXT_FILE}"
+       echo "$TEST_TEXT ${x}" >&3
     done
+    exec 3>&-  # close
```
andrewgaul: Instead of this could you just make the loop pipe into the file, e.g.,

```
for x in `seq 1 $TEST_TEXT_FILE_LENGTH`
do
   echo "echo ${TEST_TEXT} to ${TEST_TEXT_FILE}"
done > $TEST_TEXT_FILE
```

test/integration-test-main.sh
```diff
@@ -258,7 +261,10 @@ function test_remove_nonempty_directory {
     echo "Testing removing a non-empty directory"
     mk_test_dir
     touch "${TEST_DIR}/file"
-    rmdir "${TEST_DIR}" 2>&1 | grep -q "Directory not empty"
+    if ! (rmdir "${TEST_DIR}" 2>&1 | grep -q "Directory not empty"); then
+        echo "Did not get \"Directory not empty\""
+        exit 1
+    fi
```
andrewgaul: What is the intent of this change?  Previously we exit if we do not see "Directory not empty" and pass only if we do see it.

test/integration-test-main.sh
RobbKistler: Done.  I didn't know if anyone would miss the separate echo to stdout.

test/integration-test-main.sh
RobbKistler: This change should do the same thing but its not necessary.  I removed it in the last commit (I'll squash commits).  

I was confused because this test is also failing against Amazon S3 servers.  Same problem -- object consistency.  The ```touch testdir/file``` object can show up after the directory has been removed, leaving the "testdir/file" object behind with no directory "testdir" directory object.

PR #267, opened 2015-09-15
Added the _netdev option to the fstab example.

nickstinger:
Although the network device option (_netdev) may not work everywhere, the option likely does no harm on systems where it's not supported. By adding this option to the example, it will inform users of the necessity for post-network activation and how that might be accomplished.


andrewgaul:
:+1:

PR #268, opened 2015-09-16
Use 127.0.0.1 not localhost in s3proxy wait loop

RobbKistler:
localhost doesn't always resolve to 127.0.0.1 which can bash loop that waits for s3proxy to start will fail.


andrewgaul:
:+1:

PR #271, opened 2015-09-17
Add support for standard_ia storage class

andrewgaul:
This enables storage with lower at-rest prices, higher request prices,
and lower availability.  Also rework existing reduced redundancy
parsing into a more generic storage class.  More background on
standard_ia:

https://aws.amazon.com/blogs/aws/aws-storage-update-new-lower-cost-s3-storage-option-glacier-price-reduction/


ggtakec:

ggtakec:
@andrewgaul #274 for man page, please see it and if you found something wrong, please let me know.

Issue #270, opened 2015-09-17
Support for KMS


jpeacock-mypoints:
Does S3FS support AWS-KMS keys?  I don't see a way to implement those.

ggtakec:
@jpeacock-mypoints 
I merged new codes for this issue(support for SSE-KMS) - #278 
And modified use_sse option for this change.
After this change, s3fs supported No-SSE(SSE disable), SSE-S3, SSE-C(custom key using), and SSE-KMS(AWS Key Management Id).

Please try to use and check latest code on master branch.
Thanks in advance for your help.

##### About use_sse option:
- not specify use_sse option  
default is SSE-DISABLE
- "use_sse" or "use_sse=1"(old type parameter)  
uses Amazon S3-managed encryption keys
- "use_sse=custom:<filepath>" or "use_sse=<filepath>"(old type parameter)  
uses customer-provided encryption keys.  
The custom key file must be 600 permission.  
The file can have some lines, each line is one SSE-C key.  
The first line in file is used as Customer-Provided Encryption Keys for uploading and changing headers etc.  
If there are some keys after first line, those are used downloading object which are encrypted by not first key.  
So that, you can keep all SSE-C keys in file, that is SSE-C key history.  
- "use_sse=custom"  
If you specify "custom"("c") without file path, you need to set custom key by load_sse_c option or AWSSSECKEYS environment.  
(AWSSSECKEYS environment has some SSE-C keys with ":" separator.)  
This option is used to decide the SSE type.  
So that if you do not want to encrypt a object object at uploading, but you need to decrypt encrypted object at downloaing, you can use load_sse_c option instead of this option.  
- "use_sse=kmsid" or "use_sse=kmsid:<kms id>"  
uses the master key which you manage in AWS KMS.  
You can use "k" for short "kmsid".  
If you san specify SSE-KMS type with your <kms id> in AWS KMS, you can set it after "kmsid:"(or "k:").  
If you specify only "kmsid"("k"), you need to set AWSSSEKMSID environment which value is <kms id>.  

##### notice
You must be careful about that you can not use the KMS id which is not same EC2 region.

ggtakec:
I closed this issue. but you have(find) a problem, please let us know.
Regards,

PR #278, opened 2015-10-06
Supported for SSE KMS(#270)

ggtakec:
Supported for Server Side Encrypting with KMS id.  
And modified use_sse option for this change.  
After this change, s3fs supported No-SSE(SSE disable), SSE-S3, SSE-C(custom key using), and SSE-KMS(AWS Key Management Id).  

##### About use_sse option:
- not specify use_sse option  
default is SSE-DISABLE
- "use_sse" or "use_sse=1"(old type parameter)  
uses Amazon S3-managed encryption keys
- "use_sse=custom:<filepath>" or "use_sse=<filepath>"(old type parameter)  
uses customer-provided encryption keys.  
The custom key file must be 600 permission.  
The file can have some lines, each line is one SSE-C key.  
The first line in file is used as Customer-Provided Encryption Keys for uploading and changing headers etc.  
If there are some keys after first line, those are used downloading object which are encrypted by not first key.  
So that, you can keep all SSE-C keys in file, that is SSE-C key history.  
- "use_sse=custom"  
If you specify "custom"("c") without file path, you need to set custom key by load_sse_c option or AWSSSECKEYS environment.  
(AWSSSECKEYS environment has some SSE-C keys with ":" separator.)  
This option is used to decide the SSE type.  
So that if you do not want to encrypt a object object at uploading, but you need to decrypt encrypted object at downloaing, you can use load_sse_c option instead of this option.  
- "use_sse=kmsid" or "use_sse=kmsid:<kms id>"  
uses the master key which you manage in AWS KMS.  
You can use "k" for short "kmsid".  
If you san specify SSE-KMS type with your <kms id> in AWS KMS, you can set it after "kmsid:"(or "k:").  
If you specify only "kmsid"("k"), you need to set AWSSSEKMSID environment which value is <kms id>.  
You must be careful about that you can not use the KMS id which is not same EC2 region.


soufrk:
Unable to get 'load_sse_c' option working. Results of trials below,
```
$ sudo s3fs mybucket /var/s3 -o allow_other -o load_sse_c=/etc/mykey
fuse: unknown option `load_sse_c=/etc/mykey'
$ sudo s3fs mybucket /var/s3 -o allow_other -o use_sse=custom -o load_sse_c /etc/mykey
s3fs: specified unknown third optioni(/etc/mykey).
$ sudo s3fs mybucket /var/s3 -o allow_other -o use_sse=custom:/etc/mykey -o load_sse_c
fuse: unknown option `load_sse_c'
$ sudo s3fs mybucket /var/s3 -o allow_other -o use_sse=custom -o load_sse_c
s3fs: something wrong about SSE options.
$ sudo s3fs mybucket /var/s3 -o allow_other -o use_sse=custom -o load_sse_c /etc/mykey
s3fs: specified unknown third optioni(/etc/mykey).
$ sudo s3fs mybucket /var/s3 -o allow_other -o use_sse=custom -o load_sse_c=/etc/mykey
fuse: unknown option `load_sse_c=/etc/mykey'
$ sudo s3fs mybucket /var/s3 -o allow_other -o use_sse=c -o load_sse_c=/etc/mykey
fuse: unknown option `load_sse_c=/etc/mykey'
$ sudo s3fs mybucket /var/s3 -o allow_other -o use_sse=c -o load_sse_c /etc/mykey
```

Any help will be appreciated.

PR #280, opened 2015-10-18
Supported a object which is larger than free disk space

ggtakec:
#### Supported a object which is larger than free disk space, and added ensure_diskfree option
About this change, s3fs uploads the object which is larger than free disk space on demand.
In that situation, s3fs stops the multipart upload of in parallel, and uplaods each part as serial.

##### Following option is added/deleted.
* Adds new option, ensure_diskfree  
This option means the threshold of free space size on disk which is used for the cache file by s3fs.
If the free disk space is smaller than this threshold, s3fs download/upload the object without using the cache files.
The default value for this option is the same as the value of multipart_size.
With this option, s3fs can upload a file which is larger than free disk space.
If nomultipart is specified, s3fs failed to upload when over free disk space.

* fd_page_size option is no longer used  
This option was not the designation of one page size of the cache file, this designation is no longer required.


ggtakec:
This failed at comparing large objects by downloading, so I will fix it as soon as possible.
Please wait this merge.

ggtakec:
Changed multipart_size option minimum value because multipart minimum part size is 5MB.
Then minimum size for multipart_size option should be 5MB.

PR #289, opened 2015-11-03
Print source file in log messages

RobbKistler:
@ggtakec The new logging macros are very good, thank you!

This PR adds the source file so that logs look like:
```
[INF]       curl.cpp:DeleteRequest(2089): [tpath=/testdir2/]
```
Though maybe you want to keep the log lines as short as possible.


andrewgaul:
:+1:

ggtakec:
@RobbKistler thanks.

Issue #290, opened 2015-11-04
File opened with O_TRUNC is not flushed


RobbKistler:
s3fs-fuse @ ad2a4062058db3ebed85dd4d350221e177ab31fe

*Scenerio:*
Open an existing file, non-empty file O_TRUNC and close it (no writes), 

*Expected results:*
File with zero length. (Verified on local ext4 filesystem)

*Actual results:*
The file still contains original content.

RobbKistler:
Created PR with repro test and potential fix.

ggtakec:
@RobbKistler Thanks for your fixed codes.

I made a few changes to your correction.
Only when the file size is needed to change, s3fs calls RowFlush function.

When the open function is called with O_TRUNC, the file size is set Zero as soon as possible.
So it means that s3fs calls REST api(upload zero size object) when calling open with O_TRUNC.
Then you are correct about calls flush, but it does not need evey time.
My changes after your cahnages, If the file size is zero, s3fs does not call flush.

Please check it.
Thanks in advance for your help.

RobbKistler:
Look great!  Thank you.

PR #291, opened 2015-11-04
Issue #290: File opened with O_TRUNC is not flushed

RobbKistler:
Test to repro Issue #290 and a potential fix.

I am not happy with the fix.  O_TRUNC is common, and this adds an extra REST call for every open.

@ggtakec Is there a better way to mark the file as "dirty" so it will be uploaded when FUSE calls flush and release?


ggtakec:
@RobbKistler 
Thanks for this fix, we are not enough care about O_TRUNC.
And I will try to check codes for not call every open.
Regards,

PR #302, opened 2015-11-25
Fix syslog level used by S3FS_PRN_EXIT()

RobbKistler:
Without converting from s3fs log levels to syslog levels, the syslog
ends up being LOG_EMERG which can cause a broadcast message to all
users.


ggtakec:
@RobbKistler Thank you for fixing a bug.

PR #306, opened 2015-11-30
Fix read concurrency to work in parallel count

guymguym:
When the prefetch size is limited to the multipart size, the entire parallel logic of the read flow does not have an opportunity to use parallel get.
This fix increases the read performance significantly over our own s3 on-premise solution.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -1448,7 +1448,7 @@ ssize_t FdEntity::Read(char* bytes, off_t start, size_t size, bool force_load)
     // load size(for prefetch)
     size_t load_size = size;
     if(static_cast<size_t>(start + size) < pagelist.Size()){
-      size_t prefetch_max_size = max(size, static_cast<size_t>(S3fsCurl::GetMultipartSize()));
+      size_t prefetch_max_size = max(size, static_cast<size_t>(S3fsCurl::GetMultipartSize() * S3fsCurl::GetMaxParallelCount()));
 
       if(static_cast<size_t>(start + prefetch_max_size) < pagelist.Size()){
         load_size = prefetch_max_size;


guymguym:
Would love to know if this helps anyone else, and makes sense from the aspect of the default configuration which is parallel_count=5 and multipart_size=10 (MB) so default prefetch size will be 50 MB.

kahing:
just curious, what's your before/after numbers?

guymguym:
Hi @kahing 

Before - 30 MB/sec
After - 100 MB/sec (practically maxing out the 1Gbit network)

The more latency you have between the client and s3 server, the more it will suffer from lack of read concurrency, so it's quite crucial for high-bandwidth-high-latency networks.

kahing:
That's interesting, I ran some benchmark a little while ago in EC2 and I was able to get 100MB/s with the default settings (https://github.com/kahing/goofys/raw/master/bench/bench.png?raw=true). You change makes sense and explains the high variance that I was seeing.

guymguym:
Great. Could be that in AWS the latency between EC2 to S3 is quite low so only noticeable once there's latency hiccups. Also from your benchmarks it seems that s3fs read 1GB in 20 secs so is that actually 50 MB/s?

kahing:
The average of 10 runs was 20 seconds, but the lower bound was 10 seconds

guymguym:
:+1:

guymguym:
Hey @ggtakec would love to get your feedback too. Thanks

ggtakec:
Hi @guymguym, I'm sorry for replying late.
Increasing the size of the prefetch to enable parallel reading at consecutive reading. Consequently it will be faster.
Conversely, if it is not a continuous read, it will become an overhead.
For example, only for the reading of only 4096 bytes, it means that reading the 50MB.
However, I think most clients to use the s3fs requests a continuous readout.

Therefore, we decide to merge your patch.
However, I would want to add to the merged one additional modification.
It is a set of minimum free disk space that is set in the FdManager::SetEnsureFreeDiskSpace().
It will change this minimum size to S3fsCurl::GetMultipartSize() * S3fsCurl::GetMaxParallelCount().

Thank you for your PR.

ggtakec:
I merged #308 for fixing free disk space default value.
Regards,

guymguym:
@ggtakec Thanks! Your comment on sequential vs. random read is accurate.

As an improvement I think it will be fairly easy to detect sequential access by keeping the last few read ranges requested, and check new read requests whether the new range is intersecting or adjacent to the last ranges, and only in this case to set the prefetch size high, but otherwise better fetch only the exact requested range without extending it.

BTW I remember that the linux kernel is supposed to do readahead to make concurrent prefetching and has rather sophisticated sequential read detection. Does readahead work with FUSE, and do you know if this makes any effect?

kahing:
kernel readahead tends to be small and not useful when latency is super high (it's more for disks, not over the network). libfuse has its own readahead (max_readahead) option that you can play with.

guymguym:
Thanks @kahing! I couldn't find any documentation on how to pass it from command line to s3fs. Is it with `-o max_readahead=99999999`? Or is it still not exposed.

ggtakec:
@guymguym @kahing  Thanks about kernel readahead.
The max_readahead option is an optional fuse, you can specify to s3fs startup.
When I try to set max_readahead option for s3fs, I confirmed that fuse is a change in the read size for instructing to s3fs.(Its size was twice the size of max_readahead).
Regards,

PR #307, opened 2015-12-03
Fix pthread portability problem

rockuw:
I compiled s3fs-fuse on my Mac and got the following errors:

    fdcache.cpp:622:38: error: use of undeclared identifier 'PTHREAD_MUTEX_RECURSIVE_NP'
        pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE_NP);   // recursive mutex
                                         ^

This pull request fixes this problem.


ggtakec:
@rockuw Thanks for PR.
Merged codes is good, but I think that we should check symbol(enum) in configure.
I will push changed codes which checks symbol in configure.ac as soon as possible.
regards,

PR #311, opened 2015-12-04
Change error log to debug log in s3fs_read()

RobbKistler:
This is not an error condition, no need to log it as an error.  For instance, `touch some_empty_file; cat some_empty_file` will cause the log to be emitted.


mcellis33:
LGTM

ggtakec:
@RobbKistler Thank you for PR.

PR #313, opened 2015-12-04
fix gitignore

mcellis33:
@ggtakec @RobbKistler these are a few files that are spewed on build.


ggtakec:
@mcellis33 Thank you for PR.

PR #319, opened 2015-12-15
Clean up mount point on errors in s3fs_init()

RobbKistler:
There are error conditions that don't get caught until s3fs_init() runs.  Currently, s3fs_init calls exit().  This leaves the mount point behind and results in "Transport endpoint not connected" errors.  

With this change, s3fs_init()  will exit the fuse loop cleanly.   This allows libfuse to close its session with the kernel module.   The mount point will be unmounted.


mcellis33:
LGTM

ggtakec:
@RobbKistler Thanks.

Issue #320, opened 2015-12-18
fsync does not update stat cache


mcellis33:
When I perform these operations on a file:

1. open
2. write (count > 0)
3. fsync
4. stat

`stat` returns `st_size == 0` even though I wrote more than 0 bytes and synced them.

The offending call stack is

```
FUSE stat handler
s3fs_getattr
check_object_access
get_object_attribute
StatCache::getStatCacheData()->GetStat
```

`GetStat` encounters a stat cache hit, and the stat cache entry has `st_size == 0`.

`s3fs_create` creates the stat cache entry with `st_size == 0`. The root cause of the bug is that `s3fs_fsync` does not update the stat cache to reflect the new size.

Here is a C program to reproduce this. You will need to point `name` at your s3fs-fuse instance. The output should look like

```
/vagrant/testmount/testfile
stat size 0
```

```
#include <fcntl.h>
#include <string.h>
#include <errno.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <sys/stat.h>

const char* test_content = "test-content";

int try(char* name) {
	int ret;
	int fd = open(name, O_RDWR|O_CREAT|O_EXCL, 0600);
	if (fd == -1) {
		ret = errno;
		perror("open");
		return ret;
	}

	// Write to file.
	ret = write(fd, test_content, 12);
	if (ret == -1) {
		ret = errno;
		perror("write");
		close(fd);
		return ret;
	}

	// Sync to ensure content is written.
	ret = fsync(fd);
	if (ret == -1) {
		ret = errno;
		perror("sync");
		close(fd);
		return ret;
	}

	// Check that stat is 12.
	struct stat st;
	ret = stat(name, &st);
	if (ret == -1) {
		ret = errno;
		perror("stat");
		return ret;
	}
	printf("stat size %ld\n", st.st_size);

	close(fd);

	return 0;
}

int main() {
	char* name = "/vagrant/testmount/testfile";
	printf("%s\n", name);
	return try(name);
}
```

ggtakec:
@mcellis33 Thans for fixing this bug.

PR #321, opened 2015-12-18
320: delete stat cache entry in s3fs_fsync so st_size is refreshed

mcellis33:
@ggtakec 

My proposed solution to #320 is for `s3fs_fsync` to delete the stat cache entry. Then, any subsequent operations will need to refresh the stat cache with a HEAD request. This will negatively impact performance in that it will cause more round trips. However, this is the simplest way to make `s3fs_fsync` correct. Someone who understands more about how the stat cache works could try to surgically update the stat cache to reflect the new file size in `s3fs_fsync`.

With this change, the test program from the #320 description outputs

```
/vagrant/testmount/testfile
stat size 12
```

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -2169,6 +2169,9 @@ static int s3fs_fsync(const char* path, int datasync, struct fuse_file_info* fi)
   }
   S3FS_MALLOCTRIM(0);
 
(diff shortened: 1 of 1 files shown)


mcellis33:
@RobbKistler

RobbKistler:
LGTM

PR #331, opened 2016-01-11
Address various clang warnings

andrewgaul:
Found with:

-Wc++11-extensions
-Wc++11-extra-semi
-Wmissing-variable-declarations
-Wundef


ggtakec:
Thanks!

PR #336, opened 2016-01-19
update REAME.md for fstab

Jirapong:
I've make a modify to allow s3fs start on boot with fstab (tested on Ubuntu 14)


ggtakec:
@Jirapong Thank you.

Issue #340, opened 2016-01-25
Memory Leak - s3fs using over 12GB memory


justinfalk:
Hi,

First, thank you for the great product.  I've really enjoyed working with it thus far.

We have s3fs running on an AWS instance connected to a bucket with ~110 million small objects.  No single directory contains more than a few hundred files and our application accesses these files by their full path, there should be no traversing the directory structure or anything like that.  s3fs is used only for retrieval of these objects (no new objects are written) and occasional deletion.  On an average night, there are only a few thousand downloads (say 5-20k objects downloaded).  I am not using a disk cache because objects are rarely downloaded more than once.

After running for about a week, the kernel killed s3fs because it was consuming all available system memory (~12GB).  See logs below.  

I've searched and found other memory leak issues related to old libcurl versions and non-ssl usage but those don't seem to apply to my environment.  I'm not certain if this is a configuration error or a memory leak, but it seems to be a memory leak because I can easily reproduce this by simply downloading files via my application and watching s3fs memory use grow.  Please let me know if there is any other information that might be helpful.  

_Environment:_
OS:  Amazon Linux 4.1.13-18.26
s3fs:  1.79
Fuse:  2.9.4-1.17
libcurl:  7.40.0-3.52

_fstab config:_

```
defaults,noatime,allow_other,umask=022,url=https://s3.amazonaws.com,use_sse,_netdev
```

Log snippet  (full /var/log/messages output attached)

```
Jan 25 02:09:42 ip-10-2-0-77 kernel: [3926969.438013] Out of memory: Kill process 18320 (s3fs) score 686 or sacrifice child
Jan 25 02:09:42 ip-10-2-0-77 kernel: [3926969.441504] Killed process 18320 (s3fs) total-vm:12493224kB, anon-rss:11604912kB, file-rss:0kB
```

[oom.txt](https://github.com/s3fs-fuse/s3fs-fuse/files/103222/oom.txt)

hryang:
Hi, 

I met the same problem. In my scenario, the memory leak is due to fdcache.cpp:325, where the page is erased from the list, but not be deleted. 

Also, the code may have other logic error. Please see my comments in the following piece of code.

```
for(fdpage_list_t::iterator iter = pages.begin(); iter != pages.end(); ){
      if(static_cast<size_t>((*iter)->next()) <= size){
        ++iter;
      }else{
        if(size <= static_cast<size_t>((*iter)->offset)){
[...]

PR #341, opened 2016-01-28
Fix the memory leak issue in fdcache.

hryang:


RobbKistler:
This may not resolve Issue #340 but it does fix a memory leak.  LGTM

ggtakec:
@hryang Thanks and sorry for replying late.

PR #346, opened 2016-02-05
Fix empty directory check against AWS S3

### LINKED ISSUE #345: Can rmdir a non-empty directory running against AWS S3
The integration test `test_remove_nonempty_directory` fails if s3fs-fuse is running against Amazon AWS.  It passes against S3Proxy 1.4.0.  See https://github.com/andrewgaul/s3proxy/issues/119

RobbKistler:
For ListBucketResult on an empty directory, AWS S3 and S3Proxy 1.4
differ.  AWS will match the directory name, S3Proxy does not.

Changing max-keys=1 to max-keys-2 works for both implementations.
append_objects_from_xml() will swallow the directory key.  The log
level of this message is changed from ERROR to DBG.

Fixes #345

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -2429,7 +2429,9 @@ static int list_bucket(const char* path, S3ObjList& head, const char* delimiter,
     query_prefix += urlEncode(s3_realpath.substr(1));
   }
   if (check_content_only){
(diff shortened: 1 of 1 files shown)


ggtakec:
Thanks for your great help.

PR #348, opened 2016-02-05
Integration test summary, continue on error

RobbKistler:
Details in README.md and s3fs-integration-test-common.sh

Factor out s3fs-fuse and s3proxy start/stop.  The plan is to make it easier to
add test suites besides small-integration-test.sh that can test with various
s3fs options.

Each test run starts in a uniquely named at the top of the bucket.  This allows
multiple runs against persistent storage without worrying about cleaning
up in error conditions that leave artifiacts behind.

Tests continues if a test case fails.

Results are summarized at the end of the test run

Environment variable to control debug level of s3fs-fuse

Environment variable to enable public bucket (makes it easier to poke
around with tools like curl)

Environment variable to start s3fs-fuse under valgrind

Environment variable that casues script sets up s3fs-fuse and then wait
indefinitely, making it easy to experiment manually with the mount
point.

Additional test case

### DIFF
--- a/test/integration-test-common.sh
+++ b/test/integration-test-common.sh
@@ -1,11 +1,57 @@
-#!/bin/bash -e
-set -x
+#!/bin/bash
+
+#
+# Common code for starting an s3fs-fuse mountpoint and an S3Proxy instance 
+# to run tests against S3Proxy locally.
+#
+# To run against an Amazon S3 or other S3 provider, specify the following 
+# environment variables:
+#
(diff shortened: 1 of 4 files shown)


RobbKistler:
I'd like to start adding more tests.  These changes have helped me a lot.  Sorry for such a huge PR.

ggtakec:
No problem, on the contrary I really appreciate what you've done.

Issue #347, opened 2016-02-05
Disk space error for other things


will-parsons:
```
# s3fs -o noatime,use_cache=/tmp/s3fs_cache BUCKET-NAME /mnt/s3fs/
s3fs: There is no enough disk space for used as cache(or temporary) directory by s3fs.
```
I had plenty of disk space. In this case, ```/tmp/s3fs_cache``` didn't exist. When I created it, the mount worked fine:

```
# mkdir /tmp/s3fs_cache
# s3fs -o noatime,use_cache=/tmp/s3fs_cache BUCKET-NAME /mnt/s3fs/
# tail -1 /proc/mounts 
s3fs /mnt/s3fs fuse.s3fs rw,nosuid,nodev,noatime,user_id=0,group_id=0 0 0
```

Additionally, my first attempt gave a similar error because SElinux was enabled and, rightly so, s3fs wasn't allowed to write to /tmp/ for the cache.  Disabled SElinux and the mount worked as above. 

Suggestions: 

- [ ] Check ```use_cache``` directory exists when mounting, and either create it or throw a meaningful error
- [ ] Check we can get a write lock on the ```use_cache``` directory before attempting to use it

Thanks :)

ggtakec:
@will-parsons I merged new codes #351.
It throw a error at executing s3fs. so I think that s3fs should not make cache directory, it should be existed before executing s3fs.

ggtakec:
@will-parsons 
This problem was fixed by #351, but this fix causes a problem at OS startup like #538.
Therefore, check_cache_dir_exist option was added.
To check the cache directory when starting s3fs, specify this new option together.

It will be revised again, and will inconvenience you.
Regards,

PR #351, opened 2016-02-06
Check cache dirctory path and attributes - #347

ggtakec:
Added checking cache directory and it's attributes in parsing parameter function at s3fs starting.


ggtakec:
This is refixed by #557

PR #360, opened 2016-02-08
Fix clock_gettime autotools detection on Linux

RobbKistler:
On Debian 7.8, AC_CHECK_FUNCS([glock_gettime]) returns false without adding rt to the search libs.


ggtakec:
@RobbKistler thank you!

ggtakec:
@RobbKistler I'm sorry for that I will change the codes.
#361 is due to gcc version on ubuntu, so I think s3fs should use gettimeofday.
Regards,

Issue #375, opened 2016-03-16
Can't write after lseek()


RobbKistler:
**Version**: 
s3fs-fuse master at cf56b35766c4450c789968335ffbc01258cbbdb8

**Repro:**
As part of a bigger test (`cp -R /usr/lib abucket/`), I found that copying a certain file into an s3fs-fuse mount point fails: 
```
cp  /usr/lib/locale/locale-archive abucket/bug/
```
The cp always fails on the file locale-archive because the source file is [sparse](https://en.wikipedia.org/wiki/Sparse_file#Sparse_files_in_Unix).  It has "holes" in it because it was written using lseek() to advance past the end of the file.  The cp command is smart, and tries to do the same lseek() when writing the file into the s3fs-fuse mount point.  Since this file might not be sparse on all systems, I will create a PR with a test that uses lseek() directly.

**Results:**
The file is partially written, and then `cp` gets this error:
```
cp: writing `abucket/bug/locale-archive': Operation not permitted
```
while s3fs-fuse reports this error:
```
s3fs: [DBG] s3fs.cpp:s3fs_write(2167): [path=/bug/locale-archive][size=4096][offset=61440][fd=4]
s3fs: [DBG] fdcache.cpp:ExistOpen(1896): [path=/bug/locale-archive][fd=4][ignore_existfd=false]
s3fs: [DBG] fdcache.cpp:Open(1845): [path=/bug/locale-archive][size=-1][time=-1]
s3fs: [DBG] fdcache.cpp:Dup(712): [path=/bug/locale-archive][fd=4][refcnt=2]
s3fs: [DBG] fdcache.cpp:Write(1503): [path=/bug/locale-archive][fd=4][offset=61440][size=4096]
s3fs: [DBG] fdcache.cpp:Load(1027): [path=/bug/locale-archive][fd=4][offset=0][size=61440]
s3fs: [INF]       curl.cpp:GetObjectRequest(2632): [tpath=/bug/locale-archive][start=12288][size=0]
s3fs: [DBG] s3fs.cpp:get_object_attribute(417): [path=/bug/locale-archive]
s3fs: [DBG] cache.cpp:GetStat(259): stat cache hit [path=/bug/locale-archive][time=176085.960913293][hit count=5]
s3fs: [INF]       curl.cpp:PreGetObjectRequest(2562): [tpath=/bug/locale-archive][start=12288][size=0]
[...]

Issue #15, opened 2014-02-21
Cache files are blank


Matan:
Please look at this binary comparison image:
![image-differences](https://f.cloud.github.com/assets/184429/2228232/8b30ea2a-9ad7-11e3-88de-3329d439147f.JPG)
On the left we have the source image file downloaded from S3 console and on the right we have the cache file created by the s3fs mount.

Is this the expected behaviour? I noticed that the cache doesn't make a speed difference as the file is being downloaded from S3 each time. I tested latency to my web server in a couple of ways:
* Direct image load (no s3 mount): ~200 latency
* Direct image load (with mount, no cache): ~1000 latency
* Direct image load (with mount, with cache): ~1000 latency

This lead me to believe that the cache isn't working properly, thus I did the binary comparison and found the bytes of the source vs cache doesn't match up.

Urgent help would be much appreciated.

Thanks,
Matan

ggtakec:
I'm sorry for that this issue had been left at a long period of time.

If you have this problem yet, please try to use latest code.

Thanks in advance for your help.

Morc001:
Hi Takeshi,

  This problem of files containing only zeros (empty sparse files?) is still happening on master as of today. Files are small, a few hundred bytes only, so no multi-part. I'm just mounting a bucket and viewing files, nothing fancy. I'm using eu-central-1 endpoint if it makes any difference. In your last comment you wrote it's fixed or has that referred to some other problem?
  Using 1.79 the problem does not show, but that version lacks ensure_diskfree.

davrodfer:
I had the same issue. I wrote my files from two different machines with the same role. When I read a file from a machine that didn't wrote the file, all I have is a bunch of zeroes. Every machine only can read the files that they wrote, but the files are Ok in S3.

Y had this configuration in /etc/fstab:
```
s3fs#s3bucketname  /mnt/s3fs/s3bucketname     fuse    rw,nosuid,nodev,allow_other,uid=48,gid=48,use_cache=/var/tmp/s3fs,default_acl=public-read-write,storage_class=standard,_netdev,use_rrs,iam_role=role_s3bucketname 0 0
```
I changed to:
```
s3fs#s3bucketname  /mnt/s3fs/s3bucketname     fuse    noatime,rw,nosuid,nodev,allow_other,uid=48,gid=48,default_acl=public-read-write,storage_class=standard,_netdev,iam_role=role_s3bucketname 0 0
```
[...]

PR #394, opened 2016-04-12
Revert "Fixed a bug about writing sparsed file - #375"

ggtakec:
Reverts s3fs-fuse/s3fs-fuse#379

For #15 and other issue, this commit merge is broken cache file.
So I revert this.


ggtakec:
This pr did not pass travis ci, because now test case has spare file.
but we need to revert assap.

PR #403, opened 2016-04-22
Fix a bug of truncating empty file

rockuw:
The following steps report error, which should NOT:

```bash
ubuntu@ip-172-31-29-108:~/s3fs-fuse$ touch /tmp/s3fs/a
ubuntu@ip-172-31-29-108:~/s3fs-fuse$ ll /tmp/s3fs/a
-rw-rw-r-- 1 ubuntu ubuntu 0 Apr 22 06:48 /tmp/s3fs/a
ubuntu@ip-172-31-29-108:~/s3fs-fuse$ truncate /tmp/s3fs/a -s 1024
truncate: failed to truncate ‘/tmp/s3fs/a’ at 1024 bytes: Operation not permitted
```

This patch fixes this bug.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -41,6 +41,24 @@ function test_truncate_file {
     rm_test_file
 }
 
+function test_truncate_empty_file {
+    echo "Testing truncate empty file ..."
+    # Write an empty test file
+    touch ${TEST_TEXT_FILE}
+
+    # Truncate the file to 1024 length
+    t_size=1024
+    truncate ${TEST_TEXT_FILE} -s $t_size
+
+    # Verify file is zero length
+    size=$(stat -c %s ${TEST_TEXT_FILE})
+    if [ $t_size -ne $size ]
+    then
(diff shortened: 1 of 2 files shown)


test/integration-test-main.sh
```diff
@@ -41,6 +41,24 @@ function test_truncate_file {
...
+    # Truncate the file to 1024 length
+    t_size=1024
+    truncate ${TEST_TEXT_FILE} -s $t_size
+
+    # Verify file is zero length
+    size=$(stat -c %s ${TEST_TEXT_FILE})
+    if [ $t_size -ne $size ]
+    then
+        echo "error: expected ${TEST_TEXT_FILE} to be $t_size length, got $size"
+        exit 1
```
RobbKistler: Please `return 1` here so other tests continue to run on failure

rockuw: done

rockuw:
@ggtakec

ggtakec:
@rockuw I'm sorry for late reply.
Thanks for fixing a bug.

PR #404, opened 2016-04-22
Add curl handler pool to reuse connections

rockuw:
Currently s3fs creates a new tcp connection for each request to s3. Together with the fact that curl(or HTTP) defaults to use 'Connection:keep-alive', the client may ends with lots of 'TIME_WAIT' connections:

```
ubuntu@ip-172-31-29-108:~/s3fs-fuse$ for i in $(seq 100); do ls /tmp/s3fs > /dev/null; done
ubuntu@ip-172-31-29-108:~/s3fs-fuse$ sudo ss -tan | grep TIME | wc -l
101
```

This patches enables reusing curl handlers and thus utilizes HTTP keep-alive, which has the effects:

1. boost performance
2. largely reduce TIME_WAIT connections

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -221,6 +221,88 @@ const char* BodyData::str(void) const
 }
 
 //-------------------------------------------------------------------
+// Class CurlHandlerPool
+//-------------------------------------------------------------------
+
+bool CurlHandlerPool::Init()
+{
+  if (0 != pthread_mutex_init(&mLock, NULL)) {
+    S3FS_PRN_ERR("Init curl handlers lock failed");
+    return false;
+  }
+
+  mHandlers = new CURL*[mMaxHandlers](); // this will init the array to 0
+  for (int i = 0; i < mMaxHandlers; ++i, ++mIndex) {
+    mHandlers[i] = curl_easy_init();
+    if (!mHandlers[i]) {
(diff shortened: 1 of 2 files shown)


RobbKistler:
Does AWS S3 actually respect keep-alive? I thought it didn't.  Other S3 providers might.

rockuw:
@RobbKistler Why shouldn't it? Since s3 provides RESTful API, which implies HTTP, which implies 'Connection' header.

rockuw:
@RobbKistler And the 101 connections of 'TIME_WAIT' is an indication that s3 respects 'keep-alive' and **does not close the connection** after the HTTP request. The client then closes it and enters 'TIME_WAIT'. (Only the part who initiates FIN enters this state).

RobbKistler:
@rockuw I was recalling information in this thread:

https://forums.aws.amazon.com/thread.jspa?threadID=12062

But the thread is very old and probably outdated.

rockuw:
@RobbKistler As  John Cormie(from S3) says:

> This should be fixed now. Sorry about the inconvenience.

rockuw:
@ggtakec

ggtakec:
@rockuw @RobbKistler 
If s3fs can use the KeepAlive, I think it will give a good performance and reduce the resources.
Thus I'm going to merge this PR, and to test it.
If you found bugs about KeepAlive, please post issue.

Issue #419, opened 2016-05-20
"s3fs: could not determine how to establish security credentials" when using iam_role


nturner:
I have an /etc/fstab entry like the following:

`s3fs#mybucket       /srv/object-store       fuse    _netdev,dbglevel=info,iam_role  0       2`

When I try to mount /srv/object-store, s3fs fails with "s3fs: could not determine how to establish security credentials."

The log in /var/log/messages shows only:

`May 20 16:06:23 ip-172-31-20-21 s3fs[2015]: s3fs.cpp:set_s3fs_log_level(280): change debug level from [CRT] to [INF] `
`May 20 16:06:23 ip-172-31-20-21 s3fs[2015]:     s3fs.cpp:set_moutpoint_attribute(4298): PROC(uid=0, gid=0) - MountPoint(uid=0, gid=0, mode=40755)`
`May 20 16:06:23 ip-172-31-20-21 s3fs[2015]: s3fs: could not determine how to establish security credentials.
`

nturner:
The patch linked above may need some work.

ggtakec:
@nturner Thanks for your help.
#420 and #422 is merged.

PR #420, opened 2016-05-20
Skip early credential checks when iam_role=auto

nturner:
If user specifies iam_role=auto (or just iam_role), credentials will not
be loaded during early phase, so skip credential checks there.

Bug: https://github.com/s3fs-fuse/s3fs-fuse/issues/419
Signed-off-by: Nathaniel W. Turner <nate@houseofnate.net>


ggtakec:
@nturner Thanks!

Issue #421, opened 2016-05-24
s3fs with iam_role uses Profile name where it should use Role name


nturner:
When using an instance with an IAM Role, transient credentials can be found in http://169.254.169.254/latest/meta-data/ at iam/security-credentials/_role-name_ and s3fs tries to do this. However, it is using the _profile-name_ where _role-name_ is needed. In many cases the role and profile name are the same, but they are not always. 

The simplest way to find the role name appears to be to GET http://169.254.169.254/latest/meta-data/iam/security-credentials/ itself, which returns a listing of the role names for which temporary credentials exist. (I think there will probably only be one, but we probably want to split on newlines and take the first one here in case that assumption is not valid). This is the approach the AWS SDK appears to use (based on WireShark analysis).

ggtakec:
@nturner Thanks for your help.
The process getting iam role name has become very simple.:-)
I merged #420 and #422 to master branch.

PR #422, opened 2016-05-24
Fixes for iam_role=auto

nturner:
Patches for several issues:
https://github.com/s3fs-fuse/s3fs-fuse/issues/419
https://github.com/s3fs-fuse/s3fs-fuse/issues/421


ggtakec:
@nturner Thanks again.

PR #432, opened 2016-06-13
Include location constraint when creating bucket

andrewgaul:
This allows creating buckets in non-default regions.  Also improve
`do_create_bucket` error handling.


ggtakec:
@andrewgaul Thanks and Sorry for my late reply.

PR #433, opened 2016-06-13
Correct search and replace typo

andrewgaul:


ggtakec:
@andrewgaul Thank you.

PR #443, opened 2016-06-30
Fix for leaks during stat cache entry expiry / truncation (#340)

tlevi:
As discussed in #340 - I've run this for a few days in production, and under an intense test load and it has resolved valgrind reported leaks, allocated from AddStat.


ggtakec:
@tlevi thanks for your PR.
It seems like good to fix leak for me.
I merged this patch and if you can please let us know about the result of your test.
Regards,

PR #449, opened 2016-07-13
Accept mount options compatible with mtab

treestem:
Fixes an issue with "mount -a" failing for an already-mounted s3fs file system, on linux.


ggtakec:
@treestem Thanks for your patch.
I'm merging it and changing some codes(man page, -h command etc) after that.
Regards,

PR #451, opened 2016-07-14
Correct path in README

kasbah:
I think you meant this?


andrewgaul:
This original path is intended; users can have a passwd file in /etc or anywhere on their file system.

kasbah:
But it should match the path of the command just before, no?

kasbah:
Also, wouldn't `chmod 640` be better? That would match what it says in the man page as well.

ggtakec:
@kasbah Thanks

PR #486, opened 2016-10-21
Fix SSE_C and SSE_KMS that return ETag that is not object MD5

driskell:
This PR fixes EIO error when uploading parts with SSE_KMS (and probably SSE_C too) due to the ETag response from S3 not actually being the MD5 of the data. It disables the ETag check when SSE_KMS or SSE_C is in use, and it also uses the S3 returned ETag in CompleteMultiPartUpload otherwise you get a 400 error with InvalidPart.

Also, instead of using a `headerdata` string this now makes user of the existing header parser. It has to do this in order to parse out the returned ETag since it is different to the one we sent.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -1138,10 +1138,20 @@ bool S3fsCurl::UploadMultipartPostCallback(S3fsCurl* s3fscurl)
     return false;
   }
   // check etag(md5);
-  if(NULL == strstr(s3fscurl->headdata->str(), s3fscurl->partdata.etag.c_str())){
+  headers_t::const_iterator iter = s3fscurl->responseHeaders.find("ETag");
+  if (iter == s3fscurl->responseHeaders.end()) {
     return false;
   }
-  s3fscurl->partdata.etaglist->at(s3fscurl->partdata.etagpos).assign(s3fscurl->partdata.etag);
+  // The ETAG when using SSE_C and SSE_KMS does not reflect the MD5 we sent
+  // SSE_C: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html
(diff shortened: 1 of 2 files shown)


driskell:
Checks failed. I think because of the Dyn outages? Let me know if you need any further adjustments here.

One thing I did concern about was no way to validate transfer except by enabling `-oenable_content_md5` - but I believe with SSE_KMS forcing HTTPS (not sure about SSE_C) that's probably not required as the channel itself is validated from corruption.

driskell:
Hey @ggtakec - Do you need anything from me to get this merged? Thanks!

ggtakec:
@driskell Thanks for sending your patches.

I tried to rebuild this PR codes on travis, and got two error.
Which mean test_multipart_upload and test_multipart_copy failed.

I tested my local box, and I knew following errors.
```
[ERR] curl.cpp:CompleteMultipartPostRequest(3100): 1 file part is not finished uploading.
```
It seems that s3fs called multipart complete before multipart uploading did not finished all parts.
Please check your PR codes.

About DEBUGFUNCTION, I think that this flag is set only curldbg option specified, thus I think it is better to leave this as it is.
But what about making your suggested code work when dbglevel=dbg(using S3FS_PRN_DBG macro)?

ggtakec:
@driskell I'm sorry about this PR.
I have fixed about this issue by another PR(#585) and issue(#584).
Because this PR had not been fixed to test failure, and this code was also conflicting with the master code since considerably time passed since being PR.
And I found another bug about rename function was failed.
Then I made new issue(#584) and fixed it from your codes base.

So that, I closed this PR, BUT I appreciate your cooperation.
Regards,

driskell:
Thanks @ggtakec, I'm sorry I didn't find the extra time to look at it again, but it's brilliant you were able to fix it. Thanks!

Issue #480, opened 2016-10-05
Content-Length is searched for in case sensitive manner


jdanders:
RFC 7230 states: "Each header field consists of a **case-insensitive** field name followed by a colon (":")"

If the HTTP header coming back from the S3 server has 'content-length:' in it instead of 'Content-Length:' then s3fs fails to find the size and reports the file has 0 length.

I believe the problem code is src/s3fs_util.cpp line [721](https://github.com/s3fs-fuse/s3fs-fuse/blob/1a93897e8510a618686510dd25bb8bf033fbfd69/src/s3fs_util.cpp#L721), where a case-sensitive find is performed.

ggtakec:
@jdanders Thanks for your help.
I fixed this issue by #488(#489), please try to check latest codes.
Regards,

PR #488, opened 2016-10-23
Fixed searching Content-Length without case sensitive - #480

ggtakec:
s3fs has changed the Content-Length(and Last-Modified) header to search not case-sensitive.
This change corresponds to the #480.

### DIFF
--- a/src/s3fs_util.cpp
+++ b/src/s3fs_util.cpp
@@ -60,18 +60,24 @@ string get_realpath(const char *path) {
   return realpath;
 }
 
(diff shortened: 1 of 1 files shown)


andrewgaul:
@ggtakec Would it be more straightforward to change the `headers_t` typedef to a case-insensitive map:

https://stackoverflow.com/questions/1801892/making-mapfind-operation-case-insensitive

ggtakec:
I'm going to merge codes which is using map compare function.(#489)
Regards,

PR #494, opened 2016-11-11
Fix typo s/destroied/destroyed/

mapreri:


src/curl.cpp
```diff
@@ -1622,7 +1622,7 @@ bool S3fsCurl::CreateCurlHandle(bool force)
       S3FS_PRN_ERR("could not destroy handle.");
       return false;
     }
-    S3FS_PRN_INFO3("already has handle, so destroied it.");
+    S3FS_PRN_INFO3("already has handle, so destroyed it.");
```
andrewgaul: Should this be "destroy"?

mapreri: Dunno, I just followed the original tense.

driskell: Since the destruction occurred before the message, I would say keep the existing tense, `destroyed`.

ggtakec: Thanks all, I'm sorry for my english.

PR #501, opened 2016-11-19
Fixed issue#435 branch codes for remaining bugs

ggtakec:
- Fixed the remaining defects


ggtakec:
Sorry for my fault.
This patch is not fixed all codes.
Thus close assap.

PR #509, opened 2016-11-24
Use server-provided ETag during complete upload

andrewgaul:
This avoids calculating the MD5 locally and enables use with object
stores which do not use MD5 as ETag.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -127,6 +127,17 @@ static string get_bucket_host()
   return url_to_host(host);
 }
 
+// compare ETag ignoring quotes
+static bool etag_equals(std::string s1, std::string s2) {
+  if(s1.length() > 1 && s1[0] == '\"' && s1[s1.length() - 1] == '\"'){
+	s1 = s1.substr(1, s1.size() - 2);
+  }
+  if(s2.length() > 1 && s2[0] == '\"' && s2[s2.length() - 1] == '\"'){
+	s2 = s2.substr(1, s2.size() - 2);
+  }
+  return s1 == s2;
+}
+
 #if 0 // noused
 static string tolower_header_name(const char* head)
 {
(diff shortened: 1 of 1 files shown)


andrewgaul:
I found another dodgy `strstr` call in `S3fsCurl::UploadMultipartPostRequest`.  How does this get called?

ggtakec:

I think that we should change the logic in UploadMultipartPostRequest as following after you change the callback for CURLOPT_HEADERFUNCTION.
This codes is as same as codes in your UploadMultipartPostCallback.

```
int S3fsCurl::UploadMultipartPostRequest(const char* tpath, int part_num, const string& upload_id)
{
  int result;

  S3FS_PRN_INFO3("[tpath=%s][start=%jd][size=%zd][part=%d]", SAFESTRPTR(tpath), (intmax_t)(partdata.startpos), partdata.size, part_num);

  // setup
  if(0 != (result = S3fsCurl::UploadMultipartPostSetup(tpath, part_num, upload_id))){
    return result;
  }

  // request
  if(0 == (result = RequestPerform())){
    headers_t::iterator it = responseHeaders.find("ETag");
    if (it == responseHeaders.end()) {
      result = -1;
    }else{
      // check etag(md5);
      if(S3fsCurl::is_content_md5 && !etag_equals(it->second, partdata.etag)){
        result = -1;
      }else{
        partdata.etaglist->at(partdata.etagpos).assign(it->second);
        partdata.uploaded = true;
      }
    }
  }

  // closing
  delete bodydata;
  bodydata = NULL;
  delete headdata;
  headdata = NULL;

  return result;
}
```

And the headdata buffer is not used no longer, thus we do not need to delete it and new it in UploadMultipartPostSetup, it should be initialized NULL.
(this can be fixed after this PR.)

andrewgaul:
@ggtakec Agree that `S3fsCurl::UploadMultipartPostRequest` should change as you suggest.  But how to I test this code path?

ggtakec:
@andrewgaul 
I think it is hard to test about S3fsCurl::UploadMultipartPostRequest method.

This method is called from FdEntity::NoCacheLoadAndPost, and NoCacheLoadAndPost is called at the following special state.
It happens when you upload a file larger than the available local disk size of s3fs.
Available disk space for s3fs is the free space minus the size by ensure_diskfree(MB) option.

Then we need to run s3fs with ensure_diskfree option and upload file which is over aailable size for s3fs.

Example)
```
$ df -k
Filesystem        1K-blocks    Used    Available Use% Mounted on
/dev/xvda1          8125880 2810936      5214704  36% /

$ s3fs mybucket /mnt/s3 -o allow_other,ensure_diskfree=5000 -f &

$ ls -l bigfile
-rw-r--r-- 1 test users 660000000 Aug 13 06:47 bigfile

$ cp bigfile /mnt/s3
```

I think this is very difficult to test.
Do you have any idea about this?

Regards,

ggtakec:
@andrewgaul Because it is difficult to create a test case, I will merge as it is.

ggtakec:
Because there was a missing fix, I fixed it in #560.

Issue #435, opened 2016-06-15
Some files partially contain NUL content after uploading via S3FS (file size matches)


ClemensSchneider:
Some files (currently 8 out of ~31500 files) do contain only NUL characters when checking their contents in the bucket. The filesize exactly matches the source-file but the file contents obviously not.
As it seems to occur only in rare cases it is hard to provide any steps to reproduce the behavior..

Are there any cases where S3FS fills a file with NUL before processing it further?

Using S3FS 1.80, filesizes range around 10 - 60MB.

ClemensSchneider:
After some more investigation I found some files that do partially contain correct content but then have a block of NUL content following with correct content afterwards again... How could this error-pattern occur? It leaves me with a bad feeling about the consitency of the data being written with S3FS.

andrewgaul:
@ClemensSchneider Can you verify whether the file failed to upload or download correctly using another tool?  s3fs has had other bugs around network errors which we have fixed but another alarms me as well.  I wrote a tool [Chaos HTTP Proxy](https://github.com/bouncestorage/chaos-http-proxy) which helped diagnose these and encourage you to test with it as well.

ClemensSchneider:
@andrewgaul I verified that the files failed to upload correctly as I downloaded them using both S3-Browser and Dragondisk. In our case these files are FLAC files we are syncing to S3 using BitTorrent Sync. We noticed the data-corruption as our tool that is running in the cloud and is picking up changed metadata in the files complained about invalid FLAC-metadata. 
It seems that an initial sync of the files worked correctly and following modifications of the metadata or audio-content leave some of the FLACs corrupted.

The bucket is mounted via /etc/fstab:
`bucketname /mnt/s3 fuse.s3fs _netdev,url=http://s3.amazonaws.com,enable_noobj_cache,max_stat_cache_size=50000,stat_cache_expire=300,retries=10,multireq_max=50,allow_other,rw,nonempty,uid=1000,gid=1000 0 0`

ClemensSchneider:
@ggtakec Seems like writing only parts of a file leaves the file corrupt. Should S3FS be able to handle this case? I would expect S3FS to correctly apply the changes and then re-upload the whole file to S3.

I tried to enable MD5 checksum tests and to disable multipart uploads without any success:

[...]

PR #511, opened 2016-12-04
Fixed a bug about uploading NULL to some part of the file contents

ggtakec:
When s3fs performs continuous upload, part of the file can not be uploaded normally and some part of the file was written as a NULL code.
(See the issue of #435, and there are other issues related to this bug.)
After fixing this bug to another branch, it seems to be stable, so I try to merge it into the master branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ cache: apt
 before_install:
  - sudo apt-get update -qq
  - sudo apt-get install -qq cppcheck libfuse-dev openjdk-7-jdk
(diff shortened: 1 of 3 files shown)


ggtakec:
@andrewgaul 
It seems that TravisCI's java environment has changed, and Oracle JDK 9 is the default.
Therefore, I am modifying .travis.yaml a bit.
This change seems not to be a very good setting, so please contact me if you have an opinion.

Thanks for your help.

PR #517, opened 2016-12-16
Update s3fs.1 - removed duplicated word

hudsantos:
removed duplicated word


ggtakec:
@hudsantos Thank you very much.

PR #550, opened 2017-03-29
Add umount instruction for unplivileged user.

Vascom:
#### Details
Ordinary user has no permissions to use umount command, so it is good to add fuse unmount instruction in man file.


ggtakec:
@Vascom Thank a lot.

PR #552, opened 2017-04-02
switch S3fsMultiCurl to use foreground threads

orozery:
#### Details
Currently, s3fs uses a multicurl handle to use parallel connections to s3.
The connections are indeed parallel, but are handled by a single CPU thread.
This creates a CPU bottleneck when reaching high throughput with TLS (HTTPS) connections.
To solve this issue, I propose a change to the S3fsMultiCurl class to use foreground pthreads instead of a multicurl handle.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -3754,7 +3754,7 @@ int S3fsMultiCurl::SetMaxMultiRequest(int max)
 //-------------------------------------------------------------------
 // method for S3fsMultiCurl 
 //-------------------------------------------------------------------
-S3fsMultiCurl::S3fsMultiCurl() : hMulti(NULL), SuccessCallback(NULL), RetryCallback(NULL)
+S3fsMultiCurl::S3fsMultiCurl() : SuccessCallback(NULL), RetryCallback(NULL)
 {
 }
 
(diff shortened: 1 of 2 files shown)


src/curl.cpp
```diff
@@ -3827,189 +3819,126 @@ bool S3fsMultiCurl::SetS3fsCurlObject(S3fsCurl* s3fscurl)
...
-  int       still_running;
+  std::vector<pthread_t>   threads;
+  bool                     success = true;
+
+  for(s3fscurlmap_t::iterator iter = cMap_req.begin(); iter != cMap_req.end(); ++iter) {
+    pthread_t   thread;
+    S3fsCurl*   s3fscurl = (*iter).second;
+    int         rc;
+
+    rc = pthread_create(&thread, NULL, RequestPerformWrapper, static_cast<void*>(s3fscurl));
```
andrewgaul: What is the maximum number of threads this can spawn?  If many, can we configure this somehow?  Some people use s3fs on limited memory systems.

orozery: This number equals the "parallel_count" argument (which defaults to 5).
Threads will be created only when handling files which are larger than "multipart_size" (defaults to 10MB).

ggtakec: I think that there is no problem with that option.

ggtakec:
I think that there is no problem other than the definition of the RequestPerformWrapper method.

orozery:
@ggtakec RequestPerformWrapper is defined as a private method of S3fsCurl, not public.
You are still suggesting to move it from S3fsCurl to S3fsMultiCurl?
I think it makes more sense to leave it as part of S3fsCurl since it works on a S3fsCurl object (not S3fsMultiCurl).
However, I'm also OK with moving it to S3fsMultiCurl, as you suggested.
Your thoughts?

ggtakec:
@orozery thanks for replying.
You defined RequestPerformWrapper method in S3fsCurl class in header file.
But implement RequestPerformWrapper in cpp file is global scope.
```
void* RequestPerformWrapper(void* arg)
```
I think that at first it should be following:
```
void* S3fsCurl::RequestPerformWrapper(void* arg)
```

After changing, but if you changed it to "S3fsCurl::RequestPerformWrapper", I think you will get a error at calling it.
```
rc = pthread_create(&thread, NULL, RequestPerformWrapper, static_cast<void*>(s3fscurl));
```
(This arguments for pthread_create function should be changed S3fsCurl::RequestPerformWrapper, thus maybe it will put a error at compliing because this method is private in S3fsCurl class)

Then I think this method should be moved to S3fsMultiCurl class(private static method).
Do you think about it?

Regards,

orozery:
Yes, my mistake: the method was indeed public on the cpp file.
Fixed it now, thanks!

PR #553, opened 2017-04-02
add TLS cipher suites customization

orozery:
#### Details
Expose a command-line argument to select a custom TLS (HTTPS) cipher-suite.
Some cipher suites are faster than the other. This will enable the client to select the best cipher suite that suits its needs.


src/s3fs_util.cpp
```diff
@@ -1151,6 +1151,10 @@ void show_help (void)
     "   curldbg - put curl debug message\n"
     "        Put the debug message from libcurl when this option is specified.\n"
     "\n"
+    "   cipher_suites - customize TLS cipher suite list\n"
+    "        Customize the list of TLS cipher suites.\n"
+    "        Expects a colon separated list of cipher suite names.\n"
```
andrewgaul: Can this emit whatever ciphers the underlying TLS library supports?

orozery: If you're using the OpenSSL TLS engine, then it is possible to discard ciphers using the "!" operator.
I did not include a detailed description of the expected format as it depends on the TLS engine (openssl, gnutls, etc.)

andrewgaul: Can we ask the engine what it supports and emit those, e.g., `rsa_rc4_128_md5`, `rsa_aes_128_sha`?

andrewgaul: We need some kind of documentation so that users know how to interact with this flag.  If curl or the cipher suite can emit all valid ciphers, let's do that.  Otherwise link to the relevant documentation like curl does:

```
       --ciphers <list of ciphers>
              (SSL) Specifies which ciphers to use in the connection. The list of ciphers must specify valid ciphers.
              Read up on SSL cipher list details on this URL: http://www.openssl.org/docs/apps/ciphers.html

              NSS ciphers are done differently than OpenSSL and GnuTLS. The full list of NSS ciphers is in the NSSCi‐
              pherSuite entry at this URL: http://git.fedora‐
              hosted.org/cgit/mod_nss.git/plain/docs/mod_nss.html#Directives

              If this option is used several times, the last one will be used.
```

orozery: I've added instructions on the doc (man and command-line usage printout) with a link to the list of cipher suites.

ggtakec:
Please wait.

PR #554, opened 2017-04-02
cleanup cache directory when running out of disk space

orozery:
#### Details
When running out of disk space when using a cache, try to perform a global cache cleanup.
E.g. instead of cleaning up only the cache for the CURRENT file descriptor, go over all cache files an try to clean them up.
This will proof useful when the disk is mostly occupied by big unused cache files.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -31,6 +31,7 @@
 #include <errno.h>
 #include <string.h>
 #include <assert.h>
+#include <dirent.h>
 #include <curl/curl.h>
 #include <string>
 #include <iostream>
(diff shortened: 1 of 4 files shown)


src/fdcache.cpp
```diff
@@ -2014,7 +2048,7 @@ FdEntity* FdManager::Open(const char* path, headers_t* pmeta, ssize_t size, time
   }
 
   // open
-  if(-1 == ent->Open(pmeta, size, time)){
+  if(0 != ent->Open(pmeta, size, time, no_fd_lock_wait)){
```
ggtakec: Thanks for fixing a bug!(about checking retun value)

ggtakec:
This PR is not a problem.
Please wait.

PR #556, opened 2017-04-02
fix multipart upload handling without cache

orozery:
#### Details
The multipart upload function which is called when running out of disk space is broken.
The etag for the parts are not stored as necessary to complete the upload.
I refactored the multipart upload success callback to be called on this case as well.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -1148,18 +1148,8 @@ bool S3fsCurl::UploadMultipartPostCallback(S3fsCurl* s3fscurl)
   if(!s3fscurl){
     return false;
   }
-  headers_t::iterator it = s3fscurl->responseHeaders.find("ETag");
-  if (it == s3fscurl->responseHeaders.end()) {
-    return false;
-  }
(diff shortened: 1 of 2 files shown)


andrewgaul:
@orozery I believe this follows on to #509 -- how can we exercise this code path in our tests?

orozery:
I was able to test this using "use_cache=" which points to a small in-memory filesystem (e.g. tmpfs).
Then I created 10MB objects until the cache get full, and this exercised the no cache code paths.
Do you have a suggestion on making this part of the test suite?

ggtakec:
I fixed conflict about #560 change.(no changed this PR codes)

ggtakec:
@orozery @andrewgaul 
For example, ensure_diskfree and use_cache is specified.
ensure_diskfree value is not over cache size for the existing test part.
The cache size will also increase during the existing test, and finally upload files(additinal test for this) above the cache.
How about this?

ggtakec:
I will think about the test method later.
First of all, we merge this code.

Issue #125, opened 2015-02-19
Can't upload file larger than 20MB with 1.78 (1.77 works fine)


hplato:
To reproduce

root@x:/tmp# dd if=/dev/zero of=20500.dat bs=1024 count=20500
20500+0 records in
20500+0 records out
20992000 bytes (21 MB) copied, 0.213735 s, 98.2 MB/s
root@x:/tmp# ls -l
total 20540
-rw-r--r-- 1 root     root     20992000 Feb 18 20:02 20500.dat
root@x:/tmp# cp 20500.dat /mnt/s3/
cp: failed to close ‘/mnt/s3/20500.dat’: Operation not permitted
root@x:/tmp# dd if=/dev/zero of=20000.dat bs=1024 count=20000
20000+0 records in
20000+0 records out
20480000 bytes (20 MB) copied, 0.215942 s, 94.8 MB/s
root@x:/tmp# ls -l
total 40540
-rw-r--r-- 1 root     root     20480000 Feb 18 20:03 20000.dat
-rw-r--r-- 1 root     root     20992000 Feb 18 20:02 20500.dat
root@x:/tmp# cp 20000.dat /mnt/s3

No problem. Now remount with version 1.77 and:

root@x:/tmp# cp 20500.dat /mnt/s3/
root@x:/tmp# ls -l /mnt/s3/
total 40502
-rw-r--r-- 1 root root 20480000 Feb 18 20:03 20000.dat
-rw-r--r-- 1 root root 20992000 Feb 18 20:10 20500.dat

Logs from 1.78

s3fs_getattr(720): [path=/20500.dat]
    HeadRequest(2112): [tpath=/20500.dat]
    RequestPerform(1588): HTTP response code 404
    RequestPerform(1612): HTTP response code 404 was returned, returning ENOENT
    HeadRequest(2112): [tpath=/20500.dat/]
    RequestPerform(1588): HTTP response code 404
    RequestPerform(1612): HTTP response code 404 was returned, returning ENOENT
    HeadRequest(2112): [tpath=/20500.dat_$folder$]
    RequestPerform(1588): HTTP response code 404
    RequestPerform(1612): HTTP response code 404 was returned, returning ENOENT
  list_bucket(2277): [path=/20500.dat]
    ListBucketRequest(2585): [tpath=/20500.dat]
    RequestPerform(1588): HTTP response code 200
    append_objects_from_xml_ex(2373): contents_xp->nodesetval is empty.
[...]

PR #567, opened 2017-04-22
Do not send ACL unless overridden

andrewgaul:
Some providers such as StorageGRID do not support canned ACLs.  Amazon
defaults to the private canned ACL if not provided so do not send it
unless overridden.  References #125.


andrewgaul:
@macropin Please test this pull request.

ggtakec:
I have onde question about your PR.
Please check it.

src/curl.cpp
```diff
@@ -346,7 +346,7 @@ long             S3fsCurl::connect_timeout     = 300;  // default
 time_t           S3fsCurl::readwrite_timeout   = 60;   // default
 int              S3fsCurl::retries             = 3;    // default
 bool             S3fsCurl::is_public_bucket    = false;
-string           S3fsCurl::default_acl         = "private";
+string           S3fsCurl::default_acl         = "";
```
andrewgaul: I do not think s3fs should send `x-amz-acl` if the user does not set a value.  No value corresponds to private access, from https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html:

> Additionally, you can use the following access control–related headers with this operation. By default, all objects are private: only the owner has full control.

This will ensure that partial S3 implementations like StorageGRID work without additional configuration.

sqlbot: @andrewgaul `x-amz-acl: private` is *almost* equivalent to sending nothing, but removing it by default might be a breaking change for one specific configuratuon that needs to be verified, because `s3:x-amz-acl` is an IAM policy condition key.  What if a bucket has a policy that denies uploads without `x-amz-acl: private` set?  If the header is missing, does that policy suddenly deny uploads, even though the net effect of the upload is unchanged?

ggtakec: @andrewgaul @sqlbot Thanks for your help.

If there is a possibility that the behavior of existing s3fs users will be changed, I think that it is better not to change the default value.
This is why s3fs collaborators including us do not make it hard to support about this(if there is an issue).
By merging your patches without changing default value, s3fs can provide a way to not send x-amz-acl and users can control the sending of this headers as needed.
I think that it will be able to deal with StorageGRID etc even without changing the default value.
(In order not to send this header, I think that it is possible by specifying option 'default_acl=""'. If necessary, I think that we may make no_acl_header with the same meaning as it.)

I'd like to get your thoughts on it.

andrewgaul: I tested a deny rule which fails in master and succeeds with this PR as @sqlbot suggested:

```json
{
  "Version":"2012-10-17",
  "Statement":[
    {
      "Sid":"s3fs-567",
      "Effect":"Deny",
      "Principal": "*",
      "Action":["s3:PutObject"],
      "Resource":["arn:aws:s3:::s3fs-567/*"],
      "Condition":{"StringEquals":{"s3:x-amz-acl":["private"]}}
    }
  ]
}
```

I believe that this is the correct behavior and s3fs deviates from other S3 libraries like Apache jclouds which do not send a canned ACL unless specified.  s3fs should support partially-compatible S3 implementations without additional configuration.  If users want the explicit ACL they can configure it via `-o default_acl=private`.  If you both feel strongly about this I will change the value.

sqlbot: I believe your test proves that this is a breaking change for anyone with a `Deny` policy with `StringNotEquals` -- a policy requiring `x-amz-acl: private` in order to prevent public objects -- would fail. But, how common is such a configuration?  Comically, in spite of being the one to raise this objection, I find your arguments compelling that the empty acl is the more correct/desirable default behavior.

ggtakec: @andrewgaul 
I am planning to release v1.81 soon.
I definitely would like to incorporate this fix, but could you set default to output headers with values as private in the same way as before?
(Because this default change is scary of the impact.)

andrewgaul: Done.

ggtakec: @andrewgaul Thanks!

PR #572, opened 2017-05-02
s3fs.cpp:using string::back() instead as the C++11 recommend

liuyuhong66:
Signed-off-by: liuyuhong <liuyuhong@cmss.chinamobile.com>
#### Details
using string()::back() instead of length()-1 as the C++11 recommend


andrewgaul:
@liuyuhong66 s3fs compiles against C++03 so we cannot take advantage of this API which C++11 introduced:

http://en.cppreference.com/w/cpp/string/basic_string/back

liuyuhong66:
@andrewgaul , all right, got it

PR #593, opened 2017-05-13
Backward compatible for changing default transport to HTTPS

ggtakec:
#588 

#### Details
- Backward compatibility  
By #588 we changed the default protocol from HTTP to HTTPS, but it is affected by some users who do not specify the url option.  
To maintain backward compatibility, when s3fs fails the bucket check on HTTPS, s3fs falls back to HTTP and rechecks the bucket.
- Updated the explanation of url option of help, man page.


andrewgaul:
@ggtakec Including an HTTPS to HTTP fallback allows an attacker to force a downgrade of the transport, defeating the purpose of SSL.  I strongly recommend reverting this change and releasing 1.81.1.  You can read about a more involved but less severe downgrade attack here:

https://en.wikipedia.org/wiki/POODLE

ggtakec:
@andrewgaul Thanks for your advice.
I will fix this patch to revert soon(#596), and upgrade version number(maybe 1.82 because not want to change the version format).
I think that 1.81 will be a short-lived version that we can not use.

PR #605, opened 2017-05-22
Group permission checks sometimes fail with large number of groups

tlevi:
On some of my systems users cannot access directories, which they have group membership to access.
One if the groups in particular is quite large, with about 50 members, but I've also got ~150 groups and users in total.

I've traced the issue back to the getgrgid_r call within is_uid_include_group failing with ERANGE, indicating the buffer is too small.
The patch in this pull request solves the problem by expanding the buffer and trying again until it works, but there might be a better way.

### DIFF
--- a/src/s3fs_util.cpp
+++ b/src/s3fs_util.cpp
@@ -510,8 +510,17 @@ int is_uid_include_group(uid_t uid, gid_t gid)
     return -ENOMEM;
   }
   // get group information
-  if(0 != (result = getgrgid_r(gid, &ginfo, pbuf, maxlen, &pginfo))){
-    S3FS_PRN_ERR("could not get group information.");
+  while(ERANGE == (result = getgrgid_r(gid, &ginfo, pbuf, maxlen, &pginfo))){
+    free(pbuf);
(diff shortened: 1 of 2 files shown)


src/s3fs_util.cpp
```diff
@@ -510,8 +510,16 @@ int is_uid_include_group(uid_t uid, gid_t gid)
     return -ENOMEM;
   }
   // get group information
-  if(0 != (result = getgrgid_r(gid, &ginfo, pbuf, maxlen, &pginfo))){
-    S3FS_PRN_ERR("could not get group information.");
+  while (ERANGE == (result = getgrgid_r(gid, &ginfo, pbuf, maxlen, &pginfo))) {
+    maxlen *= 2;
+    if(NULL == (pbuf = (char*)malloc(sizeof(char) * maxlen))){
```
andrewgaul: This leaks memory -- need to call `free` or `realloc`.

tlevi: oh, you're right of course, I was doing this REALLY late ^_^

PR #608, opened 2017-05-25
Fix chown_nocopy losing existing uid/gid if unspecified

tlevi:
When the s3fs_chown_nocopy function is used, performing chown with only owner or group loses the unspecified parameter.
This PR fills that data in from stbuf, the same as s3fs_chown already does.

I've also experimented with removing the nearby getpwuid/getgrgid calls, because I can't figure out what they're for, it seems like a no-op. If it needs to stay let me know.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1705,15 +1705,6 @@ static int s3fs_chown(const char* path, uid_t uid, gid_t gid)
     return result;
   }
 
-  struct passwd* pwdata= getpwuid(uid);
-  struct group* grdata = getgrgid(gid);
-  if(pwdata){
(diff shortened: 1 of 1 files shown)


ggtakec:
@tlevi Thank you for this PR.
The same processing as s3fs_chown was not found in s3fs_chown_nocopy.

Since calling getpwuid/getgrgid function has no meaning, I think that it may be deleted.
Perhaps, I think I was checking if it is a valid uid/gid.
(However, this code has no meaning because processing continues even if it is invalid uid/gid)
And uid/gid should be allowed even if it does not exist on the system, so this process(judgment) is unnecessary.
Therefore, I think that we should delete this code.

Thanks for your kindness.

PR #609, opened 2017-05-25
Group permission checks sometimes fail with large number of groups

tlevi:
(Same as #605 with different branch)
On some of my systems users cannot access directories, which they have group membership to access.
One if the groups in particular is quite large, with about 50 members, but I've also got ~150 groups and users in total.

I've traced the issue back to the getgrgid_r call within is_uid_include_group failing with ERANGE, indicating the buffer is too small.
The patch in this pull request solves the problem by expanding the buffer and trying again until it works, but there might be a better way.


ggtakec:
@tlevi Thanks for your help. I merge this.

PR #628, opened 2017-08-01
Add test for all valid characters

gaul:
References kahing/goofys#210.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -297,9 +297,12 @@ function test_special_characters {
     describe "Testing special characters ..."
 
     ls 'special' 2>&1 | grep -q 'No such file or directory'
-    ls 'special?' 2>&1 | grep -q 'No such file or directory'
-    ls 'special*' 2>&1 | grep -q 'No such file or directory'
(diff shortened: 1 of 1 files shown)


andrewgaul:
@kahing Please review.

kahing:
this test is problematic, although you can use any ascii characters (except '/' and '\0') in POSIX, aws s3 doesn't allow anything except valid utf8 sequences. I don't know if s3fs passes this but it's definitely problematic for goofys

andrewgaul:
I reduced the character set to ASCII 11-127.  However, I believe that ASCII 1-10 can be used by setting the `encoding-type` parameter to `url`.

ggtakec:
I looked at this case.(although it is not detailed)
For example, the request for "1" is sent as "HEAD /special%01 HTTP/1.1" by url encode and s3fs does not get signature error from s3.

andrewgaul:
goofys actually issues a list here which triggers the XML errors while s3fs does not.  I would need to expand the test to create a file with ASCII 1-10 then list a directory containing that object to see similar symptoms.  Let's leave this pull request as-is.

Issue #634, opened 2017-09-01
Large file copies failing on macOS 10.12


swt2c:
I'm using s3fs on macOS 10.12.6 and I'm seeing large file copies fail most of the time.

For example:
```
$ time dd if=file_1GB_new.bin of=~/s3fuse/ScottTest3/1GB_dd bs=1048576
1000+0 records in
1000+0 records out
1048576000 bytes transferred in 189.940385 secs (5520553 bytes/sec)

real	3m13.023s
user	0m0.007s
sys	0m4.264s
$ ls -l ~/s3fuse/ScottTest3/1GB_dd
-rw-r--r--  1 stalbert  staff  0 Sep  1 11:05 /Users/stalbert/s3fuse/ScottTest3/1GB_dd
```
Some of the time dd or cp will report an I/O error and some of the time it appears successful.  However, the result is always a 0-byte file on the S3 side.  I am using the latest s3fs from git (fa8c417), compiled with `--with-gnutls`, and OSXFUSE v3.6.3.  Smaller files (e.g., 2MB) are usually copied successfully.

An additional data point, I am using a bucket in US-West, but I'm on the US East coast.

#### Additional Information

- Version of s3fs being used (s3fs --version)
 - Amazon Simple Storage Service File System V1.82(commit:fa8c417) with GnuTLS(gcrypt)

- Version of fuse being used (pkg-config --modversion fuse)
 - 2.9.7

- System information (uname -a)
 - Darwin Scotts-Mac-Pro.local 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64 x86_64

- Distro (cat /etc/issue)
 - macOS 10.12.6

- s3fs command line used (if applicable)
```
s3fs grack12test ~/s3fuse/
```
- /etc/fstab entry (if applicable): 
```
```
- s3fs syslog messages (grep s3fs /var/log/syslog, or s3fs outputs)  
_if you execute s3fs with dbglevel, curldbg option, you can get detail debug messages_
```
```

ggtakec:
@swt2c Thanks for posting a issue and sending codes.
#636 is merged, and I'm closing this issue.
If you find another issue, please post new issue.
Thanks,

PR #636, opened 2017-09-07
Fix intermittent upload failures on macOS

swt2c:
There were multiple problems with the FdManager::GetFreeDiskSpace() function
on macOS:
1) When calling statvfs(), f_frsize should be used instead of f_bsize when converting available blocks to bytes.  This was causing the free space calculation to be incorrect.
2) On macOS, fsblkcnt_t is a 32-bit integer.  Thus, when calculating available disk space, there were frequently overflows.  This caused s3fs to incorrectly determine that the cache location was out of space in the middle of a transfer which caused uploads to fail.  Changing this to a uint64_t resolves the problem.

Resolves issue #634.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -1908,7 +1908,7 @@ size_t FdManager::SetEnsureFreeDiskSpace(size_t size)
   return old;
 }
 
-fsblkcnt_t FdManager::GetFreeDiskSpace(const char* path)
+uint64_t FdManager::GetFreeDiskSpace(const char* path)
 {
   struct statvfs vfsbuf;
   string         ctoppath;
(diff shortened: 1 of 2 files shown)


src/fdcache.cpp
```diff
@@ -1930,12 +1930,12 @@ fsblkcnt_t FdManager::GetFreeDiskSpace(const char* path)
     S3FS_PRN_ERR("could not get vfs stat by errno(%d)", errno);
     return 0;
   }
-  return (vfsbuf.f_bavail * vfsbuf.f_bsize);
+  return (vfsbuf.f_bavail * vfsbuf.f_frsize);
```
andrewgaul: We should multiply `f_bavail` by `f_bsize` or `f_favail` by `f_frsize` but not mix block and fragment units.  Instead do you want to cast one of these to `uint64_t`?

ggtakec:
@swt2c Thanks for fixing this bug, it is my fault.
I'm going to merge your patch soon.
Regards,

PR #641, opened 2017-09-15
Enable Requester Pays Buckets

3DTOPO:

#### Details

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -2276,6 +2276,7 @@ void S3fsCurl::insertV4Headers(const string &op, const string &path, const strin
   requestHeaders = curl_slist_sort_insert(requestHeaders, "host", get_bucket_host().c_str());
   requestHeaders = curl_slist_sort_insert(requestHeaders, "x-amz-content-sha256", contentSHA256.c_str());
   requestHeaders = curl_slist_sort_insert(requestHeaders, "x-amz-date", date8601.c_str());
+  requestHeaders = curl_slist_sort_insert(requestHeaders, "x-amz-request-payer", "requester");
 
   if(!S3fsCurl::IsPublicBucket()){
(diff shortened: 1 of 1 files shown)


src/curl.cpp
```diff
@@ -2276,6 +2276,7 @@ void S3fsCurl::insertV4Headers(const string &op, const string &path, const strin
   requestHeaders = curl_slist_sort_insert(requestHeaders, "host", get_bucket_host().c_str());
   requestHeaders = curl_slist_sort_insert(requestHeaders, "x-amz-content-sha256", contentSHA256.c_str());
   requestHeaders = curl_slist_sort_insert(requestHeaders, "x-amz-date", date8601.c_str());
+  requestHeaders = curl_slist_sort_insert(requestHeaders, "x-amz-request-payer", "requester");
```
andrewgaul: Why do you add this header unconditionally?

3DTOPO: Because it fixed the issue I was having, and as I stated, it would ideally be added as an option

andrewgaul: Sending unneeded headers to some S3 implementations will cause requests to fail, as with S3Proxy running in Travis.  Please add a flag.

ggtakec:
@3DTOPO 
You can not add the x-amz-request-payer header for every request.
Instead of it, you can prepare a dedicated startup option and add a header only when you specify that option.

I think this method is good.
And I reopened your #635 issue, thus we need to discuss about request payer.

Regards,

ggtakec:
@3DTOPO This PR has been left unmodified.
Since we have made considerable changes from the base code, I will close this PR.
Please open again if there is code update.

src/curl.cpp
ahuarte47: Hi, I have created this [PR](https://github.com/s3fs-fuse/s3fs-fuse/pull/1199) implementing this flag as a new option. I hope this helps!

Issue #640, opened 2017-09-14
unclear if certs validation is occuring?


quadgnim:
I noticed in https://github.com/s3fs-fuse/s3fs-fuse/blob/master/src/s3fs.cpp, line 4938 indicates the cert validation logic is commented out.  Can anyone verify if this is being done elsewhere, and if not, why is it commented out, is there a problem with it?  to protect from man in the middle attacks, we'd very much like to see cert validation against the s3 endpoint ocuring.  Is this in the plans to enable soon?

ggtakec:
@quadgnim 
This comment is outdated.
(We will delete comments at a later date.)

By default we are checking with CURLOPT_SSL_VERIFYHOST.
https://github.com/s3fs-fuse/s3fs-fuse/blob/master/src/curl.cpp#L1589

If you do not want to check, please specify ssl_verify_hostname = 0.

Currently, even if "." (Period) is included in the bucket name, it will work if you set use_path_request_style.
https://github.com/s3fs-fuse/s3fs-fuse/wiki/FAQ#q-https-connecting-failed-if-bucket-name-includes-dod

gaul:
As demonstrated by S3Proxy unit tests, SSL validation occurs and we specifically need to disable it for this self-signed certificate.  Please reopen if I misunderstand this issue.

PR #645, opened 2017-09-17
Configure S3Proxy for SSL

andrewgaul:
This also demonstrates that SSL certificate checking occurs and the
tests must disable it for S3Proxy's self-signed certificate.
References #640.


ggtakec:
@andrewgaul Thanks

PR #652, opened 2017-09-30
Fix s3fs_init message

jurafxp:
none

#### Details
Seems informative message to me, not critical. Also quite annoying to see it, when reviewing systemd journal.


ggtakec:
@jurafxp 
As you say, I think this message is INFO (or NOTICE), not CRIT.
We have been outputting messages at this level for a long time, so I would like to hear the opinions of the surroundings.

@gaul May I have your opinion?

ggtakec:
I agree with @jurafxp and decided that the message at startup is INFO level.
If you have a problem bout this, please post PR or Issue.
Regards,

PR #658, opened 2017-10-14
Do not fail updating directory when removing old-style object

tlevi:
Fixes #625

When modifying directories, removing the existing object should not cause a failure if the older style directory does not exist.

Issue #625, opened 2017-07-25
s3fs-fuse with GCS - chmod/chown of a dir causes chown: changing ownership of 'newdir': No such file or directory


agmistry:
Environment
--------------
s3fs --version
Amazon Simple Storage Service File System V1.82(commit:b1fe419) with OpenSSL
Copyright (C) 2010 Randy Rizun <rrizun@gmail.com>

Debian 9.0 - Google Compute Engine VM

Steps to recreate bug
------------------------
/etc/fuse.conf has user_allow_other enabled

Mounting the GCS bucket
s3fs <GCS Bucket> <Mount Point> -o url=https://storage.googleapis.com -o sigv2 -o nomultipart -d -d -f -o f2 -o curldbg -o passwd_file=/etc/passwd-s3fs  -o allow_other

As a Local User
cd <Mount Point>
mkdir mytestdir
ls -ld mytestdir
###Verify that the directory was created with the proper ownership (i.e. the User)

As root
cd <Mount Point>
mkdir myrootdir
chown <User> myrootdir
chown: changing ownership of 'myrootdir': No such file or directory

Hopefully I've provided enough information to be able to triage this problem. FYI I have the same setup to an AWS S3 bucket which works fine. I also get the same problem doing chmod or even moving a directory onto the mount point.

I need to be able to create directories as root and change permissions because I'm using Cloud Storage to store home dirs of CHRooted SFTP accounts.

Thanks
Anant 

Relevant output from curldbg
---------------------------------
unique: 50, opcode: LOOKUP (1), nodeid: 1, insize: 50, pid: 871
LOOKUP /myrootdir
getattr /myrootdir
[INF] s3fs.cpp:s3fs_getattr(812): [path=/myrootdir]
[DBG] s3fs.cpp:check_parent_object_access(669): [path=/myrootdir]
[DBG] s3fs.cpp:check_object_access(563): [path=/]
[DBG] s3fs.cpp:get_object_attribute(416): [path=/]
[DBG] s3fs.cpp:check_object_access(563): [path=/myrootdir]
[DBG] s3fs.cpp:get_object_attribute(416): [path=/myrootdir]
[INF]       curl.cpp:HeadRequest(2516): [tpath=/myrootdir]
[INF]       curl.cpp:PreHeadRequest(2453): [tpath=/myrootdir][bpath=][save=][sseckeypos=-1]
[DBG] curl.cpp:GetHandler(283): Get handler from pool: 31
[...]

PR #659, opened 2017-10-15
Do not fail updating directory when removing old-style object(ref #658)

ggtakec:
base of #658 PR for issue #625

#### Details
This is PR based on @tlevi - #658.
It is a code that also corresponded to other related processing related to the bug found @tlevi.
For details, see #658.


tlevi:
Sure, solves my problem, so looks good to me. ;-)

ggtakec:
@tlevi Thanks for your help.

PR #669, opened 2017-11-05
add IBM IAM authentication support

orozery:
#### Details
IBM COS (Cloud Object Storage) supports OAuth 2.0 tokens as a mean of authentication, in addition to traditional AWS v2/v4 authentication.

This PR allows s3fs to use this new authentication mechanism.
The user implicitly selects this authentication by supplying a password file with the  following format:

IBMServiceInstanceID=...
IBMAPIKey=...

Where the first (Service Instance ID) is only required when creating a bucket.

I chose to use this password file format, as a similar extension to the AWS password file format (which is already supported in s3fs):

AWSAccessKeyId=...
AWSSecretKey=...

I tested this PR using a real IBM COS account, and got all tests passing:

ubuntu@ubuntu:~/s3fs-fuse/test$ S3FS_CREDENTIALS_FILE=/home/ubuntu/s3fs-fuse/ibm-passwd TEST_BUCKET_1=ozeri-test-1 S3PROXY_BINARY="" S3_URL="https://s3.us-south.objectstorage.softlayer.net" ./small-integration-test.sh

...

PASS: test_append_file
PASS: test_truncate_file
PASS: test_truncate_empty_file
PASS: test_mv_file
PASS: test_mv_directory
PASS: test_redirects
PASS: test_mkdir_rmdir
PASS: test_chmod
PASS: test_chown
PASS: test_list
PASS: test_remove_nonempty_directory
PASS: test_multipart_upload
PASS: test_multipart_copy
PASS: test_special_characters
PASS: test_symlink
PASS: test_extended_attributes
PASS: test_mtime_file
PASS: test_rm_rf_dir
PASS: test_write_after_seek_ahead
SUMMARY for ./integration-test-main.sh: 19 tests passed.  0 tests failed.
./small-integration-test.sh: tests complete.


ggtakec:
@orozery 
It seems there is no problem with the code.
Actually I have not been able to test with the IBM COS account, so I can not confirm the execution result.

Because it works as IBM in the password file format, it does not seem to affect the existing one.

If there is no other problem, I would like to merge.
Please let me know if there is any opinion.

orozery:
@gaul I would be happy to get your feedback as well.

orozery:
@ggtakec 
In my opinion we can merge this.

If you want to test it, you can create a free account here:
https://console.bluemix.net/registration/free/?target=%2Fcatalog%2Finfrastructure%2Fcloud-object-storage

As I said, I already tested it with my account.

ggtakec:
@orozery Thanks for your kindness.
But I'm very sorry to keep you waiting, I will review by the end of the week.

src/curl.cpp
```diff
@@ -1488,6 +1507,69 @@ bool S3fsCurl::SetIAMRoleFromMetaData(const char* response)
...
+
+    S3FS_PRN_DBG("(IBM token response line) %s: %s", key.c_str(), val.c_str());
+
+    if(key == "access_token"){
+      creds.access_token = val;
+    }else if(key == "expiration"){
+      creds.expiration = strtol(val.c_str(), NULL, 0);
+    }
+  }
+  return true;
```
gaul: It is possible that you parse an empty response yet return `true`.

src/curl.cpp
```diff
@@ -1488,6 +1507,69 @@ bool S3fsCurl::SetIAMRoleFromMetaData(const char* response)
...
+  S3FS_PRN_INFO3("IBM credential response = \"%s\"", response);
+
+  ibm_credentials creds;
+  creds.access_token = "";
+  creds.expiration = 0;
+
+  if(!ParseIBMCredentialResponse(response, creds)){
+    return false;
+  }
+  if(creds.access_token == "" || creds.expiration == 0){
```
gaul: Prefer `creds.access_token.empty`.

src/curl.cpp
```diff
@@ -1488,6 +1507,69 @@ bool S3fsCurl::SetIAMRoleFromMetaData(const char* response)
...
+
+  while(getline(sscred, oneline, ',')){
+    string::size_type pos;
+    string            key;
+    string            val;
+    if(string::npos == (pos = oneline.find(':'))){
+      continue;
+    }
+    key = trim(trim(oneline.substr(0, pos), "{\""));
+    val = trim(trim(oneline.substr(pos+1)), "}\"");
```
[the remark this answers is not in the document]
gaul: #671 proposes adding libjsoncpp to parse IAM roles.  Maybe we should use it here too?

src/curl.cpp
```diff
@@ -327,6 +327,9 @@ void CurlHandlerPool::ReturnHandler(CURL* h)
 #define IAMCRED_EXPIRATION          "Expiration"
 #define IAMCRED_KEYCOUNT            4
 
+#define IBM_TOKEN_EXPIRE_MERGIN     (5 * 60)         // update timing
+#define IBM_CRED_URL                "https://iam.bluemix.net/oidc/token"
```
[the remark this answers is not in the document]
gaul: Opened #675.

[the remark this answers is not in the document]
ggtakec: @orozery Please check latest codes, and fix your codes like it.

src/curl.cpp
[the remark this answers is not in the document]
ggtakec: @orozery I think you can change codes like forrowing codes:
https://github.com/s3fs-fuse/s3fs-fuse/blob/master/src/curl.cpp#L1424-L1447

src/s3fs.cpp
```diff
@@ -3888,6 +3888,96 @@ static int check_for_aws_format(void)
   }
 }
 
+// Return:  1 - OK(could read and set apikey etc.)
+//          0 - NG(could not read)
+//         -1 - Should shutdown immediately
+static int check_for_ibm_format(void)
```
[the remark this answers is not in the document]
ggtakec: @orozery I merged new codes about reading passwd file.
I think that helps to merge your codes.
Please see following codes:
https://github.com/s3fs-fuse/s3fs-fuse/blob/master/src/s3fs.cpp#L3940-L3960
https://github.com/s3fs-fuse/s3fs-fuse/blob/master/src/s3fs.cpp#L4055-L4062

ggtakec:
@orozery We merged some codes before merging your codes, and we want to change your codes about our comment on review.
Please change your PR's codes, and let us know.

orozery:
@gaul @ggtakec Thank you for your comments!

I think I've addressed all of them in my latest commit.
I've re-used as much code as I can from the AWS token usage, as well as using the existing passwd file format (instead of adding a new custom format).
The new approach simply requires passing a new command line argument (-o ibm_auth).

Please let me know your thoughts.

src/curl.cpp
```diff
@@ -1456,21 +1493,26 @@ bool S3fsCurl::SetIAMCredentials(const char* response)
...
   }
 
-  S3fsCurl::AWSAccessKeyId       = keyval[string(IAMCRED_ACCESSKEYID)];
-  S3fsCurl::AWSSecretAccessKey   = keyval[string(IAMCRED_SECRETACCESSKEY)];
-  S3fsCurl::AWSAccessToken       = keyval[string(IAMCRED_ACCESSTOKEN)];
-  S3fsCurl::AWSAccessTokenExpire = cvtIAMExpireStringToTime(keyval[string(IAMCRED_EXPIRATION)].c_str());
+  S3fsCurl::AWSAccessToken       = keyval[string(S3fsCurl::IAM_token_field)];
+
+  if(S3fsCurl::is_ibm_auth){
+    S3fsCurl::AWSAccessTokenExpire = strtol(keyval[string(S3fsCurl::IAM_expiry_field)].c_str(), NULL, 0);
```
gaul: Should the third argument be 10 instead of 0?  Also why call `string()` when `IAM_expiry_field` is already a string?  Repeated below.

src/curl.cpp
```diff
@@ -2369,9 +2410,26 @@ void S3fsCurl::insertV2Headers()
   }
 }
 
+void S3fsCurl::insertIBMHeaders()
+{
+  requestHeaders = curl_slist_sort_insert(requestHeaders, "Authorization", ("Bearer " + S3fsCurl::AWSAccessToken).c_str());
+
+  if(op == "PUT" && path == mount_prefix + "/"){
+    // create bucket request
```
gaul: Can you expand this comment a bit since I don't understand why you add this extra header.

src/s3fs.cpp
```diff
@@ -4540,14 +4545,33 @@ static int my_fuse_opt_proc(void* data, const char* arg, int key, struct fuse_ar
       passwd_file = strchr(arg, '=') + sizeof(char);
       return 0;
     }
+    if(0 == strcmp(arg, "ibm_auth")){
+      S3fsCurl::SetIsIBMAuth(true);
+      S3fsCurl::SetIAMCredentialsURL("https://iam.bluemix.net/oidc/token");
+      S3fsCurl::SetIAMTokenField("access_token");
+      S3fsCurl::SetIAMExpiryField("expiration");
+      S3fsCurl::SetIAMFieldCount(2);
+      // IBM auth only supports a-xmz-acl: public-read
```
gaul: Typo: `x-amz-acl`.

src/s3fs.cpp
```diff
@@ -4540,14 +4545,33 @@ static int my_fuse_opt_proc(void* data, const char* arg, int key, struct fuse_ar
...
+      if(S3fsCurl::SetDefaultAcl("") == "public-read"){
```
gaul: We need to add `GetDefaultAcl` since this call will set `default_acl` to "" even if it was previously non-empty.  Also I think you want to check this outside of `my_fuse_opt_proc` in case users specify IBM auth before ACL.

src/s3fs.cpp
```diff
@@ -4540,14 +4545,33 @@ static int my_fuse_opt_proc(void* data, const char* arg, int key, struct fuse_ar
...
+        S3fsCurl::SetDefaultAcl("public-read");
```
gaul: I don't understand this; does COS support `private` as well as `public-read`?

doc/man/s3fs.1
```diff
@@ -219,6 +219,9 @@ Because s3fs always checks MD5 when uploading large object, this option does not
 \fB\-o\fR iam_role ( default is no IAM role )
 This option requires the IAM role name or "auto". If you specify "auto", s3fs will automatically use the IAM role names that are set to an instance. If you specify this option without any argument, it is the same as that you have specified the "auto".
 .TP
+\fB\-o\fR ibm_auth ( default is not using IBM authentication )
+This option instructs s3fs to use IBM IAM authentication. In this mode, the AWSAccessKey and AWSSecretKey will be used as IBM's Service-Instance-ID and APIKey, respectively.
```
gaul: Should this describe the service, e.g., IBM Bluemix COS, IBM Cloud, or whatever the preferred name is?  Repeated below.

gaul:
Also would be good to call out IBM authentication in the README.

doc/man/s3fs.1
orozery: I think "IBM IAM authentication" is the correct description.
I will fix this instance (did not see any other).

src/curl.cpp
orozery: Will change to:
// ibm-service-instance-id header is required for bucket creation requests

src/s3fs.cpp
orozery: My bad, it supports both private (implicitly as the default ACL) and public.
Will fix.

doc/man/s3fs.1
gaul: I agree that this is correct, but I'm more concerned that a user will see the flag and not associate it with IBM's cloud offering.  Adding some aside like (used with IBM Cloud) might help.

orozery:
@gaul Thanks again for taking the time to review this PR.
I've added another commit according to your suggestions.

orozery:
@gaul I'm now addressing it as "IBM IAM Authentication", instead of "IBM Authentication".
IBM Cloud Object Storage supports both IAM authentication and legacy AWS-style signed authentication.

@ggtakec Can you please make a final review?

ggtakec:
@orozery It seems good for me, but I(we) have new issue #685 which relates to your code(about jsoncpp).
I will work #685 soon, please wait a moment.

And I am interested in IBM authentication which @gaul says. (ex. That is the percentage of its aws / IBM users.)
Of course, incorporating the IBM type is not a problem.

ggtakec:
@orozery I'm sorry for another PR merging before this, then this PR has conflicting now.
Please fix it, after that I try to merge this.

orozery:
@ggtakec @gaul I've rebased the code, and made 2 necessary changes to `S3fsCurl::ParseIAMCredentialResponse` (since we don't use jsoncpp anymore):

1. Changed the parser to split on `,` instead of a newline (`\n`).
2. Added parsing of integer value (the previous version supported only string values).

ggtakec:
@orozery 
I confirmed the two additional changes, I think that there is no problem.
Thanks for that you have done a lot for merging this feature.
I will merge this PR soon.

orozery:
@ggtakec Thank you!
Do you have plans for a new release (1.83)?

ggtakec:
@orozery Yes, I wanted to release it sooner but I would like to release it this month.

PR #672, opened 2017-11-14
README.md Addition

lmlsna:
#### Details
I added an example fstab line for mounting a non-Amazon implementation.

Probably pretty obvious to most that you have to provide the same command line options in fstab, but maybe not to everyone, so included for completeness.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,14 @@ or
 mybucket /path/to/mountpoint fuse.s3fs _netdev,allow_other 0 0
 ```
 
+When using fstab to mount a non-Amazon S3 implementation, the URL and
+path-style options also must be provided.
+
+```
+s3fs#mybucket /path/to/mountpoint fuse _netdev,allow_other,use_path_request_style,url=http://url.to.s3/ 0 0
+
+```
+
 Note: You may also want to create the global credential file first
 
 ```


strk:
Looks good to me

README.md
```diff
@@ -96,6 +96,14 @@ or
 mybucket /path/to/mountpoint fuse.s3fs _netdev,allow_other 0 0
 ```
 
+When using fstab to mount a non-Amazon S3 implementation, the URL and
+path-style options also must be provided.
+
+```
+s3fs#mybucket /path/to/mountpoint fuse _netdev,allow_other,use_path_request_style,url=http://url.to.s3/ 0 0
+
```
gaul: Extra space.

lmlsna: Yeah definitely not needed, I just thought another example wouldn't hurt, especially for people who do not already understand how mount options are specified in fstab. 

I rarely find myself frustrated by excessive documentation, it's almost always the other way around. But it's your baby, if you don't want, that's cool.

ggtakec:
@lmlsna Thanks for your PR.
But I do not merge this PR, because duplicate content will be present in README.md, as @gaul says.
Instead, I will do a new PR for your README based on your PR.
It changes the order of the contents of the README slightly and makes the example of both command line and fstab even in the case of No S3.
I'll PR soon, and merge it.
Regards,

ggtakec:
I merged README.md which is including this PR content.

PR #683, opened 2017-11-19
Update instructions for CentOS and Ubuntu

gaul:
CentOS requires EPEL for jsoncpp.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,13 +27,14 @@ Installation
 On Ubuntu 14.04:
 
 ```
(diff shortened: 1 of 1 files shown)


gaul:
Updating https://github.com/s3fs-fuse/s3fs-fuse/wiki/Installation-Notes#fedora--centos is a little more complicated since there is some implication of CentOS 5 support.  I think we should limit our instructions to supported releases: CentOS 6, Ubuntu 14.04, latest Fedora and non-LTS Ubuntu.  The current mishmash of older instructions is confusing and incorrect in some areas.

ggtakec:
@gaul Thanks for your kindness.
But I changed s3fs codes which does not use jsoncpp(#685), then this PR will be closed without merging.
Sorry for my decision was careless.

Issue #685, opened 2017-11-22
Consider using jsoncpp


ggtakec:
#### Additional Information
- Version of s3fs being used (s3fs --version)  - current master branch
- Version of fuse being used (pkg-config --modversion fuse) - n/a
- System information (uname -a)  - n/a
- Distro (cat /etc/issue)  - n/a

#### Details about issue
Issue(PR) #669 #678 #682 is related

Adding jsoncpp is the best way to analyze JSON.
However, there are differences in jsoncpp's package depending on each distribution.
Analysis of JSON code for s3fs is not main, it is only a part of processing, it is limited.
Therefore, I am planning to consider a method that does not use jsoncpp (as before).

And the current code needs to be modified, but It needs C++11(Dependency is troublesome).
- current codes
```
  Json::Value root;
  Json::Reader reader;
  if (!reader.parse(response, root)) {
    return false;
  }
```
- should be(example)
```
  Json::Value root;
  Json::CharReaderBuilder rbuilder;
  Json::CharReader* reader(rbuilder.newCharReader());
  string error;
  if(!reader->parse(response, &response[strlen(response)], &root, &error)){
    delete reader;
    return false;
  }
  ...
  delete reader;
```

I'm checking if we can make the code simply without jsoncpp.
And please give us your opinions.

orozery:
@ggtakec Are you suggesting we write our own Json parser, instead of using jsoncpp?

ggtakec:
@orozery It is as a possibility.
If s3fs uses jsoncpp, consistency with OSX and other Linux distributions may not be unique, so people who build s3fs may have trouble.
Of course, I think that it is good to use jsoncpp, but I would like to consider including support.
About this, I'd be glad to give me any opinion from a broad perspective.

gaul:
@ggtakec Trying to understand the symptoms -- do you mean that older systems with jsoncpp 0.x work with C++03 and newer systems with jsoncpp 1.x require C++11?  Can we just specify the older version which seems to be maintained, commits as of April 2017?  Agreed that reverting to the simpler parser may cause fewer headaches.

ggtakec:
@gaul I will organize the problem.
In the current master, the following warnings are output for osx(xcode 8.3) building.
This warning means to use the 1.y.z version of jsoncpp.
```
curl.cpp:1431:9: warning: 'Reader' is deprecated: Use CharReader and CharReaderBuilder instead [-Wdeprecated-declarations]
```

[...]

PR #686, opened 2017-11-22
remove use of jsoncpp

orozery:
#685 

#### Details
Reverting the use of jsoncpp - introduced recently in #671.


ggtakec:
From #685 , we need this patch.

PR #695, opened 2017-12-04
fix condition for parallel download

orozery:
### Details
The parallel download request condition needs a small fix.
The parallel upload request condition is correct
(see: https://github.com/s3fs-fuse/s3fs-fuse/blob/566961c7a5db78066bdceefd385acbb964feb1c4/src/fdcache.cpp#L1505)


ggtakec:

PR #699, opened 2017-12-14
Fix dbglevel usage

casidiablo:


ggtakec:
@casidiablo and @gaul Thanks for fixing.

PR #710, opened 2018-01-29
add disk space reservation

orozery:
### Details
This PR includes the following changes:

- Atomically reserving disk space instead of calling IsSafeDiskSpace before Load, and freeing reserved disk-space when Load completes. This is to avoid a race where two or more threads think they have enough disk space to download, but they don't account for each other. If not enough space exists on disk to reserve, then two methods are tried to free up disk space: freeing disk space from the current fd, followed by trying other fd's (when using -o use_cache).
- Changed EnsureDiskFree setting to be 0 for default (since we have the disk space reservation, we no longer need to default it to MaxParallelCount * MultipartSize).
- Added some debug printings to cache cleanup methods.
- Changed cache cleanup method to skip opened fd's (i.e. to clean from the cache only files which are not currently opened).

@ggtakec @gaul would appreciate your review.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -1429,7 +1429,7 @@ int FdEntity::NoCacheCompleteMultipartPost(void)
 
 int FdEntity::RowFlush(const char* tpath, bool force_sync)
 {
-  int result;
+  int result = 0;
 
   S3FS_PRN_INFO3("[tpath=%s][path=%s][fd=%d]", SAFESTRPTR(tpath), path.c_str(), fd);
 
@@ -1448,10 +1448,12 @@ int FdEntity::RowFlush(const char* tpath, bool force_sync)
   if(0 < restsize){
     if(0 == upload_id.length()){
       // check disk space
-      if(FdManager::IsSafeDiskSpace(NULL, restsize)){
+      if(ReserveDiskSpace(restsize)){
         // enough disk space
         // Load all uninitialized area
-        if(0 != (result = Load())){
+        result = Load();
+        FdManager::get()->FreeReservedDiskSpace(restsize);
+        if(0 != result){
           S3FS_PRN_ERR("failed to upload all area(errno=%d)", result);
           return static_cast<ssize_t>(result);
         }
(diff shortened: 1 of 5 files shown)


src/fdcache.cpp
```diff
@@ -2181,17 +2195,22 @@ bool FdManager::ChangeEntityToTempPath(FdEntity* ent, const char* path)
...
+    S3FS_PRN_INFO("cache cleanup ended");
+  }else{
+    // wait for other thread to finish cache cleanup
+    AutoLock auto_lock(&FdManager::cache_cleanup_lock);
   }
-
-  CleanupCacheDirInternal("");
 }
 
 void FdManager::CleanupCacheDirInternal(const std::string &path)
```
effi-ofer: Hey @orozery when cleaning up the cache, where do we update the free_disk_space?  I looked through the code and its not clear to me.  I see that we initialize the cache with size 0, but I can't find the place where we reduce it during a cleanup while keeping the amount of reserved space accounted for.

orozery: free_disk_space is actually reserved_disk_space.
To check if you have enough space, you check GetDiskFreeSpace() - reserved_disk_space > requested_amount

orozery:
@ggtakec Is there a problem in Travis?

I'm getting:

$ make cppcheck
cppcheck --quiet --error-exitcode=1 \
            --inline-suppr \
            --std=c++03 \
            -U CURLE_PEER_FAILED_VERIFICATION \
            -U P_tmpdir \
            -U ENOATTR \
            --enable=all \
            --suppress=missingIncludeSystem \
            --suppress=unusedFunction \
            --suppress=variableScope \
            src/ test/
[src/curl.cpp:4075]: (error) Memory leak: new_item
[src/gnutls_auth.cpp:154]: (error) Memory leak: digest
[src/gnutls_auth.cpp:159]: (error) Memory leak: digest
[src/gnutls_auth.cpp:161]: (error) Memory leak: digest
[src/gnutls_auth.cpp:174]: (error) Memory leak: digest
[src/gnutls_auth.cpp:179]: (error) Memory leak: digest
[src/gnutls_auth.cpp:181]: (error) Memory leak: digest
[src/gnutls_auth.cpp:111]: (error) Memory leak: digest
[src/gnutls_auth.cpp:120]: (error) Memory leak: digest
[src/gnutls_auth.cpp:130]: (error) Memory leak: digest
[src/gnutls_auth.cpp:139]: (error) Memory leak: digest
[src/nss_auth.cpp:128]: (error) Memory leak: digest
[src/nss_auth.cpp:132]: (error) Memory leak: digest
[src/string_util.cpp:307]: (error) Memory leak: result
[src/string_util.cpp:357]: (error) Memory leak: result
make: *** [cppcheck] Error 1

These errors have nothing to do with this PR.

ggtakec:
@orozery I'm sorry for late, and I fixed TravisCI problem.
But I'm sorry about fdcache.cpp file is conflicted, could you resolve those?
Thanks.

orozery:
@ggtakec I've rebased the PR.

ggtakec:
@orozery Thanks for PR, I merged this PR.

815331793:
Hi, @orozery ,I am some little confused about when a thread lock cache_cleanup_lock fail,  why he need wait for other thread to finish cache cleanup. In my understand, when did't have enough disk, the thread can going on with no cache way, Can you help me understand it, Thank you~ 
`void FdManager::CleanupCacheDir()
{
  S3FS_PRN_INFO("cache cleanup requested");

  if(!FdManager::IsCacheDir()){
    return;
  }

  AutoLock auto_lock_no_wait(&FdManager::cache_cleanup_lock, true);

  if(auto_lock_no_wait.isLockAcquired()){
    S3FS_PRN_INFO("cache cleanup started");
    CleanupCacheDirInternal("");
    S3FS_PRN_INFO("cache cleanup ended");
  }else{
    // wait for other thread to finish cache cleanup
    AutoLock auto_lock(&FdManager::cache_cleanup_lock);
  }
}`

orozery:
@815331793 when we call `CleanupCacheDir`, we want to free-up space.
There's no point in multiple threads doing the cleanup, so only one is cleaning up (using the lock).
The other threads wait for the cleanup to be completed.
This is because you want to return from `CleanupCacheDir` only AFTER the cleanup is completed (so that you can use the freed space).

Issue #713, opened 2018-02-03
Travis CI make cppcheck has errors on macOS


gaul:
Maybe a newer version there:

```
$ make cppcheck
cppcheck --quiet --error-exitcode=1 \
            --inline-suppr \
            --std=c++03 \
            -U CURLE_PEER_FAILED_VERIFICATION \
            -U P_tmpdir \
            -U ENOATTR \
            --enable=all \
            --suppress=missingIncludeSystem \
            --suppress=unusedFunction \
            --suppress=variableScope \
            src/ test/
[src/curl.cpp:4075]: (error) Memory leak: new_item
[src/gnutls_auth.cpp:154]: (error) Memory leak: digest
[src/gnutls_auth.cpp:159]: (error) Memory leak: digest
[src/gnutls_auth.cpp:161]: (error) Memory leak: digest
[src/gnutls_auth.cpp:174]: (error) Memory leak: digest
[src/gnutls_auth.cpp:179]: (error) Memory leak: digest
[src/gnutls_auth.cpp:181]: (error) Memory leak: digest
[src/gnutls_auth.cpp:111]: (error) Memory leak: digest
[src/gnutls_auth.cpp:120]: (error) Memory leak: digest
[src/gnutls_auth.cpp:130]: (error) Memory leak: digest
[src/gnutls_auth.cpp:139]: (error) Memory leak: digest
[src/nss_auth.cpp:128]: (error) Memory leak: digest
[src/nss_auth.cpp:132]: (error) Memory leak: digest
[src/string_util.cpp:307]: (error) Memory leak: result
[src/string_util.cpp:357]: (error) Memory leak: result
make: *** [cppcheck] Error 1
```

https://travis-ci.org/s3fs-fuse/s3fs-fuse/jobs/336597559#L1971

ggtakec:
@gaul I'm sorry for my late reply.
I'm correcting error report by cppcheck on osx.
However, the problem is probably not solved with the latest version of cppcheck(brew, 1.82).
I'm still investigating, but phenomena similar to the following are out.
https://github.com/danmar/cppcheck/pull/1078
Please wait a moment.

ggtakec:
Fixed this issue by #727.
I'm closing this, if you found a problem, please reopen this.
Thanks

PR #712, opened 2018-02-02
Added Cygwin build options

chrilith:
### Relevant Issue (if applicable)
Doesn't build on Cygwin

### Details
`libfuse` is not available on Windows.
When you try to compile `libfuse` on Windows the result is:
```
Error encountered: libfuse does not support Windows.
Take a look at http://www.secfs.net/winfsp/ instead
```

WinFSP provides an compatibility layer to FUSE using version number `2.8`, this is why I added a specific target configuration for Cygwin to `configure.ac`. Also, to use the non-standard calls used in the `s3fs` code, the `_GNU_SOURCE` flag is required.

Last change is the addition of a `.gitattributes` file. Indeed, `autogen.sh` must be in Unix format to be executed and when using Windows the file is changed to Windows line endings. hence the addition.

Next step is to have a working example for it. I wasn't able to mount a S3 bucket.
I started a discussion here: https://groups.google.com/forum/#!topic/winfsp/xXe6M_MoW8w

I someone care to help :)
Thanks!

PR #704, opened 2017-12-22
Update README.md with details about .passwd-s3fs

vadimeremeev:
### Relevant Issue (if applicable)
_If there are Issues related to this PullRequest, please list it._

### Details
_Please describe the details of PullRequest._

PR #727, opened 2018-02-28
Fixed Travis CI error about cppcheck - #713

ggtakec:
#713

### Details
The version of cppcheck prepared by TravisCI differs between OSX and ubuntu.
OSX was 1.82, and the behavior of unusedFunction was different from 1.63.
On OSX, exitcode = 0 does not become 0 even if suppressing unusedFunction error using suppress option.
Since it can not be helped, I adjust the option so as not to check unusedFunction.
In addition, we modified the code for the variableScope error and deleted it from the suppress option.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -32,11 +32,12 @@ cppcheck:
 	cppcheck --quiet --error-exitcode=1 \
             --inline-suppr \
             --std=c++03 \
+            -D HAVE_ATTR_XATTR_H \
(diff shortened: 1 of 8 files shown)


ggtakec:
About #704, #710, #712, #714, #724, #726 PR.

I'm very sorry for my late reply.
First, apply this code modification PR to make those PR's build on Travis CI clean.
And I will immediately respond to those PRs.

PR #729, opened 2018-03-02
FreeBSD build fixes

dmgk:
Fixes to get v1.83 building on FreeBSD 10 and 11 i386/amd86.


ggtakec:
@dmgk Thanks for your kindness.

Issue #732, opened 2018-03-13
Mounting S3 Bucket Fails with "same bucket(default)" error


phxvyper:
#### Version of s3fs being used (s3fs --version)
_1.83_

#### Version of fuse being used (pkg-config --modversion fuse)
_2.9.4_

#### System information (uname -r)
_4.4.0-1052-aws_

#### Distro (cat /etc/issue)
_Ubuntu 16.04.3 LTS_

#### s3fs command line used (if applicable)
```
sudo s3fs wisedotnet-projects /media/projectbucket -o passwd_file=/etc/passwd-s3fs -o dbglevel=info -f -o curldbg
```
#### /etc/fstab entry (if applicable): 
```
s3fs#wisedotnet-project /media/projectbucket fuse _netdev,allow_other,use_cache=/tmp/cache,uid=1000,gid=1000 0 0
```
#### s3fs syslog messages (grep s3fs /var/log/syslog, or s3fs outputs)  
```
[CRT] s3fs.cpp:set_s3fs_log_level(272): change debug level from [CRT] to [INF]
[INF]     s3fs.cpp:set_mountpoint_attribute(4209): PROC(uid=0, gid=0) - MountPoint(uid=0, gid=0, mode=40777)
s3fs: same bucket(default) passwd setting found in passwd file.
```
### Details about issue

Mounting S3 Bucket at command line using s3fs (as shown above) prints off the error `s3fs: same bucket(default) passwd setting found in passwd file.` and doesnt mount.

Mounting via fstab and running `mount -a` produces the same error.

phxvyper:
I solved this. The problem was that there are multiple entries in the passwd file that match the default/no bucket.

Ultimately the confusion from this issue came from the fact that the error message is not very helpful. I managed to decipher what the error message meant by looking at the code that leads to that message:

```cpp
if(resmap.end() != resmap.find(bucket)){ // this line helped me understand what the message meant.
      S3FS_PRN_EXIT("same bucket(%s) passwd setting found in passwd file.", ("" == bucket ? "default" : bucket.c_str()));
      return -1;
}
```

I'm going to submit a PR for this issue to make this message more descriptive and helpful.

PR #733, opened 2018-03-13
More useful error message for dupe entries in passwd file

phxvyper:
#732 

### Details
Error message is not very helpful in determining the cause of the error. Improved message by introducing clarity on the problem and what the problem's source is.


ggtakec:
@phxvyper @gaul Thnaks for all.

PR #753, opened 2018-04-23
fix xpath selector in bucket listing

cfz:
the original implementation in get_base_exp() depends on the order of xml return from the server.
patriotically, when listing a directory with sub directory(s), the xml document response contains more than 2 <Prefix> nodes(some of them are in <CommonPrefixes> node).
the source code arbitrarily select the first one in the documents (nodes->nodeTab[0]->xmlChildrenNode), which is OK for aws.
however, some s3 compatible service return the list-bucket result in different order, leading the s3fs to a wrong behavior

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -2724,7 +2724,7 @@ static xmlChar* get_base_exp(xmlDocPtr doc, const char* exp)
 {
   xmlXPathObjectPtr  marker_xp;
   string xmlnsurl;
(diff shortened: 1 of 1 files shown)


ggtakec:
@cfz When s3fs supported distributed storage other than AWS, it was overlooked that this code is subject to modification.
Thank you for giving me a hand.

cfz:
@ggtakec  sure thing :)

PR #756, opened 2018-05-06
Optimize defaults

orozery:
### Details

This PR includes 3 changes to default s3fs configuration, in order to boost performance:

1. Change readme to recommened OpenSSL instead of gnuTLS (I tested openssl to be much faster)
2. Increase the stat cache size from 1000 to 100000. This greatly reduces the number of HEAD requests made by s3fs, as many use-cases list buckets of more than 1000 objects.
Each entry in the cache takes upto 0.5KB, so a full cache will take upto 0.5KB*100000 = 50MB of memory.
3. set FUSE sync_read by default (instead of the default async_read).
   This can increase performance as async_read causes the creation of background threads that can create a bottleneck when reaching FUSE's max_background. On the other hand, you don't get the benefit when using async_read since s3fs allows only a single thread to read from a file at a time.
I verified that this patch increases the performance in the case of multiple threads reading different files at the same time.

@gaul @ggtakec Thanks in advance!

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Installation
 On Ubuntu 14.04:
 
 ```
-sudo apt-get install automake autotools-dev fuse g++ git libcurl4-gnutls-dev libfuse-dev libssl-dev libxml2-dev make pkg-config
+sudo apt-get install automake autotools-dev fuse g++ git libcurl4-openssl-dev libfuse-dev libssl-dev libxml2-dev make pkg-config
 ```
 
 On CentOS 7:
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -142,7 +142,7 @@ pthread_mutex_t StatCache::stat_cache_lock;
 //-------------------------------------------------------------------
 // Constructor/Destructor
 //-------------------------------------------------------------------
-StatCache::StatCache() : IsExpireTime(false), IsExpireIntervalType(false), ExpireTime(0), CacheSize(1000), IsCacheNoObject(false)
+StatCache::StatCache() : IsExpireTime(false), IsExpireIntervalType(false), ExpireTime(0), CacheSize(100000), IsCacheNoObject(false)
 {
   if(this == StatCache::getStatCacheData()){
     stat_cache.clear();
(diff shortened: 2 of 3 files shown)


ggtakec:
HI, @orozery 
Thanks for your help.

1. There was not a big reason for recommending the OpenSSL version of CURL.  
To tell the truth, it was because the test of s3fs was mainly conducted with OpenSSL.  
I think that GnuTLS may be recommended, but I think that CURL installed with RPM such as CentOS is mainly NSS, so I am concerned about this point.  
Since the environment varies depending on the user, it seems better to describe 3 in the README, and clearly stated that there is an advantage in GnuTLS.
1. I thought that it would be better to make it larger than before. I think that is good.
1. There are places to worry about.  
When accessing the same file from multiple client processes, I think that it will be access from multiple threads, but is this also affected by changing to sync_read?

orozery:
@ggtakec Thanks for you comments!

1. On CentOS  you already recommend installing openssl-devel. Why do recommend libcurl4-gnutls-dev on Ubuntu 14.04? I tested and found that libcurl4-openssl-dev is much faster.

3. I tested reading from multiple client threads (using pread), and via different processes, and found that sync_read still allows to read concurrent access.
The only place where sync_read comes to play is when the user is asking to read more than 131072 bytes. In this case, FUSE will split the read request to small sub-requests of size 131072. In the case of async_read, these requests will be processed asynchronously, and this is the case I'm trying to prevent, since it creates a bottleneck on the fdent_lock mutex. Additionally, the processing of the sub-requests is not guaranteed to be in-order, which causes inefficiency in s3fs pre-fetch mechanism (in FdEntity::Load function).

To summarize, I see only advantages in turning on sync_read by default.

orozery:
@ggtakec I added another proposed change:
- enable direct_io by default

Here's the comment I added:

// by default, FUSE uses the kernel buffer cache for each opened file,
  // and flushes the cached file data upon file close (unless -o kernel_cache is specified).
  // the kernel cache is not needed since s3fs maintains its own cache
  // thus, we set direct_io to avoid this double caching and increase performance.

I see that travis failed with this new change. Will check why tomorrow.

ggtakec:
@orozery There is no problem as the cache size increase is insignificant compared to the current leak problem.
Let's fix gcrypt later if there is a problem.
Thanks.

Issue #284, opened 2015-10-22
bucket with dot in name with HTTPS SSL


neilmillard:
- SSL: certificate subject name '*.s3.amazonaws.com' does not match target hostname 'abc.example.com.s3.amazonaws.com'

Fails with curlCode: 51 msg: SSL peer certificate or SSH remote key was not OK.

I suppose this is valid following the RFC for wildcard certificates.
Would this leave me with the PR #190 fix?

---

[root@vagrant-centos65 var]# s3fs abc.example.com:/dev/web1 /var/www -o url=https://s3.amazonaws.com -d -d -f -o f2 -o curldbg
    set_moutpoint_attribute(3379): PROC(uid=0, gid=0) - MountPoint(uid=0, gid=0, mode=40755)
FUSE library version: 2.9.3
nullpath_ok: 0
nopath: 0
utime_omit_ok: 0
unique: 1, opcode: INIT (26), nodeid: 0, insize: 56, pid: 0
INIT: 7.13
flags=0x0000b07b
max_readahead=0x00020000
s3fs_init(2650): init
s3fs_check_service(2968): check services.
    CheckBucket(2367): check a bucket.
    prepare_url(174): URL is https://s3.amazonaws.com/abc.example.com/dev/web1/
    prepare_url(204): URL changed is https://abc.example.com.s3.amazonaws.com/dev/web1/
    RequestPerform(1572): connecting to URL https://abc.example.com.s3.amazonaws.com/dev/web1/
- About to connect() to abc.example.com.s3.amazonaws.com port 443 (#0)
-   Trying 54.231.140.10... \* connected
- Connected to abc.example.com.s3.amazonaws.com (54.231.140.10) port 443 (#0)
- Initializing NSS with certpath: sql:/etc/pki/nssdb
-   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
- SSL: certificate subject name '*.s3.amazonaws.com' does not match target hostname 'abc.example.com.s3.amazonaws.com'
- NSS error -12276
- Closing connection #0
- SSL peer certificate or SSH remote key was not OK
  RequestPerform(1723): ###curlCode: 51  msg: SSL peer certificate or SSH remote key was not OK

RobbKistler:
You could try `-o use_path_request_style`.

neilmillard:
Thanks Robbkistler, would that use the https://s3.amazonaws.com/abc.example.com/dev/web1/ path.
[...]

PR #761, opened 2018-05-16
Simplify installation for Ubuntu 16.04

gaul:
Also reorganize installation vs. compilation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,24 @@ Features
 Installation
 ------------
 
+Some platforms provide pre-built binaries:
+
+* On Ubuntu 16.04 or newer:
+
+```
+sudo apt-get install s3fs
+```
+
+* On Mac OS X, install via [Homebrew](http://brew.sh/):
+
+```ShellSession
+$ brew cask install osxfuse
+$ brew install s3fs
+```
+
+Compilation
+-----------
+
 * On Linux, ensure you have all the dependencies:
 
 On Ubuntu 14.04:
@@ -45,13 +63,6 @@ cd s3fs-fuse
 ./configure
 make
 sudo make install
-```
-
-* On Mac OS X, install via [Homebrew](http://brew.sh/):
-
-```ShellSession
-$ brew cask install osxfuse
-$ brew install s3fs
 ```
 
 Examples


orozery:
@gaul this won't install the latest s3fs version, right?

gaul:
Newer versions of Ubuntu track the later s3fs versions, with 18.04 packaging 1.82 but 16.04 packaging 1.79:

https://packages.ubuntu.com/bionic/s3fs
https://packages.ubuntu.com/xenial/utils/s3fs

Homebrew packages the latest 1.83:

https://formulae.brew.sh/formula/s3fs

I would like to recommend that users install the distro packages which will provide a better experience for non-developers.  However we should continue to provide the compilation instructions for all users if they have a particular bug.  This gets a bit dicey when we know about a data loss bug but I believe we fixed most of those before 1.79.

orozery:
There are bug fixes and many performance improvements (certainly when comparing 1.79 to 1.83 or the master branch).
We can include instructions on installing from a package, but we should add a disclosure (I think we should recommend installing from source as you get the latest version).

gaul:
Software is never finished and a given distribution ships lots of software that does not have the most recent version number.  I do not see anything unique about s3fs that would cause us to recommend that everyone compiles from source.  However, for data loss bugs we may want to notify LTS releases like Ubuntu 18.04.

orozery:
@gaul and can't we update the ubuntu package to a newer s3fs version? (like `pip install <python_package>` always gets you the latest version)

ggtakec:
@gaul @orozery We will update the README from time to time if there are additional notes, so I will merge this PR.
I hope I can do something like LTS releases ...

gaul:
@orozery C++ package managers never took off in the Linux world the way `pip` did for Python and other non-compiled languages, likely because the problem is more difficult.

spolischook:
This is no longer true. Installation from apt is not working on Ubuntu 16.04.
How ever compiled version works very well

gaul:
@spolischook Could you share the output of the failed apt-get install?

spolischook:
Sorry, I have got another issue with dots in backet name - https://github.com/s3fs-fuse/s3fs-fuse/issues/284
Now it works.
And, yea, there was no error code, command finish with success, however it's all about https://github.com/s3fs-fuse/s3fs-fuse/issues/284

PR #764, opened 2018-05-22
Remove false multihead warnings

orozery:
### Details
I encountered a few false warning messages (when setting dbglevel=warn):

1. `s3fs.cpp:append_objects_from_xml_ex(2575): contents_xp->nodesetval is empty.
this happens whenever the XML listing does not contain either <Contents> or <CommonPrefixes>, which is very very common.

2. `curl.cpp:MultiRead(3956): failed a request(404: ...` and `curl.cpp:MultiPerform(3928): thread failed - rc(-2)`
These warnings appear whenever we get a 404 response on a HEAD request sent as part of `readdir_multi_head`. This can happen if the object is not yet updated in the bucket listing, or if the object is a CommonPrefix.

This PR removes those false warnings.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -3899,11 +3899,14 @@ int S3fsMultiCurl::MultiPerform(void)
 {
   std::vector<pthread_t>   threads;
   bool                     success = true;
+  bool                     isMultiHead = false;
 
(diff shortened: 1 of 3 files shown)


ggtakec:
@orozery Thanks for your help. and It seems that there is no problem.

PR #783, opened 2018-07-02
Issues with listing collections that contains over 10 000 objects

pawelmarkowski:
We were not able to perform ls in a bucket that contains over 10 000 objects. This objects starts with specified key: for instance ls /test/2018/06/02/ contains over 12 000 objects that simulates folders of our products.
### Details
We increase cache size because we faced a deadlock issue. We change max-keys option, which should be in our opinion configurable as run parameter (if user does not use max-keys, then it should be 1 000 that is the default value for S3.). Our quick fix increases max-keys value from 1000 to 15 000, and retries from 3 to 5 but that are the minor changes that should increase platform performance and stability.


ggtakec:
@pawelmarkowski Thank you for PR.

As you explained, it seemed necessary to change max-keys according to the user's environment.
However, I think that this value should not be hard-coded but should be an option.
(For example, list_object_max_keys)

About the other two changes(changing the default value):
I thought it was a good thing to change retries from 3 to 5.
We have lots of cases where retries changed from default 3.

And I think that it is better not to change the part that changes the default value of StatCache from 100000 to 300000.
It recently changed from 1000 to 100000( #756 ), and the s3fs usage memory size is also increasing.
And this value can be changed with max_stat_cache_size option.

To change these values, you should specify with each option.

Based on this PR, I make a new PR and make it reflect immediately as optionization of max_keys and change of retries default value.
(Along with this, we also change the man page, help output, etc.)

Regards,

ggtakec:
@ pawelmarkowski I merged #793 based on your this PR.
Try using the new options list_object_max_keys and max_stat_cache_size, retries options.
This PR will be closed.

Thank you for your kindness.

pawelmarkowski:
Hello @ggtakec it looks perfect right now. As I wrote above I was expecting that my changes will not be hardcoded but available as an options. Great job. I will make some tests ASAP and give you an answer.

PR #786, opened 2018-07-04
Log messages for 5xx and 4xx HTTP response code

ambiknai:

If there are Issues related to this PullRequest, please list it.

### Details
Currently S3FS fuse is not logging any message for 4xx and 5xx HTTP response codes.  
The code change enable the logging for 4xx and 5xx  codes.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -2063,6 +2063,7 @@ int S3fsCurl::RequestPerform(void)
           return 0;
         }
         if(500 <= LastResponseCode){
+          S3FS_PRN_ERR("HTTP response code = %ld Body Text: %s", LastResponseCode, (bodydata ? bodydata->str() : ""));
           S3FS_PRN_INFO3("HTTP response code %ld", LastResponseCode);
           sleep(4);
           break; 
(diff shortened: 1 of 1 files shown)


ambiknai:
@nkkashyap  @orozery Please could you review the changes

orozery:
2 comments:
1. 404 Should not be treated as error, so keep the info and the debug print, without error printing.
2. For the other cases, you should remove the info & debug print, as they are redundant with respect to the new error printing.

@ambiknai Other than that, looks good!

orozery:
@ambiknai Looks good to me. Thanks!

src/curl.cpp
```diff
@@ -2071,11 +2071,11 @@ int S3fsCurl::RequestPerform(void)
         // Service response codes which are >= 400 && < 500
         switch(LastResponseCode){
           case 400:
-            S3FS_PRN_ERR("HTTP response code %ld, Returning EIO, Body Text: %s", LastResponseCode, (bodydata ? bodydata->str() : ""));
+            S3FS_PRN_ERR("HTTP response code %ld, returning EIO. Body Text: %s", LastResponseCode, (bodydata ? bodydata->str() : ""));
```
ZVampirEM77: -EIO is more suitable, right?

src/curl.cpp
```diff
@@ -2071,11 +2071,11 @@ int S3fsCurl::RequestPerform(void)
...
             return -EIO;
 
           case 403:
-            S3FS_PRN_ERR("HTTP response code %ld, Returning EPERM, Body Text: %s", LastResponseCode, (bodydata ? bodydata->str() : ""));
+            S3FS_PRN_ERR("HTTP response code %ld, returning EPERM. Body Text: %s", LastResponseCode, (bodydata ? bodydata->str() : ""));
```
ZVampirEM77: Same

src/curl.cpp
```diff
@@ -2084,7 +2084,7 @@ int S3fsCurl::RequestPerform(void)
             return -ENOENT;
 
           default:
-            S3FS_PRN_ERR("HTTP response code %ld, Returning EIO, Body Text: %s", LastResponseCode, (bodydata ? bodydata->str() : ""));
+            S3FS_PRN_ERR("HTTP response code %ld, returning EIO. Body Text: %s", LastResponseCode, (bodydata ? bodydata->str() : ""));
```
ZVampirEM77: Same

ggtakec:
@ambiknai @orozery @sqlbot @ZVampirEM77 Thanks all. I'm going to merge this.

PR #800, opened 2018-07-13
FdManager::Rename

bruce-lz:
1.use_cache not rename cachefile, rename_object has delete rename source cachefile
2.not use_cache don't find FdEntity,because eky have perfix NOCACHE_PATH_PREFIX_FORM.

Signed-off-by: zhong.li <lz_bruce@126.com>
Signed-off-by: puchun.shen <spcjv5@gmail.com>

### Details

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -2141,19 +2141,45 @@ FdEntity* FdManager::ExistOpen(const char* path, int existfd, bool ignore_existf
   return ent;
 }
 
-void FdManager::Rename(const std::string &from, const std::string &to)
+void FdManager::Rename(const string& from, const string& to)
 {
   AutoLock auto_lock(&FdManager::fd_manager_lock);
-  fdent_map_t::iterator iter = fent.find(from);
-  if(fent.end() != iter){
+  fdent_map_t::iterator iter;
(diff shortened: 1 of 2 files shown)


gaul:
@bruce-lz Can you explain what this pull request does more clearly?

ggtakec:
@bruce-lz Thanks for your PR and I'm sorry for my late reply.

The purpose of this PR is to rename the cache file when the file is renamed, and to continue using the cache even after the rename, isn't it?
I think that efficiency will improve as behavior after rename!

First of all, I think that it is unnecessary to rename if there is no cache directory(not using the cache).
It is handled by the s3fs process that was started with the cache directory enabled.

Next, when renaming the cache file, I think it should be implemented with the method of FdEntity class.
There are cases where FdEntity class has information on the cache file, and you should avoid changing the cache file path directly from other than FdEntity class.
I think it should be implemented as a method such as FdEntity::RenameCachePath, for example.
(About the operation of the cache of s3fs, even if the cache file of the old path remains, it will be recreated when it is newly cached with the same filename.)

Last, please change the iter variable name as the version of cppcheck is upgraded.

Thanks in advance  for your help.

gaul:
@bruce-lz Do you plan to continue this pull request?

Issue #808, opened 2018-08-08
Random write behavior


kunallillaney:
I had a few questions on how s3fs handles random writes since I seeing some weird network stats (given the disclaimer in the README and this subsequent post #607)

Background:
* Setup: Existing file on S3 - 100MB. I am using fio for random writes on this and monitoring the network I/O using slurm.
* From my understanding, S3FS prefetches data in 40MB (multipart block size* threads) blocks into the cache when a region is a read. This is verified by slurm and the sparse file in the cache.
* However, when I perform random writes on S3FS, say changing a single byte, no data is fetched either into the cache or shown by slurm (double checked with /proc/net/dev/) but file size worth of data (100MB in this case) does get transmitted to S3. Here s3fs is not overwriting the other parts of the file with garbage data but correctly updating the single byte.

Question:
* In the random write case, my assumption was s3fs did a multipart-download, merge changes and multipart-upload. Is this not the case? Does it perform multipart-copy for unchanged regions for S3 (similar to the operation for rename) but the network stats do not support this since it should have only transmitted 40MB worth data and copied the other 60MB. Am I missing something?

Thanks.

kunallillaney:
And a follow on question is if does it indeed use multi-part copy, then how does it get around the [5GB target file size limit](https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPartCopy.html) for range reads placed by S3.

gaul:
Which options do you mount s3fs with?  Have you enabled the data cache?  This could explain the read behavior you experience.

For writes, s3fs does issue multipart uploads and copies data for the unchanged regions.  Hence you should expect a 1-byte update to a 100 MB file to copy (9) 10 MB chunks and upload a whole 10 MB chunk.  If the cache does not contain this range it will have to download it before uploading the whole part.

kunallillaney:
The data cache is enabled. I tried this by opening the file in "r+" mode to update. In this case, the file ends up with correct data in the updated region and zeros everywhere else (which sounds like a bug to me). I don't think s3fs copies data for the other 9 chunks. If it does, can you please explain how it gets around the 5GB file size limit I mention above? S3FS needs to do range reads on the remaining 90MB of data to do so but S3 does not allow it since the target file is only 100MB.
I believe this was verified by @orozery as well.

gaul:
[...]

PR #820, opened 2018-09-11
Enable big writes if capable

### LINKED ISSUE #813: FUSE request size
Is there a way to change the FUSE request size for S3FS? As of now, it seems that s3fs-fuse only supports 4KB request sizes in FUSE. But, libfuse does allows large I/O sized requests, up to 128KB ([Reference](http://libfuse.github.io/doxygen/fast17-vangoor.pdf)).

Is there a plan on the roadmap to expose this as a user-defined parameter?

gaul:
Fixes #813.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -3420,6 +3420,10 @@ static void* s3fs_init(struct fuse_conn_info* conn)
   }
   #endif
 
+  if((unsigned int)conn->capable & FUSE_CAP_BIG_WRITES){
+     conn->want |= FUSE_CAP_BIG_WRITES;
+  }
+
   return NULL;
 }
 


kunallillaney:
| Mode  | Read Bandwidth (MB/sec) | Write Bandwidth (MB/sec) |
| ------------- | ------------- | ------------- |
| Current version  | 175.9 | 36.4  |
| Big writes enabled | 213.8 | 70.8 |

I ran the benchmark using FIO (twice for average) on a AWS m5.4xlarge instance. S3FS was mounted with 4 threads, 16MB part size and ample RAMdisk as cache. FIO was run for a 1.2 GB file, block size of 4MB (will default to 128KB because of FUSE) and fysnc on close.

kunallillaney:
@gaul Thanks for fixing this but there is a bug for random writes as well. See #808. Is this known or just that no one has come across this before?

gaul:
Please keep this pull request focused on only the issue at hand and I responded in the other issue.  I would appreciate if you deleted the comment here.

kahing:
why did big writes increase read bandwidth? seems like that has something to do with the unstable benchmark environment

kunallillaney:
@kahing The environment is as kosher as I can make it to be. S3 reads do tend to show high variance and this was meant to be quick not through. It was an average of two for a benchmark which ran over a short period of time. Feel free to present your numbers from a stable benchmark environment.

kahing:
Did you benchmark from an empty bucket? I've actually done extensive benchmarking and my result is much different from yours: https://raw.githubusercontent.com/kahing/goofys/master/bench/bench.png

S3 does shard the buckets when there's more data so that could explain the difference.

kahing:
(my result didn't include this PR)

kunallillaney:
@kahing Yes I have seen those and the corresponding benchmark script. I am not a big fan of ```dd``` and prefer to use more detailed file system bench marking applications like [fio](https://github.com/axboe/fio), [IOzone](http://www.iozone.org/) or [filebench](https://github.com/filebench/filebench). 

(Here I speak only for the read and write benchmarks not the metadata ones)
There are multiple factors which can cause different benchmark results. This ranges from the network bandwidth on your AWS node (this was not a factor for these benchmarks since your pipe was 1.2 GB/sec and my pipe is atleast 350MB/sec), to the number of objects in the bucket (sharding as you point out) and the setup of each file system. The setup of each file system is important since there are different tweaks one can perform to make them go faster.

I provide some insight into why these benchmarks might be different apart form sharding of the S3 bucket. I do use a non-empty bucket here but one which had about 100 objects of ~1GB each and am not sure if that had a major role to play.
**S3 part sizes**:
You don't make mention of what S3 part sizes you use for S3FS. If you use the default ones then that is 10MB while I use 16MB. This may seem not important but there is a fine balance in amortizing your RESTful cost for each I/O access thus you want part sizes to be larger but not large enough that it affects parallelism.
**Cache**:
You don't use a cache for S3FS while I have one with RAMdisk. In absence of that option, s3fs defaults to using /tmp/ which is disk while these are on memory. To be clear, the performance gain comes from this part of S3FS for writes. I have separate benchmarks (cannot post those here) which benchmark the s3fs I/O bandwidth purely to RAMdisk and enabling big writes results in an increase in I/O bandwith here. The physical memory you are running this on also matters. I ran this on an old machine with DDR2 and it was much slower while the new AWS machines with DDR4 are way faster.  Also, I do ensure that the cache is cleared before I re-ran these experiments and ensured I measured from start of write, inclusive of close, since s3fs flushes on close.

kahing:
definitely agree that part sizes impact performance. I intentionally tested the defaults because that's what most people would use (and also I don't believe there's a "correct" setting even for a particular user).

anyways, agree that read variance is a problem but the same exists in write too. In my benchmark scripts I run each test 10 times (data here: https://github.com/kahing/goofys/blob/master/bench/bench.s3fs)

kunallillaney:
@ggtakec Thanks. I think even after this updates, s3fs can only utilize a fraction of the I/O bandwidth available to it from the cache (assuming the cache is RAMdisk or SSDs). Part of this issue is the relatively small size of I/O requests (max of 128KB) from FUSE and overhead of opening and closing a file for these small accesses to cache.

PR #831, opened 2018-09-30
Add support for storage class ONEZONE_IA.

JoohnSF:
### Details
Adds support for the ONEZONE_IA storage class, which was introduced in AWS in April 2018.


ggtakec:
@JoohnSF Thanks for your kindness.

PR #832, opened 2018-10-02
Simplify hex conversion

gaul:
Addresses GCC 8 warning:

```
common_auth.cpp: In function ‘std::__cxx11::string s3fs_sha256sum(int, off_t, ssize_t)’:
common_auth.cpp:84:12: warning: ‘char* strncat(char*, const char*, size_t)’ output may be truncated copying 2 bytes from a string of length 2 [-Wstringop-truncation]
    strncat(sha256, hexbuf, 2);
```


ggtakec:
@gaul Thanks.

Issue #785, opened 2018-07-04
Fedora packaging


gaul:
It would be great if Fedora included `s3fs` in its package repository for parity with the Ubuntu package.  Can someone step forward to write a spec file and submit it to Red Hat?  I found this issue which is open:

https://bugzilla.redhat.com/show_bug.cgi?id=725292

@dreampuf has a spec file mentioned in #648 but never submitted a pull request.

juliogonzalez:
If you are interested, I think I can take care of it, since I already have https://github.com/juliogonzalez/s3fs-fuse-rpm, including CI with docker containers.

gaul:
@juliogonzalez It would be great if you could get Fedora to include `s3fs`!

Also it looks like SUSE already packages `s3fs`:

https://software.opensuse.org/package/s3fs

Could you add installation instructions to our `README.md`?

juliogonzalez:
@gaul, about your request for SUSE and openSUSE, see #789.

I can also temporary add the instructions to use my repository, as it's able to build for Fedora, CentOS, Amazon Linux.

About adding the packages to Fedora, please see my comment at https://bugzilla.redhat.com/show_bug.cgi?id=725292#c31

I guess we should also submit it to [EPEL](https://fedoraproject.org/wiki/EPEL), to make it available for RHEL7/CentOS7 (as you can see at my repository RHEL6/CentOS6 is another story because of the old FUSE version), and that should make it also available for recent Amazon Linux versions.

juliogonzalez:
I fixed several things after reviewing Fedora doc, see https://github.com/juliogonzalez/s3fs-fuse-rpm/commits/master

Submission is now ready at https://bugzilla.redhat.com/show_bug.cgi?id=1631988 (provided I did not miss any steps of the looooooooooong procedure for Fedora submissions).

I also added a webhook at https://github.com/juliogonzalez/s3fs-fuse-rpm, so for each commit for master will be built at copr: https://copr.fedorainfracloud.org/coprs/juliogonzalez/s3fs-fuse/

Koji build is working for the latest copr build: https://koji.fedoraproject.org/koji/taskinfo?taskID=29814931

Now I am waiting for review to see if I need to change anything, and trying to see if I can get a sponsor. I will present myself at Fedora's devel list and we'll see.

@gaul, can you assign this issue to me?

gaul:
Awesome, thank you for pushing this forward!

juliogonzalez:
[...]

PR #833, opened 2018-10-06
New installation instructions for Fedora >= 27 and CentOS7

juliogonzalez:
### Relevant Issue

#785 

### Details
Since I submitted s3fs-fuse for Fedora 27, Fedora 28, Fedora 29 and EPEL7, I am adding the instructions about how to install, and removing the compilation details for CentOS7.

Compilation details for Fedora and CentOS are still available at https://github.com/s3fs-fuse/s3fs-fuse/wiki/Installation-Notes#fedora--centos

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,17 @@ Some systems provide pre-built packages:
   sudo zypper in s3fs
   ```
 
+* On Fedora 27 and newer:
+  ```
+  sudo yum install s3fs-fuse
+  ```
+
+* On RHEL 7 and CentOS 7 trough EPEL 7 repositories:
+  ```
+  sudo yum install epel-release
+  sudo yum install s3fs-fuse
+  ```
+
 * On Mac OS X, install via [Homebrew](http://brew.sh/):
 
   ```ShellSession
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -36,6 +36,17 @@ Some systems provide pre-built packages:
   sudo zypper in s3fs
   ```
 
+* On Fedora 27 and newer:
+  ```
+  sudo yum install s3fs-fuse
+  ```
+
+* On RHEL7 and CentOS7 trought EPEL7 repositories:
```
gaul: Please correct spelling: through and add space between distro and 7.

README.md
```diff
@@ -54,12 +65,6 @@ On Ubuntu 14.04:
 sudo apt-get install automake autotools-dev fuse g++ git libcurl4-openssl-dev libfuse-dev libssl-dev libxml2-dev make pkg-config
 ```
 
-On CentOS 7:
```
gaul: Should we remove all compilation instructions from the README and link to the wiki instead?

juliogonzalez: I guess that depends on the policy you want to follow about it :-)

My understanding is that you had CentOS 7 and Ubuntu 14.04 because neither of them had s3fs packages but it was easy to compile them without doing special tricks (such as compiling and installing fuse, as it's required for CentOS 6)

IMHO, I'd keep the compilation instructions, but only generic, and I'd point to the wiki for specific instructions for each distribution.

I can take care of that on a separate PR if you want.

gaul: Originally we only had compilation instructions until volunteers like you :) contributed packages.  I am fine either keeping all the compilation instructions or moving them to the wiki.  Ubuntu 14.04 will end its support in six months so this should not be top of mind any more!

PR #834, opened 2018-10-06
Improve template for issues

juliogonzalez:
### Details
Now that we have binary packages for Debian, Ubuntu, Fedora and RHEL/CentOS, we can improve the template for issues a little. Details on in-line comments.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,25 +1,26 @@
 ### Additional Information
 _The following information is very important in order to help us to help you. Omission of the following details may delay your support request or receive no attention at all._
+_Keep in mind that the commands we provide to retrieve information are oriented to GNU/Linux Distributions, so you could need to use others if you use s3fs on MacOS or BSD_
 
 #### Version of s3fs being used (s3fs --version)
 _example: 1.00_
 
-#### Version of fuse being used (pkg-config --modversion fuse)
+#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
 _example: 2.9.4_
 
-#### System information (uname -r)
+#### Kernel information (uname -r)
 _command result: uname -r_
 
-#### Distro (cat /etc/issue)
-_command result: cat /etc/issue_
+#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
+_command result: cat /etc/os-release_
 
-#### s3fs command line used (if applicable)
+#### s3fs command line used, if applicable
 ```
 ```
-#### /etc/fstab entry (if applicable): 
+#### /etc/fstab entry, if applicable
 ```
 ```
-#### s3fs syslog messages (grep s3fs /var/log/syslog, or s3fs outputs)  
+#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
 _if you execute s3fs with dbglevel, curldbg option, you can get detail debug messages_
 ```
 ```


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,25 +1,26 @@
...
 _example: 2.9.4_
 
-#### System information (uname -r)
+#### Kernel information (uname -r)
 _command result: uname -r_
 
-#### Distro (cat /etc/issue)
-_command result: cat /etc/issue_
+#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
+_command result: cat /etc/os-release_
```
juliogonzalez: `/etc/issue` is not about the Distribution, but about the MOTD for the system.

In the first place, it can be replaced by the system administrator.

But besides, on Fedora/RHEL/CentOS does not even show the information about the operating system (it has placeholdes by default, so the file is not useful by itself).

On the other hand `/etc/os-release` is available for a lot of GNU/Linux distributions (but not of all them, of course): Ubuntu, Debian, Fedora, SUSE, openSUSE, Fedora, RHEL, CentOS, Slackware, Alpine, Gentoo... since some years now.

For the cases where `/etc/os-release` is not available, most probably will have knowledge enough to get the distribution version.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,25 +1,26 @@
...
-#### s3fs command line used (if applicable)
+#### s3fs command line used, if applicable
 ```
 ```
-#### /etc/fstab entry (if applicable): 
+#### /etc/fstab entry, if applicable
 ```
 ```
-#### s3fs syslog messages (grep s3fs /var/log/syslog, or s3fs outputs)  
+#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
```
juliogonzalez: `/var/log/syslog` is not available anymore for some distributions using systemd, and in those cases journalctl is the replacement.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,25 +1,26 @@
 ### Additional Information
 _The following information is very important in order to help us to help you. Omission of the following details may delay your support request or receive no attention at all._
+_Keep in mind that the commands we provide to retrieve information are oriented to GNU/Linux Distributions, so you could need to use others if you use s3fs on MacOS or BSD_
```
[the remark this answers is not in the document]
ggtakec: I agree with your proposal.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,25 +1,26 @@
...
 
 #### Version of s3fs being used (s3fs --version)
 _example: 1.00_
 
-#### Version of fuse being used (pkg-config --modversion fuse)
+#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
```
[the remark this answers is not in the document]
ggtakec: If users get confused, let's write more than three command examples.
I think that this is fine now.
Thanks.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,25 +1,26 @@
...
 _example: 2.9.4_
 
-#### System information (uname -r)
+#### Kernel information (uname -r)
```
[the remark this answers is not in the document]
ggtakec: Sorry for it is my fault.

.github/ISSUE_TEMPLATE.md
ggtakec: I think that your modifications are correct and good.

.github/ISSUE_TEMPLATE.md
ggtakec: Thanks

PR #835, opened 2018-10-08
Make the compilation instructions generic

juliogonzalez:
#883

### Details
Make the compilation instructions generic for *nix, as discussed at #883

Check https://github.com/juliogonzalez/s3fs-fuse/blob/b2e2189aba94ea01083e446999089290dde022a5/README.md to see how the result looks like.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,8 @@ Features
 * user-specified regions, including Amazon GovCloud
 * authenticate via v2 or v4 signatures
 
-Installation
-------------
+Installation from pre-built packages
+------------------------------------
 
 Some systems provide pre-built packages:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,27 +54,35 @@ Some systems provide pre-built packages:
...
 
-```
-git clone https://github.com/s3fs-fuse/s3fs-fuse.git
-cd s3fs-fuse
-./autogen.sh
-./configure
-make
-sudo make install
-```
+   * fuse >= 1.84
```
gaul: Should this be 2.8.4?

README.md
```diff
@@ -54,27 +54,35 @@ Some systems provide pre-built packages:
   $ brew install s3fs
   ```
 
-Compilation
------------
+Compilation and installation from sources
+-----------------------------------------
 
-* On Linux, ensure you have all the dependencies:
+These are generic instructions to compile from the master branch, and should work on almost any GNU/Linux, Mac OS, BSD or similar.
```
gaul: Stylize as macOS.

juliogonzalez: Are you sure?

At the rest of the `README.md` file, it is `Mac OS`.

> # s3fs
> s3fs allows Linux and Mac OS X to mount an S3 bucket via FUSE. s3fs preserves the native object format for files, allowing use of other tools like s3cmd.

> # Installation
> - On Mac OS X, install via Homebrew:

I just now checked, and it seems `Mac OS` was the previous name, while currently it's `macOS`.

Maybe we can merge as it is, and then perform a `sed` (I see that at least `config.guess` still mentions `Mac OS X` as well).

README.md
juliogonzalez: Yes, it was a typo: I copied and pasted s3fs-fuse version instead of fuse version.

Fixed now.

README.md
gaul: I believe Apple has renamed to macOS.  I can address this in a subsequent commits since it is widespread.

ggtakec: It seems that Apple has changed "OS X(Mac OS)" to new name "macOS" from 10.12 Sierra.

juliogonzalez: Fixed at #840

PR #852, opened 2018-11-04
Allow credentials from ${HOME}/.aws/credentials

### LINKED ISSUE #822: s3fs should use a .aws/credentials profile
I racking my brain as to why s3fs would ask for a custom formatted custom file for aws credentials when there is one right on every facile aws users system at ~/.aws/credentials.

I for one rotate my keys regulary using a tool. I think it would be wise to just support the default way of doing things.

gaul:
This matches the configuration from popular tools like AWS CLI and
allows multiple profile names via `-o profile=name`.  The existing
credential mechanisms continue to work.  Fixes #822.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,10 @@ Keep in mind using the pre-built packages when available.
 Examples
 --------
 
+s3fs supports the standard
+[AWS credentials file](https://docs.aws.amazon.com/cli/latest/userguide/cli-config-files.html)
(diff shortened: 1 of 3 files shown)


juliogonzalez:
I think `doc/man/s3fs.1` will need this documented as well :-)

ggtakec:
@gaul @juliogonzalez Thanks.
And I added "profile" option to wiki page( https://github.com/s3fs-fuse/s3fs-fuse/wiki/Fuse-Over-Amazon#profile-defaultdefault )

PR #859, opened 2018-11-16
Upload S3 parts without batching

### LINKED ISSUE #183: ParallelMultipartUploadRequest suffers from head of line blocking
s3fs issues multipart uploads in batches (defaulting to 5) and does not issue the next batch until the current one completes.  If ones of these uploads is a straggler, it blocks subsequent uploads and hurt performance.  Instead of the blocking `curlmulti.Request` call s3fs should issues the next upload as soon as one upload completes.

gaul:
Previously s3fs would issue a batch of uploads and wait for all to
succeed before issuing the next batch.  Now it issues the first batch
and only waits for a single part to succeed before uploading the next
part.  This can improve performance when one part lags due to network
errors.  Fixes #183.


orozery:
@gaul It still looks like that MultiPerform will block until all threads are done (calling pthread_join in a for loop).
What am I missing?

gaul:
Previously `ParallelMultipartUploadRequest` issued batches of 5 requests to `MultiPerform` and waited for them to all complete.  Now it issues all requests to `MultiPerform` which issues a batch of requests then issues the next as soon as one completes.

orozery:
Got it now, thanks!
Very nice optimization!

PR #865, opened 2018-11-28
fix multihead warning check

orozery:
### Details
This PR fixes the condition for checking 404 responses on MultiHead requests.
The check should be for -ENOENT, instead of ENOENT.


ggtakec:
@orozery Thank you for your cooperation.

PR #874, opened 2018-12-16
Address cppcheck 1.86 errors

gaul:
Lifetime, shadowing, and unused variables.  Found via the Travis macOS
builder.


ggtakec:
@gaul and @Basavaraju013 Thanks all.

PR #882, opened 2019-01-05
[curl] Assume long encryption keys are base64 encoded

earlchew:
Amazon SSE-C https://tinyurl.com/ychug4cg writes:

> Use this header to provide the 256-bit, base64-encoded encryption key
> for Amazon S3 to use to encrypt or decrypt your data.

It seems likely that future keys would be as long, or longer, so
this change assumes that text longer than 32 characters are keys
that are base64 encoded.

Signed-off-by: Earl Chew <earl_chew@yahoo.com>

### Details

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -963,19 +963,38 @@ bool S3fsCurl::PushbackSseKeys(string& onekey)
   if('#' == onekey[0]){
     return false;
   }
-  // make base64
-  char* pbase64_key;
-  if(NULL == (pbase64_key = s3fs_base64((unsigned char*)onekey.c_str(), onekey.length()))){
-    S3FS_PRN_ERR("Failed to convert base64 from SSE-C key %s", onekey.c_str());
-    return false;
+  // make base64 if the key is short enough, otherwise assume it is already so
+  string base64_key;
+  string raw_key;
+  if(onekey.length() > 256 / 8){
+    char* p_key;
+    size_t keylength;
+
+    if(NULL != (p_key = (char *)s3fs_decode64(onekey.c_str(), &keylength))) {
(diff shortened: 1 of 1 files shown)


src/curl.cpp
```diff
@@ -963,19 +963,38 @@ bool S3fsCurl::PushbackSseKeys(string& onekey)
...
+	  return false;
+	}
+  } else {
+    char* pbase64_key;
+
+    if(NULL != (pbase64_key = s3fs_base64((unsigned char*)onekey.c_str(), onekey.length()))) {
+	  raw_key = onekey;
+      base64_key = pbase64_key;
+      free(pbase64_key);
+	} else {
```
gaul: Wonky indentation?

src/curl.cpp
```diff
@@ -963,19 +963,38 @@ bool S3fsCurl::PushbackSseKeys(string& onekey)
...
+  string raw_key;
+  if(onekey.length() > 256 / 8){
+    char* p_key;
+    size_t keylength;
+
+    if(NULL != (p_key = (char *)s3fs_decode64(onekey.c_str(), &keylength))) {
+      raw_key = string(p_key, keylength);
+	  base64_key = onekey;
+	  free(p_key);
+	} else {
```
gaul: Wonky indentation?

earlchew:
Well, not necessarily, there's a setting at the end of the file that claims tabs have width 4. In any case, I've replaced the tabs in the commit with whitespace.

```
/*
* Local variables:
* tab-width: 4
* c-basic-offset: 4
* End:
* vim600: noet sw=4 ts=4 fdm=marker
* vim<600: noet sw=4 ts=4
*/
```

PR #884, opened 2019-01-07
Store and retrieve file creation time

### LINKED ISSUE #771: Create time of files in bucket reported as zero
I noticed that the create time of all file in a bucket are set to zero:

`ubuntu@ip-172-31-12-14:/tmp$ ls -lc /bayesforgeS3models/csv/`
`total 36761`
`-rw-r--r-- 1 1001 root  1281706 Jan  1  1970 asia.csv`
`-rw-rw-r-- 1 1001 root   200046 Jan  1  1970 burglary20000.csv`
`-rw-rw-r-- 1 1001 root 36000089 Jan  1  1970 credit2000000.csv`
`-rw-r--r-- 1 1001 root   160043 Jan  1  1970 potato.csv`
`-rw-rw-r-- 1 1001 root        0 Jan  1  1970 rain1000.csv`

gaul:
This introduces a new header with the creation date; existing object
will report creation dates in 1969.  Fixes #771.

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -620,6 +620,9 @@ bool convert_header_to_stat(const char* path, headers_t& meta, struct stat* pst,
   // mtime
   pst->st_mtime = get_mtime(meta);
 
+  // ctime
+  pst->st_ctime = get_ctime(meta);
+
   // size
   pst->st_size = get_size(meta);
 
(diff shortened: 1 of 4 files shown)


src/cache.cpp
```diff
@@ -620,6 +620,9 @@ bool convert_header_to_stat(const char* path, headers_t& meta, struct stat* pst,
   // mtime
   pst->st_mtime = get_mtime(meta);
 
+  // ctime
+  pst->st_ctime = get_ctime(meta);
```
gaul: Calling this ctime seems wrong which should be metadata change time.  Instead call it btime for file birth?

gaul:
This is more complicated than I realized; s3fs should update mtime on data modification and ctime on metadata modification.  Closing for now.

PR #894, opened 2019-01-14
Default uid/gid/mode when object lacks permissions

### LINKED ISSUE #890: Consider 640 default permissions when x-amz-meta headers do not exist
Many users mount S3 buckets with existing objects which lack the s3fs-only `x-amz-meta-` headers.  These mount with 0000 permissions which frustrate users who need to specify the FUSE option `-o umask=037` to have permission to read their objects.  Users have reported this several times but #673 is a recent occurrence.  I propose that we give some default permission, perhaps 640, when these permission headers do not exist.  Investigate what [goofys](https://github.com/kahing/goofys) does before implementing.

gaul:
This addresses a common use case when interacting with objects from
both s3fs and other S3 tools.  Fixes #890.

### DIFF
--- a/src/s3fs_util.cpp
+++ b/src/s3fs_util.cpp
@@ -792,11 +792,13 @@ mode_t get_mode(headers_t& meta, const char* path, bool checkdir, bool forcedir)
 
   if(meta.end() != (iter = meta.find("x-amz-meta-mode"))){
     mode = get_mode((*iter).second.c_str());
+  }else if(meta.end() != (iter = meta.find("x-amz-meta-permissions"))){ // for s3sync
+    mode = get_mode((*iter).second.c_str());
+    isS3sync = true;
   }else{
-    if(meta.end() != (iter = meta.find("x-amz-meta-permissions"))){ // for s3sync
-      mode = get_mode((*iter).second.c_str());
-      isS3sync = true;
-    }
+    // If another tool creates an object without permissions, default to owner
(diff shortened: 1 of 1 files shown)


ggtakec:
I thought that it would be better not to allow access to files(objects) when meta does not exist, when permission is not given by options and so on.
However, as far as confusion of s3fs users is concerned, I am beginning to think that @gaul should respond.
I will merge this PR and we expect that user confusion will be reduced from the next version.

@gaul
Is the default permission not 0600(is too strict), 0640 is the best?
If there is a problem, we should open Issue about it.

gaul:
660, 640, and 600 all seem like reasonable permissions.  I picked 640 for parity with goofys but perhaps 600 make more sense depending how the group field is used in modern Linux.  On my system, I have both uid and gid gaul but shared systems have subgroups where this read permission may or may not be desired.  /cc @kahing

ggtakec:
@gaul In the future, if there is a problem at 0640, I would like to make it as Issue.

The default system umask for each OS is different, In addition to that there is a umask value set by the user.
For example in ubuntu, if the names of gid and uid are the same with a value of uid of 200 or more, it seems that system default umask is 0002, otherwise it is 0022.
Then the default permissions of objects that exist in S3 that do not have file permissions maybe depend on the umask value of each user.
But I think it should not depend on the user environment, static value as 0600 instead of it seems good. Because it will underestimate possible problems.

PR #899, opened 2019-01-17
Tighten up HTTP response code check

gaul:
Previously s3fs considered 3xx codes as successful.  When writing an
object to a misconfigured region, s3fs did not propagate the 301 as a
write error to the client.  References #693.


ggtakec:
@gaul Thanks, and I will change the error message for 300 series error later.
s3fs is output only "EIO" for it now, so I think it should output a more detailed response method.

Issue #807, opened 2018-08-07
response code 403 - Access denied - invalid credentials


LouNik1984:
### Additional Information
I'm testing s3fs on a virtual machine with ubuntu 16.04. I already have an s3 private bucket with a user which has a policy that allows access to everything: 

"Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": "s3:*",
            "Resource": "*"
        }

If I mount my already existing bucket, no issue at all, everything works as expected. 

But then I created a new bucket, attached this new bucket to the same user/policy/group as the previous one but when I try to mount this new bucket it doesn't work, I get the error HTTP error code 403 Access Denied Invalid Credentials.

But credentials are the same of the first bucket, so is the aws policy. So how is it that it works with one bucket but not with the other? How can I mount my second bucket?

#### Version of s3fs being used
Amazon Simple Storage Service File System V1.84 (commit:06032aa) with OpenSSL

#### Version of fuse being used
2.9.4

#### System information
4.13.0-36-generic

#### Distro
Ubuntu 16.04.4 LTS

#### s3fs command line used
/usr/bin/s3fs -f  -d mybucketname /mnt/s3 -o passwd_file=/etc/passwd-s3fs

tbooth:
Hi,

I think I am seeing the same problem. Connection to an existing bucket works, but if I make a new one, with the same credentials and permissions, it fails. I'm working in a non-default region (eu-west-1).

My guess here is that some time in the hours after the bucket is created, some metadata gets copied around the AWS system and the exact connection protocol for the bucket changes subtly. I can see that the initial connection to the existing bucket gets a "400 Bad Request" whereas a connection to the brand new bucket gets a "307 Temporary Redirect". In the first case, s3fs ends up correctly fixing its access credentials and logging in, but in the second case it never gets it right and ends up with a 404.

So it looks like there is some missing logic in s3fs regarding the authorization protocol. The following work-around worked for me, setting both options:

```-o endpoint=eu-west-1 -o url="https://s3-eu-west-1.amazonaws.com"```

tsmgeek:
> Hi,
> 
> I think I am seeing the same problem. Connection to an existing bucket works, but if I make a new one, with the same credentials and permissions, it fails. I'm working in a non-default region (eu-west-1).
> 
[...]

PR #912, opened 2019-01-20
Automatic region change made possible other than us-east-1(default)

ggtakec:
#908 #911 #807 

### Details
In the case of the default region us-east-1, when AuthorizationHeaderMalformed error occurred, automatic switching to the correct region was made.
However, when a region was specified with the endpoint option, this automatic switching was not executed.
With this new code, even if you specify a region with the endpoint option, try reconnecting to the correct region.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -134,7 +134,6 @@ static bool is_ibm_iam_auth       = false;
 static bool is_use_xattr          = false;
 static bool create_bucket         = false;
 static int64_t singlepart_copy_limit = FIVE_GB;
-static bool is_specified_endpoint = false;
 static int s3fs_init_deferred_exit_status = 0;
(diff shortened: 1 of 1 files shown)


ggtakec:
@sqlbot Thank you for your comments.
Certainly as you say, it seems that it is not good to automatically correct configuration misconfiguration.
For this PR, I will consider not to avoid automatic avoidance.

Until version 1.84, endpoint was automatically changed only when endpoint was not specified.
As a result, we might have given the user a situation not to understand why s3fs could connect.

I would like to think about reviewing the error message and clearly making it an error.
(Please wait a while until modification)

ggtakec:
@tbooth Thank you for your comment.
I will start to fix immediately.
Details are commented in #916, please  see it.

PR #913, opened 2019-01-20
Prefer abort over assert(false)

gaul:
The compiler can remove the latter when compiled with `NDEBUG` which may
cause unintended control flow.


ggtakec:
@gaul Thanks for your kindness.

PR #914, opened 2019-01-21
Issue readdir HEAD requests without batching

### LINKED ISSUE #223: readdir_multi_head suffers from head of line blocking
Similar to #183.

gaul:
Previously s3fs would issue a batch of HEAD requests and wait for all
to succeed before issuing the next batch.  Now it issues the first
batch and only waits for a single call to succeed before issuing the
next call.  This can improve performance when one call lags due to
network errors.  I measured 25% improvement with the same level of
parallelism.  This commit also reparents parallelism knobs for
consistency.  Follows on to 88cd8feb053980c808d67771d63a84ca25f6db8a.
Fixes #223.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -373,6 +373,7 @@ string           S3fsCurl::curl_ca_bundle;
 mimes_t          S3fsCurl::mimeTypes;
 string           S3fsCurl::userAgent;
 int              S3fsCurl::max_parallel_cnt    = 5;              // default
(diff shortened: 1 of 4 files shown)


gaul:
Easier to review when [ignoring whitespace](https://github.com/s3fs-fuse/s3fs-fuse/pull/914/files?w=1).

ggtakec:
@gaul Thanks!

PR #932, opened 2019-01-25
Prefer AutoLock for synchronization

gaul:
This simplifies the code and fixes an issue with unlocked access.
Also use a recursive lock for `StatCache` to avoid races between
lock..unlock..lock sequences.


ggtakec:
The code has become simple, thank you.

Issue #935, opened 2019-01-28
readdir performance regression since 1.84


gaul:
Using the goofys benchmark script, I observe much slower `ls_files` performance in master than 1.84 (results are time, lower is better):

1.84:

```
$ ~/work/go-workspace/src/github.com/kahing/goofys/bench/bench.sh '/home/gaul/work/s3fs-fuse/src/s3fs gaultest-us-west-1 /home/gaul/mnt /home/gaul/mnt ls
ls_files 2.913
ls_files 0.985
ls_files 0.829
ls_files 0.852
ls_files 0.808
ls_files 0.811
ls_files 1.093
ls_files 0.959
ls_files 0.933
ls_files 0.840
```

master:

```
$ ~/work/go-workspace/src/github.com/kahing/goofys/bench/bench.sh '/home/gaul/work/s3fs-fuse/src/s3fs gaultest-us-west-1 /home/gaul/mnt' /home/gaul/mnt ls
ls_files 16.186
ls_files 9.720
ls_files 11.963
ls_files 12.418
ls_files 10.156
ls_files 10.072
ls_files 9.872
ls_files 9.751
ls_files 9.615
ls_files 10.418
```

We should analyze this before releasing 1.85.  The script reveals other interesting results including improvements to bulk read and write performance.

ggtakec:
@gaul Thanks for the report.
I will concentrate and check it this week.

gaul:
The benchmark script may have an issue.  I observe a 25% improvement when running the same test manually:

without 46d79c5bc275c1df8cdd4e2ca0884efc68247e77:
```
$ time ls mnt/dir/ | wc -l
1000

real    0m8.109s
user    0m0.001s
sys     0m0.003s
```

with 46d79c5bc275c1df8cdd4e2ca0884efc68247e77:
```
$ time ls mnt/dir/ | wc -l
1000

real    0m5.715s
user    0m0.007s
sys     0m0.005s
```

gaul:
It appears 46d79c5bc275c1df8cdd4e2ca0884efc68247e77 increases CPU usage partially from `malloc_trim`.  Still investigating...

gaul:
I suspect that s3fs is not reusing HTTPS connections and the CPU overhead comes from the initial handshake.

ggtakec:
@gaul I confirmed that the session was not reused and HTTPS(TLS) handshake was running.
I will immediately check the contents of the code.

[...]

PR #937, opened 2019-01-29
Disable malloc_trim

### LINKED ISSUE #936: Consider removing malloc_trim calls
s3fs issues `malloc_trim` after every FUSE operation and XML parsing.  This has a lot of overhead with many threads they all call `malloc_trim` and each examines the entire heap, an O(n**2) operation.  42b74c9d2ed67da4cf1f6586d179b1d3c34fcb10 introduced these calls.  @ggtakec could you share more background on this since the change was intermingled with others?  References #935.

gaul:
This avoids walking the entire heap multiple times for complex
operations like `readdir`.  This does not entirely eliminate the
observed performance regression but does dramatically reduce s3fs CPU
usage.  References #935.  Fixes #936.

### DIFF
--- a/src/s3fs.h
+++ b/src/s3fs.h
@@ -51,7 +51,8 @@ static const int64_t FIVE_GB = 5LL * 1024LL * 1024LL * 1024LL;
 #include <malloc.h>
 
 #define DISPWARN_MALLOCTRIM(str)
-#define S3FS_MALLOCTRIM(pad)          malloc_trim(pad)
+// malloc_trim disabled due to performance overhead with many threads/heaps
+#define S3FS_MALLOCTRIM(pad)
 #define S3FS_XMLFREEDOC(doc) \
         { \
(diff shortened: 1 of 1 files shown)


ggtakec:
@gaul 
How about making https://github.com/s3fs-fuse/s3fs-fuse/blob/master/src/s3fs.h#L54 line a null value like https://github.com/s3fs-fuse/s3fs-fuse/blob/master/src/s3fs.h#L80 ?
I think you can delete the DISPWARN_MALLOCTRIM macro.
However, I would like to keep S3FS_MALLOCTRIM, S3FS_XMLxxxxxx macros for investigation using valgrind.
It's a suggestion to change a little code and change the condition of HAVE_MALLOC_TRIM.

gaul:
>  I think you can delete the DISPWARN_MALLOCTRIM macro.

Done.  Thanks for sharing this background!

PR #944, opened 2019-01-31
Repair utility mode

gaul:
This deinitialized `S3fsCurl` twice and incorrectly calculated V4
signatures.


ggtakec:
Thanks for fixing bugs

Issue #945, opened 2019-01-31
Make utility mode non-interactive


gaul:
Utility mode allows listing and aborting incomplete MPU via user input:

```
$ s3fs mybucket -u
Lists the parts that have been uploaded for a specific multipart upload.

---------------------------------------------------------------
 Path     : /128gb
 UploadId : JzQEgxwURH8h9XU9Cpa9uNFpTFhjICW7e5cBW7RmmgW9v_1TlVGqlU2F27g6r21DW9Lw97JzsPxN.bsh.pwN3fmyQNnZlQ_u006LD5ds0uQ-
 Date     : 2019-01-31T04:25:44.000Z

---------------------------------------------------------------
Would you remove all objects? [Y/N]
```

This only allows interactive uses and thus cannot be part of our CI tests.  Instead could s3fs break the operations into two, e.g., `--list-incomplete-mpu` and `--abort-incomplete-mpu`, perhaps with the latter taking a `older-than-day` parameter?  Alternatively, s3fs could automatically abort incomplete MPU older than some number of days like goofys does.

ggtakec:
@gaul Thanks for your suggestion.

I think that adding "--list-incomplete-mpu" option is good.
About the "--abort-incomplete-mpu" option, I think this option should have parameters of "XX[YMDhm]" or "all" as its arguments.

The "-u" option can not be deleted because it is backward compatible.
But what about assigning "--incomplete-mpu" as an equivalent option?
Along with that, all option name prefix has "--incomplete-mpu", how about as follows?
```
--incomplete-mpu(-u)
--incomplete-mpu-list
--incomplete-mpu-abort ([1-9][0-9][Y|M|D|h|m] | all)
```
_"YMDhm" is the year/month/day/hour, and "all" is all._

ggtakec:
@gaul This issue is fixed by #951 , then this will be closed.
If you have more problem about this, please reopen.

PR #951, opened 2019-02-03
Added a non-interactive option to utility mode

ggtakec:
#945 

### Details
Added non-interactive mode to Utility mode.
The following options are added.
```
--incomplete-mpu(-u)
--incomplete-mpu-list
--incomplete-mpu-abort ([1-9][0-9][Y|M|D|h|m] | all)
```
@gaul please check this PR.


src/s3fs_util.cpp
```diff
@@ -1349,6 +1349,17 @@ void show_help ()
     "   There are many FUSE specific mount options that can be specified.\n"
     "   e.g. allow_other  See the FUSE's README for the full set.\n"
     "\n"
+    "Utility mode Options:\n"
+    "\n"
+    " -u, --incomplete-mpu   List incomplete multipart upload incomplete\n"
+    "                        objects and delete them all interactively.\n"
+    " --incomplete-mpu-list  List incomplete multipart upload incomplete\n"
+    "                        objects.\n"
+    " --incomplete-mpu-abort=number[Y|M|D|h|m]\n"
```
gaul: Perhaps a simpler interface would be to just provide a day offset, e.g., abort any uploads older than 7 days?  It seems unlikely that users would want to abort downloads with smaller granularity.

src/s3fs.cpp
```diff
@@ -3523,26 +3533,56 @@ static bool abort_uncomp_mp_list(uncomp_mp_list_t& list)
...
-    if(NULL != fgets(buff, sizeof(buff), stdin)){
-      if(0 == strcasecmp(buff, "Y\n") || 0 == strcasecmp(buff, "YES\n")){
-        break;
-      }else if(0 == strcasecmp(buff, "N\n") || 0 == strcasecmp(buff, "NO\n")){
-        return true;
+  bool   is_all;
+  time_t expire_time = 0;
+  time_t now_time    = time(NULL);
+
+  if(is_cli){
```
gaul: Maybe we should mark this as deprecated so we can remove it in the future?  Although given the breakage fixed in #944 it is likely not many people use interactive mode today.

src/string_util.cpp
```diff
@@ -283,6 +284,59 @@ string get_date_iso8601(time_t tm)
...
+        unixtime *= (60 * 60);
+      }else if('m' == *ptmp){
+        unixtime *= 60;
+      }else{
+        return false;
+      }
+      if('\0' != *(ptmp + sizeof(char))){
+        return false;
+      }
+      break;
```
gaul: This ignores more complex strings, e.g., 12h30m, or any junk data, 123jkl.  Maybe it should error instead?

src/s3fs_util.cpp
ggtakec: I forgot to write the "all" parameter.
In addition to specifying detailed parameters, there is an "all" parameter.
Let's use this "all" as the default value.

src/s3fs.cpp
ggtakec: I see. Certainly the user seems to suffer.
Currently, let's deprecate the function which operates as CLI mode.
Leave the options themselves and let them do the same thing as --incomplete-mpu-list .

src/string_util.cpp
ggtakec: I was wondering if only simple parameters are enough, but considering the possibility of future corrections, I will try to make it complicated.

ggtakec:
@gaul Thanks for your check.
And I'll change those code, please wait a moment.

I am wondering if incomplete mpu data should be deleted at regular intervals in the main process of s3fs.
It might be better to wait for a subprocess or another process, it looks like a daemon.

gaul:
I don't think s3fs should provide the maximum flexibility since other common tools like AWS CLI do this already:

```
aws s3api list-multipart-uploads --bucket BUCKET
aws s3api abort-multipart-upload --bucket BUCKET --key KEY --upload-id ID
```

Instead just removing all uploads older than N days, perhaps automatically, should cover most use cases.

ggtakec:
@gaul I try modifying the code to delete incomplete uploads that have passed N days during startup, and review including options.

ggtakec:
I made the final code correction.
The "-u" option has the same definition as "--incomplete-mpu-list". Thus the interactive mode is not supported.
The "--incomplete-mpu-abort" option can specify the "time format(YMDhms)" and "all" parameters, and those parameters are optionally.
In the case of no parameters, it defaults to 24 hours. This is the intention that there is almost no problem if it is 24 hours.
Automatic deletion of incompalete multipart upload parts is considered separately. (I anticipate that s3fs will not be able to delete them immediately after failing)

PR #952, opened 2019-02-03
Automatically abort failed multipart requests

gaul:
This can avoid dangling parts.  However, many transfers fail due to
network errors so we still need other mechanisms to handle these
parts.


ggtakec:
@gaul 
> we still need other mechanisms to handle these parts.
I think so.

Issue #45, opened 2014-06-28
incron / lsyncd (inotify or fsevents) not detecting changes at s3.


paulm17:
Hi,

I've been pulling my hair out all day on finding out why an s3fs mounted directory won't be updated via incron or lsyncd.

Then I happened to come across this: 

```
http://unix.stackexchange.com/questions/100837/incrontab-doesnt-detect-modifications-on-a-s3fs-mount
```

I am running the latest fuse: 

```
fuse-devel-2.9.3-1.el6.x86_64
fuse-libs-2.9.3-1.el6.x86_64
fuse-2.9.3-1.el6.x86_64
```

I created the rpms myself for centos 6.5.

and the latest s3fs: 

```
[root@ip-10-0-4-78 www]# s3fs --v
Amazon Simple Storage Service File System V1.77 with OpenSSL
```

Yet, when the s3 bucket is updated, the change is available when i view the s3fs mount point but incron or lsyncd do not update nor to they do suggest any activity in their logs.

However, if I go into the s3fs mount and I do an update there.  Everything works as it should.

So why am I having the issue of nothing being detected when a change at s3 happens?

Thanks

micgrivas:
inotify cannot work on s3fs, because s3fs works over HTTPS. Practically, it is the same as DAV. 
The s3fs makes a pseudo-filesystem, for our ease, but cannot be 100% a file system.
Each time one copies files to s3 (over a mount with s3fs) , then the s3fs creates a local file and then run an HTTP request to upload it.  
There is no way for such a system to understand changes in s3-subsystem. Like an apache server will not automatically find any changes in the /var/lib/www directory and refresh any page.

PR #958, opened 2019-02-13
Document lack of inotify support

gaul:
References #45.


gaul:
> It's not really **no** inotify support, though, is it? I would suggest something more like...

reworded

PR #960, opened 2019-02-19
Wtf8

kristjanvalur:
### Details
S3 errors if object names are not utf-8.  File system encodings are not required to be utf8, and in fact windows happily uses cp1252 to name files in NFS exports. 

This change optionally examines the input path names and turns them into valid utf-8 code by using a private area of the Unicode namespace to store invalid utf8 bytes.  The data are then converted back for listdir operations and readlink.

This allows encodings other than utf-8 for filenames.


src/s3fs_util.cpp
```diff
@@ -1340,6 +1340,17 @@ void show_help ()
...
     "\n"
+    "   use_wtf8 - support arbitrary file system encoding.\n"
+    "        S3 requires all object names to be valid utf-8. But some\n"
+    "        clients, notably Windows NFS clients, use their own encoding.\n"
+    "        This option re-encodes invalid utf-8 object names into valid\n"
+    "        utf-8 by mapping offending codes into a 'private' codepage of the\n"
+    "        Unicode set.\n"
+    "        Useful on clients not using utf-8 as their file system encoding.\n"
+    "\n"
+    "   notsup_compat_dir (not support compatibility directory types)\n"
```
ggtakec: Please remove L:1351 to L:1353 about "notsup_compat_dir", because it is already explained above this line.

kristjanvalur: Right, I  left it in by mistake.  Fixed.

ggtakec: Thanks.

PR #961, opened 2019-02-20
Work around cppcheck warnings

gaul:
Follows on to eeb839242ba826707205b0b776c042203226d01b.


ggtakec:
@gaul Thanks for following.

PR #965, opened 2019-02-25
Improvement of curl session pool for multipart

ggtakec:
#935 

### Details
Improvement of renegotiation of SSL sessions in s3fs multipart request.

#### Current
s3fs is pooling curl handles inside it.
This pool behaves like a LIFO.
In the multipart request of HEAD, PUT, POST, and GET, SSL re-negotiation has occurred even though s3fs uses the curl handle of pool.
The session has been disconnected from the S3 server in about 5-10 seconds.

#### About HEAD request
In the case of a HEAD request, s3fs may send a request with a 400/404 error in order to confirm the existence of the file.
This normally CURL closes the session.
The solution is to set the CURLOPT_KEEP_SENDING_ON_ERROR option to the CURL handle.
Also, since the S3 server does not use ALPN, you should also use the CURLOPT_SSL_ENABLE_ALPN option.

However, the curl version of OS on TravisCI used by s3fs is low, and these options are not available.
In this PR, these options were commented out and could not be used.

Thus, if s3fs receives a 400/404 error after a HEAD request, renegotiation will occur.
(If it is a directory or file created in s3fs, 400/404 does not occur, so we left it as it is.)

#### About Other multipart request
Currently, in the flow of s3fs multipart request (POST etc.), first create a CURL wrapped object to be executed in parallel and then use it to communicate.
s3fs allocates a CURL handle when creating these objects.
For this reason, when actually communicating (getting a semaphore), the session has already been disconnected from the S3 server.
Therefore, renegotiation occurred.

With this PR, CURL handles are not allocated when creating CURL wrapped objects. It allocates it just before communication.
CURL handles are pooled inside s3fs and they can be retrieved as LIFOs.
Since it is a LIFO, sessions will never be lost if you use the handle that you used before.

This time, I also changed the pool of CURL handles.
When the pool becomes full, s3fs discards the oldest handle.

#### Result
SSL renegotiation does not occur with uploading muiltiart.
(Only the first time occurred)

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -238,7 +238,6 @@ const char* BodyData::str() const
 //-------------------------------------------------------------------
 // Class CurlHandlerPool
 //-------------------------------------------------------------------
-
 bool CurlHandlerPool::Init()
 {
   if (0 != pthread_mutex_init(&mLock, NULL)) {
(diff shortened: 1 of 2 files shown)


gaul:
I measured more than 5x improvement when listing 5,000 files:

1.84 - 06032aa661f8cfd06b997147efb47f6e0f0bbb48
0m23.023s
0m20.165s
0m21.115s
0m22.282s
0m21.369s

master - 0d43d070ccf55a8745d8fe95ec9558e29a85cbb2
0m3.848s
0m3.731s
0m3.502s
0m4.095s
0m3.779s

Issue #976, opened 2019-03-10
Preparation for release of version 1.85


ggtakec:
## Additional Information
Multiple defects have been fixed.
And upload, readdir performance is also improved.

## Details about issue
We will prepare for the release of the new version 1.85.

## Appreciation
Thanks to your cooperation, I was able to make it a better application.  
I greatly appreciate their cooperation that gave me the PR listed below.  
@Basavaraju013  
@DreamFlasher  
@earlchew  
@gaul  
@JoohnSF  
@juliogonzalez  
@kristjanvalur  
@kzidane  
@LutzFinsterle2019  
@mapreri  
@mcgitty  
@orozery  
@soulprovidr

ggtakec:
### Launched new release 1.85 now.

#### Appreciation
Thanks @Basavaraju013 @DreamFlasher @earlchew @gaul @JoohnSF @juliogonzalez @kristjanvalur @kzidane @LutzFinsterle2019 @mapreri @mcgitty @orozery @soulprovidr @sqlbot .
This version has been able to fix bugs, improve performance, and other enhancements.
Your help was much appreciated.
I could not have done this new release without you.

#### New owner member
And @gaul has joined s3fs-fuse owner member from this version release.
We will continue to cooperate and improve s3fs-fuse.

### Please packaging
@mapreri @jollyroger
Just let you know, new release 1.85 version is tagged now.
Please check and start to package it if it have no problem.

Thanks in advance for your assistance.

ggtakec:
I found a bug report #979 .
This version does not contain any fixes and will be fixed in the next update.

juliogonzalez:
Did we break compatibility with RES6/CentOS6?

While preparing the SPEC for v1.85 (https://github.com/juliogonzalez/s3fs-fuse-rpm/pull/11) I found that s3fs-fuse does not compile anymore for CentOS6: https://jenkins.juliogonzalez.es/job/s3fs-fuse-rpm-build-pr/41/DISTRO=centos6,label=docker/console
```
23:53:31 curl.cpp: In member function 'bool S3fsCurl::ResetHandle()':
23:53:31 
curl.cpp:1851: error: 'CURLOPT_TCP_KEEPALIVE' was not declared in this scope
23:53:32 make[2]: *** [curl.o] Error 1
23:53:32 make[1]: *** [all-recursive] Error 1
23:53:32 make: *** [all] Error 2
```

juliogonzalez:
[...]

PR #987, opened 2019-03-22
Support undefined CURLoption in libcurl library used in build

ggtakec:
#976 #935 

## Details
Maintaining an SSL session with CURL will improve performance, and s3fs should use the CURLOPT_TCP_KEEPALIVE, CURLOPT_KEEP_SENDING_ON_ERROR and CURLOPT_SSL_ENABLE_ALPN CURLoptions for curl_easy_setopt().

However, the version of libcurl you use when building may not support these options.

This PR can create s3fs regardless of the version of libcurl used in the build.
That is, if these options are not defined in curl.h, these options will be defined in the s3fs header file and used.
Certainly, defining option values ​​in curl.h to be present in s3fs may not be good, but we expect that libcurl will not change the defined values ​​and assign a fixed value.
So far, this PR allows you to build s3fs without reliance on the libcurl version.

#### configure output
When you start configure, the following message is output.
It will output "no" if the target option is undefined.
```
checking checking CURLOPT_TCP_KEEPALIVE... (yes or no)
checking checking CURLOPT_SSL_ENABLE_ALPN... (yes or no)
checking checking CURLOPT_KEEP_SENDING_ON_ERROR... (yes or no)
```
#### common.h
In this header file, CURLOPT_TCP_KEEPALIVE, CURLOPT_KEEP_SENDING_ON_ERROR and CURLOPT_SSL_ENABLE_ALPN CURLoptions are defined with S3FS prefix.
And s3fs calls curl_easy_setopt () using the S3FS prefix symbol.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -265,6 +265,51 @@ AC_COMPILE_IFELSE(
 )
 
 dnl ----------------------------------------------
+dnl check CURLoption
+dnl ----------------------------------------------
+dnl CURLOPT_TCP_KEEPALIVE (is supported by 7.25.0 and later)
+AC_MSG_CHECKING([checking CURLOPT_TCP_KEEPALIVE])
+AC_COMPILE_IFELSE(
+  [AC_LANG_PROGRAM([[#include <curl/curl.h>]],
+                   [[CURLoption opt = CURLOPT_TCP_KEEPALIVE;]])
+  ],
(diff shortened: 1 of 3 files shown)


ggtakec:
@gaul I will merge this PR. If you find a problem please let me know.

PR #989, opened 2019-03-23
Added a nonempty option to man page

dstori:
### Details
Missing nonempty option in the man page

### DIFF
--- a/doc/man/s3fs.1
+++ b/doc/man/s3fs.1
@@ -302,6 +302,9 @@ When s3fs catch the signal SIGUSR2, the debug level is bumpup.
 .TP
 \fB\-o\fR curldbg - put curl debug message
 Put the debug message from libcurl when this option is specified.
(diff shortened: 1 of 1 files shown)


gaul:
Should `s3fs --help` include `nonempty` as well?

ggtakec:
@dstori @gaul 
The `nonempty` option is an option provided and processed by fuse itself.
And s3fs does nothing about this option.
The fuse man page describes this option as follows:
```
nonempty
              Allows mounts over a non-empty file or directory. By default these  mounts are rejected to prevent accidental covering up of data, which could for example prevent automatic backup.
```
I think we should not write in the man page and help, because s3fs does not handle this option.
If necessary, it is better to create a "SEE ALSO" section on the man page and write fuse (8).

dstori:
Closing based on @ggtakec suggestion.

PR #995, opened 2019-04-02
Work with filenames that include spaces

jaygooby:
### Details
• Simplified the stat generation line (single exec, using only stat)
• Quote variables so that the cache directory can also include spaces
• while/read loop to handle spaces in cached files
  (IFS was causing problems when all the files were saved into a single variable)
• Replace erroneous exit -1 with exit 1
• Prints current cache size if below supplied maximum

### DIFF
--- a/test/sample_delcache.sh
+++ b/test/sample_delcache.sh
@@ -34,33 +34,28 @@ if [ "X$1" = "X-h" -o "X$1" = "X-H" ]; then
 fi
 if [ "X$1" = "X" -o "X$2" = "X" -o "X$3" = "X" ]; then
   func_usage $PRGNAME
-  exit -1
+  exit 1
 fi
 
 BUCKET=$1
-CDIR=$2
+CDIR="$2"
 LIMIT=$3
 SILENT=0
 if [ "X$4" = "X-silent" ]; then
   SILENT=1
 fi
-FILES_CDIR=$CDIR/$BUCKET
-STATS_CDIR=$CDIR/\.$BUCKET\.stat
-
+FILES_CDIR="${CDIR}/${BUCKET}"
+STATS_CDIR="${CDIR}/.${BUCKET}.stat"
+CURRENT_CACHE_SIZE=`du -sb "$FILES_CDIR" | awk '{print $1}'`
 #
 # Check total size
 #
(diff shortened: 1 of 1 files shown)


test/sample_delcache.sh
```diff
@@ -34,60 +34,59 @@ if [ "X$1" = "X-h" -o "X$1" = "X-H" ]; then
...
-  TMP_ATIME=`echo $part | sed 's/\:/ /' | awk '{print $1}'`
-  TMP_STATS=`echo $part | sed 's/\:/ /' | awk '{print $2}'`
-  TMP_CFILE=`echo $TMP_STATS | sed s/\.$BUCKET\.stat/$BUCKET/`
-
-  if [ `stat -c %X $TMP_STATS` -eq $TMP_ATIME ]; then
-    rm -f $TMP_STATS $TMP_CFILE > /dev/null 2>&1
+#
+# Make file list by sorted access time
+#
+find "$STATS_CDIR" -type f -exec stat -c "%X:%n" "{}" \; | sort | while read part
```
gaul: Could you simplify this logic with:

```
find -printf '%A@\t%p\n' | sort -n
```

jaygooby: Kinda. You could use `-printf '%A@:%p\n'` (you need the `:` as a separator for the `cut` command that comes later), but this `printf` gives you fractional seconds:

```
1554192949.6842425020:/mnt/.s3-cache/.my-bucket.stat/some/file with a space
```

but the existing comparison code expects to be fed plain-old integer epoch seconds. If you could get the `printf` to drop the fractions, you'd be good.

gaul:
@jaygooby I wonder if [shellcheck](https://www.shellcheck.net/) would help prevent these quoting issues?  Could you investigate?

jaygooby:
I've not seen shellcheck.net before, but it looks good for core shell linting. It's picked up e.g. the use of legacy backticks `` ` `` instead of `$()`. And when I tested it with the code prior to my PR, it also flagged the erroneous `-1` values to `exit`.

It didn't flag the key problem (gathering all the file paths in a variable), but it's not an issue that quoting would have helped with (which is why I refactored that portion of the code). 

So in summary, yes, it looks good, but it's not a silver bullet 😄

PR #997, opened 2019-04-05
Prefer HTTPS links in documentation where possible

gaul:
This also corrects a test URL which was HTTPS in practice.


ggtakec:
@gaul Thanks.

PR #1003, opened 2019-04-08
Standardize installation section

gaul:
Sort Linux distributions and simplify/standardize grammar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,37 +25,39 @@ Installation
 
 Many systems provide pre-built packages:
 
-* On Debian 9 and Ubuntu 16.04 or newer:
-
-  ```
-  sudo apt-get install s3fs
-  ```
-
-* On SUSE 12 or newer and openSUSE 42.1 or newer:
-
-  ```
-  sudo zypper in s3fs
-  ```
-
-* On Fedora 27 and newer:
-  ```
-  sudo yum install s3fs-fuse
-  ```
-
-* On RHEL/CentOS 7 and newer through EPEL repositories:
-  ```
-  sudo yum install epel-release
-  sudo yum install s3fs-fuse
-  ```
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,37 +25,39 @@ Installation
...
-* On SUSE 12 or newer and openSUSE 42.1 or newer:
+* Debian 9 and Ubuntu 16.04 or newer:
 
   ```
-  sudo zypper in s3fs
+  sudo apt-get install s3fs
   ```
 
-* On Fedora 27 and newer:
+* Fedora 27 or newer:
```
juliogonzalez: Just one note: [Fedora 27 is EoL](https://fedoraproject.org/wiki/End_of_life), and while v1.84 is there (as I submitted before EoL date), v1.85 was not submitted (package submissions are not allowed after EoL).

Same happens with openSUSE ([42.1 and 42.3 are EoL](https://en.opensuse.org/Lifetime) as well)

Not sure how we should address this problem, but if we don't, I expect some people can complain about recent versions not being available at EOL'ed distributions.

gaul: I think it is fine to mention the first version that packaged s3fs.  I dislike people using old versions since we sometimes fix data loss issues but I don't think it really hurts anything.

ggtakec: If necessary, should we add a proviso about fedora 27 etc?

gaul: We should not worry about documenting which distros package earlier versions.  The distros themselves warn users to upgrade as they have thousands of packages which have bugs and lack security fixes.  We can either mention the first distro which has support or remove the version number altogether.

ggtakec:
@juliogonzalez @gaul 
I will try to merge this PR.
If you still need to discuss, let's create issues while watching the progress.

Issue #929, opened 2019-01-25
man page and --help are inconsistent


gaul:
We should add some mechanism to keep these in sync if we cannot have a single source of truth.

ggtakec:
@gaul Thanks

gaul:
Can you repoen this issue?  I suspect we have other inconsistencies and need some kind of script to detect them.

ggtakec:
@gaul OK, reopen this.
(We have to check the man page and help options, and WIKI.)

gaul:
I believe that all options are now documented except for `createbucket`, `host`, and `servicepath`.  The first is hidden for only tests while the latter I am not sure what they do.

PR #1011, opened 2019-04-14
Make man page and --help more consistent

gaul:
Mostly small fixes but also some reordering.  References #929.  Found
via:

```
diff -u <(man --no-hyphenation doc/man/s3fs.1 | tr -s ' ' '\n' | sed '/^-o$/d' ) <(src/s3fs --help | tr -s ' ' '\n' | sed '/^-$/d')
```


ggtakec:
@gaul Thank you for the proofreading etc.

PR #1015, opened 2019-04-18
Remove uses of std::map::at

gaul:
This provides compatibility with C++03 and removes duplicate lookups.


gaul:
Tracking why `gcc -std=c++03` does not flag this here:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90135

ggtakec:
@gaul Thanks.

PR #1021, opened 2019-04-25
fix issue with aws session token provided inside ~/.aws/credentials file

michallula:
### Details
I fixed one issue that I discovered recently related to AWS session token support: when credentials are provided by ~/.aws/credentials file and session token is present, session token is not to request. I decided to remove set_is_session_token_use function from public interface of curl.cpp file. Instead of this I set the flag to true inside method SetAccessKeyWithSessionToken.

I hope code is cleaner right now and works fine. Sorry for introducing a bug in last PR.


ggtakec:
@michallula Thanks for your kindness.

Issue #607, opened 2017-05-24
Improve random writes and appends


CAFxX:
In the README.md the performance of random writes and appends is listed as a limitation:

> # Limitations
> Generally S3 cannot offer the same performance or semantics as a local file system. More specifically:
> - random writes or appends to files require rewriting the entire file

If this is still the case, was it ever considered to use multipart upload, and specifically both [Upload Part](https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPart.html) and [Upload Part - Copy](https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPartCopy.html), to speed it up? 

Specifically, an append operation would become (simplifying):
- initiate multipart upload
- upload part - copy (old object)
- upload part (data to append)
- complete multipart upload
- delete old object

A random write operation (e.g. "helloworld" -> "hello World") would become:
- initiate multipart upload
- upload part - copy (old object, off 0 len 5 "hello")
- upload part (changed data, " W")
- upload part - copy (old object, off 6 len 4 "orld")
- complete multipart upload
- delete old object

This approach would probably be beneficial when the existing object is big and the amount of data to append or changed is small, so some heuristics may be needed. In addition, these heuristics should also consider that (as noted below by @sqlbot) the upload part API has some restrictions in terms of minimum/maximum allowable part size ("Part size: 5 MB to 5 GB, last part can be < 5 MB")

sqlbot:
S3 itself would not allow such an operation at small scale.  The Multipart Upload API, whether the part data is uploaded or copied from another object, requires that each part (except the last part) have a minimum size of 5 MB.

http://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html

CAFxX:
@sqlbot so the appending scenario would work without any problem (assuming the old object is at least 5mb in size). 

For the random write it could be still useful in certain limited scenarios (e.g. VM images) where it would still provide substantial benefits.

ggtakec:
@CAFxX @sqlbot I'm sorry for very late reply.
I implemented this feature in #1027 .
If you can, please try to use master branch latest code after merging it.

ggtakec:
This issue is closed because #1098 has been merged.

PR #1027, opened 2019-05-02
Fixed to use copy api in multipart upload

ggtakec:
#607 

### Details
Modified to use `Upload part` and `Copy part(api)` when multipart upload of object.

This will allow you to download and upload the smallest sized parts if you upload after writing a part of the object.
(In accordance with the minimum part size 5MB of multipart upload, useless download and upload will not be performed.)

As much as possible, s3fs will improve performance by uploading only the updated area.

The `nomixupload` option has been added to keep the previous behavior backward compatible.
When this option is specified, s3fs operates as before.
Also, when the nocopyapi, norenameapi and nomultipart options are specified, the same operation as before is performed.

Besides this, test cases for this modified are added. (and the two typo corrections.)


gaul:
Sorry for my delayed response; I will look at this soon.

test/integration-test-main.sh
```diff
@@ -352,6 +352,30 @@ function test_multipart_copy {
...
+
+    if [ `uname` = "Darwin" ]; then
+       cat /dev/null > $BIG_FILE
+    fi
+    dd if=/dev/urandom of="/tmp/${BIG_FILE}" bs=$BIG_FILE_LENGTH seek=0 count=1
+    dd if="/tmp/${BIG_FILE}" of="${BIG_FILE}" bs=$BIG_FILE_LENGTH seek=0 count=1
+
+    # modify directly(seek 7.5MB offset)
+    echo -n "0123456789ABCDEF" | dd of="${BIG_FILE}" bs=1 seek=7864320 conv=notrunc
+    echo -n "0123456789ABCDEF" | dd of="/tmp/${BIG_FILE}" bs=1 seek=7864320 conv=notrunc
```
gaul: This issues 16 1-byte writes -- do you intend this?

src/fdcache.cpp
```diff
@@ -494,6 +502,117 @@ int PageList::GetUnloadedPages(fdpage_list_t& unloaded_list, off_t start, size_t
...
+//
+bool PageList::GetLoadPageListForMultipartUpload(fdpage_list_t& dlpages)
+{
+  // compress before this processing
+  if(!Compress()){
+    return false;
+  }
+
+  size_t    before_loaded_size = 0;
+  bool      before_indigent    = false;
```
gaul: What does indigent mean?

ggtakec:
@gaul I'm sorry about this, I will fix conflict and response your review comments.

src/fdcache.cpp
ggtakec: Sorry, it means "not enough of 5MB", I will change the flag name to "before_insufficient".

test/integration-test-main.sh
ggtakec: Yes, I intend to issue 16 1-byte writes.
Would you like to write one 16 byte or something else?

ggtakec:
@gaul 
> I did some further testing and found that the following multi-byte test fails:
> 
> ```shell
> dd if=/dev/zero of="${BIG_FILE}"      bs=4 count=1 seek=7864320 conv=notrunc
> dd if=/dev/zero of="/tmp/${BIG_FILE}" bs=4 count=1 seek=7864320 conv=notrunc
> ```
> 
> The logs suggest some bad input to S3:
> 
> ```
> s3fs: [ERR] curl.cpp:RequestPerform(2412): HTTP response code 400, returning EIO. Body Text: <?xml version='1.0' encoding='UTF-8'?><Error><Code>InvalidArgument</Code><Message>Bad Request</Message><RequestId>4442587FB7D0A2F9</RequestId></Error>
> ```
I was able to reproduce this bug.
The process of writing beyond the original file size seems wrong.

ggtakec:
@gaul This PR is closed once for rebasing and generated as a new PR.
(Because rebase modification is various and you want to change the branch of the merge source)
I will close PR, so please wait.

test/integration-test-main.sh
gaul: I think either is fine, but the test case below needs more than one byte to trigger a failure for some reason.

ggtakec:
Now that the bug has been fixed, I will avoid the conflict , rebase and create a new PR.
I will close this PR now, but I will set the anchor from the new PR.

PR #1049, opened 2019-06-22
Remove cache file when object time differs

### LINKED ISSUE #1047: With cache enabled, data corrupted after update to an object
### Additional Information

1.85

2.9.2

3.10.0-123.el7.x86_64

NAME="Red Hat Enterprise Linux Server"
VERSION="7.0 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="7.0"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.0 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.0:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.0
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION=7.0

/usr/bin/s3fs amol-test-s3fs /awsmount -o use_cache=/s3fscache -o dbglevel=debug -o curldbg

NA

Attached.
[mylog.txt](https://github.com/s3fs-fuse/s3fs-fuse/files/3313987/mylog.txt)

### Details about issue
I am observing that if cache is enabled, when an object is accessed through s3fs, and then updated by means external to s3fs, and later re-accessed through s3fs,  s3fs detects that the object has changed, but does not retrieve the entire object. It instead retrieves only the additional number of byes in the new object, retaining the earlier bytes from the old object.  This corrupts the retrieved object, by mixing content of the earlier and latest versions of the object.

Here are the steps to do so:

First remove the cache directory and re-create it.

    [root@<machine1>~]# rm -rf /s3fscache/
    [root@<machine1>~]# mkdir /s3fscache

Now mount with the just-created empty cache directory

    [root@<machine1>~]# /usr/bin/s3fs amol-test-s3fs /awsmount -o use_cache=/s3fscache -o dbglevel=debug -o curldbg

Now print the content of the object "tempaws.txt"

    [root@<machine1>~]#  cat /awsmount/tempaws.txt
    old data

Now change the data in the tempaws.txt externally (using AWS Management Console) to "new data added". I did this by updating my local file and re-uploading it using AWS Management Console.
Now access the file again

    [root@<machine1>~]#  cat /awsmount/tempaws.txt
    old data added

Note that the data should have been **"new data added"**; instead it is **"old data added"**.

You will notice that the data is a combination of old data ("old data ") and new data  ("added"). If you see the logs, you will see that it detected that the file has changed, but only requested bytes 8-13, assuming that the first 8 bytes have remained same. It then combined bytes 0-7 from old data and 8-13 from new data.

This behavior is only seen with caching enabled.

Is this expected behavior? If so, is there a workaround without turning the cache off altogether?

gaul:
Check the modification times to determine whether an object has
updated.  This relies on low clock skew between s3fs and the S3
server; a more robust approach could use the ETag.  Fixes #1047.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ matrix:
         - sudo apt-get update -qq
         - sudo apt-get install -qq attr cppcheck libfuse-dev openjdk-7-jdk
         - sudo update-alternatives --set java /usr/lib/jvm/java-7-openjdk-amd64/jre/bin/java
+        - sudo pip install --upgrade awscli
       script:
         - ./autogen.sh
         - ./configure CPPFLAGS='-I/usr/local/opt/openssl/include' CXXFLAGS='-std=c++03'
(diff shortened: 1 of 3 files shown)


gaul:
This seems to fail on Ubuntu 14.04 due to an awscli issue:

```
'NoneType' object has no attribute 'startswith'
```

Maybe we should upgrade to 16.04 since the former is end-of-life anyway?

gaul:
It should suffice to just pip upgrade awscli.

Issue #839, opened 2018-10-12
git clone premature end of pack file


quancore:
**Version of s3fs being used:** V1.84
**Kernel information :** 4.15.0-1023-aws
**GNU/Linux Distribution:**
![image](https://user-images.githubusercontent.com/15036825/46879261-d36cfa80-ce45-11e8-911d-547d1c7df098.png)
**/etc/fstab:**
`s3fs#csgo-ranking-prediction    /home/ubuntu/s3         fuse    _netdev,exec,allow_other,passwd_file=/etc/passwd-s3fs,cipher_suites=AESGCM,kernel_cache,max_background=1000,max_stat_cache_size=100000,multipart_size=52,parallel_count=30,multireq_max=30,umask=0002,uid=1000,gid=1000,dbglevel=warn   0        0`
**system logs:**
[log.txt](https://github.com/s3fs-fuse/s3fs-fuse/files/2473790/log.txt)

**error details:**
![image](https://user-images.githubusercontent.com/15036825/46880605-6a878180-ce49-11e8-9eb2-8e111d253b18.png)

cortesea:
Hi,
I have the same issue.
Have you succeeded in solving it?

quancore:
I have switched to goofys and decided to handle all kind of git operations on my local

ggtakec:
@quancore 
I can not reproduce this problem.
I looked for similar problems, and they seemed to be occurring when repository capacity is large, git command is old, etc.
We would like to confirm the following for this Issue solution.
Is this problem also occurring in other repositories?
Is the git command occurred even at the latest?

tkang007:
Hi,

Hi,
It still happened, please reference below.

Error message when clone to s3 mounted directory:

Cloning into '/public/reference/jupyter-tutorial'...
remote: Enumerating objects: 142, done.
remote: Total 142 (delta 0), reused 0 (delta 0), pack-reused 142
Receiving objects: 100% (142/142), 223.22 KiB | 38.00 KiB/s, done.
fatal: premature end of pack file, 63 bytes missing
fatal: index-pack failed

[root@ip-10-94-89-236 ]# cat /etc/fstab
#
UUID=e9c3e7c2-e039-4b2b-9991-955a8c7cd8c0     /           xfs    defaults,noatime  1   1
/dev/xvdb  swap  swap defaults 0 0
/dev/xvdc /dat1  ext4  defaults 0 2
[...]

Issue #1050, opened 2019-06-24
1.86 release


gaul:
master has a number of [changes since 1.85](https://github.com/s3fs-fuse/s3fs-fuse/compare/v1.85...master), some in-progress pull requests, and a few issues that I am investigating:

- [x] Time problem with large files #1019 (regression from 1.84)
- [x] file corruption with `use_cache` and external modifications #1047 
- [x] `git clone` premature end of pack file #839
- [x] The upload element blocking subsequent transfers #928

amolthorat1:
Hello,

Any update on the next release date for s3fs?

Thanks.

ggtakec:
@amolthorat1 I want to add # 1098 in addition to the contents suggested by @gaul.
Since some correspondence still remains now, please wait a little more.

gaul:
You do not need to wait for a release version; just `git clone` the repository to get the latest fixes.

amolthorat1:
We need to be able to point customer to a release version - getting the source code and building it may not work for them.

gaul:
s3fs will release when it is ready.  I suggest you fork the repo and tag a version for your customer.

gaul:
@ggtakec Fedora 31 and Ubuntu 19.10 both release in October and it would be nice if they could include the many bug fixes since 1.85.  Could we merge #1066 and #1107 and tag 1.86?  I worry that delaying the release for #1098 to mature poses too much risk.  I can handle the release process if this helps.

gaul:
@ggtakec I would like to release 1.86 in the hope that Ubuntu 20.04 LTS can include it.  Would you mind if I tagged the current master and wrote the release notes?  For the latter I would like to follow a simpler, user-focused format since GitHub already provides the commit list between versions.  I drafted this:

* add intelligent_ia storage tier
* add requester_pays support
* add session token support
* add symlink cache
* allow concurrent metadata queries during data operations
* allow large files on 32-bit systems like Raspberry Pi
* allow SSE-C keys to have NUL bytes
* enable various optimizations when using modern curl
* fix clock skew errors when writing large files
* fix data corruption when external modification changes a cached object
* fix data corruption when opening a second fd to an unflushed file
* fix multiple concurrency issues
* use server-side copy for partially modified files

ggtakec:
@gaul I suppose that your description is okay, but ChangeLog and ReleaseNotes should match the old ChangeLog format.
```
#<PR ID> <Title>
```
[...]

PR #1063, opened 2019-07-02
Flush file when opening second fd

gaul:
Previously when s3fs had dirty local data and an application opened a
second fd it would remove the stat cache entry, query S3 for the size
which was still zero, then reinitialize `FdEntry` with this incorrect
size.  Instead flush local data to S3 so this query works.  It is
possible that a more involved patch could do this with a less
heavyweight approach but this requires changing open.  This does not
completely fix `git clone` but allows it to proceed further.  Also make
some cache methods `const`-correct.  References #839.

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -220,7 +220,7 @@ void StatCache::Clear()
   S3FS_MALLOCTRIM(0);
 }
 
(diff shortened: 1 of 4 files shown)


tvoinarovskyi:
@gaul Do you plan to cut a release anytime with this change?) It would really help to have this fixed in a stable release.

gaul:
Please follow #1050.  We are trying to get a release soon.

PR #1066, opened 2019-07-02
Sign request immediately before sending

### LINKED ISSUE #1019: Time problem with large files
This is using the latest built version of s3fs.

While uploading a very large file, it eventually starts throwing the following error until retries run out

```
[ERR] curl.cpp:RequestPerform(2273): HTTP response code 403, returning EPERM. Body Text: <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>RequestTimeTooSkewed</Code><Message>The difference between the request time and the current time is too `large.</Message><RequestTime>20190425T030525Z</RequestTime><ServerTime>2019-04-25T03:20:29Z</ServerTime><MaxAllowedSkewMilliseconds>900000</MaxAllowedSkewMilliseconds><RequestId>6DF0BC81DF47A</RequestId><HostId>Bk0RsQE3gAMknO52Din1h13t0b57Vpphx6fTizHRlTspybFoIDCmvhbW+s6mEwz96EUybZcA=</HostId></Error>`

```
File size is 342,798,923,129 bytes and the error at around [start=342360064000][size=438859129][part=654]

For reference:

```
# ntpdate us.pool.ntp.org
24 Apr 20:23:53 ntpdate[4203]: adjust time server 74.117.214.2 offset -0.002418 sec

```

gaul:
Previously s3fs could create a long list of pre-signed requests which
could take longer than the default S3 clock skew limit of 15 minutes.
This also changes SHA-256 computation from single- to multi-threaded
since this is now computed in the worker threads.  Regression from
88cd8feb053980c808d67771d63a84ca25f6db8a.  Fixes #1019.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -1582,7 +1582,6 @@ bool S3fsCurl::UploadMultipartPostSetCurlOpts(S3fsCurl* s3fscurl)
   curl_easy_setopt(s3fscurl->hCurl, CURLOPT_INFILESIZE_LARGE, static_cast<curl_off_t>(s3fscurl->partdata.size)); // Content-Length
   curl_easy_setopt(s3fscurl->hCurl, CURLOPT_READFUNCTION, UploadReadCallback);
   curl_easy_setopt(s3fscurl->hCurl, CURLOPT_READDATA, (void*)s3fscurl);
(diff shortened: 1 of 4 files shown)


gaul:
Somehow this breaks multipart uploads.  Investigating...

gaul:
@ggtakec Any feedback on this pull request?

PR #1067, opened 2019-07-02
Fix mismatched new[]/free

gaul:
Found via Valgrind.


src/curl.cpp
```diff
@@ -2450,7 +2450,7 @@ string S3fsCurl::CalcSignatureV2(const string& method, const string& strMD5, con
     free(md);
```
ggtakec: Here too, it must be modified to `delete[] md`.

gaul: Done.

ggtakec: Thanks!

PR #1075, opened 2019-07-04
Compile with FORTIFY_SOURCE

gaul:
This can catch some classes of buffer overflows.


ggtakec:
@gaul Thanks for this, it is a good fix.

PR #1078, opened 2019-07-05
Issue multipart when object size exceeds part size

### LINKED ISSUE #1058: Fragmentation parameters do not work when the file size is greater than 8MB and less than 16MB
### Additional Information
S3fs has a multipart_size option, but when the file size is larger than 8MB and less than 16MB, the multipart_size option is specified as 8MB but it does not work. still using normal PUT upload.

s3fs-fuse-1.83
s3fs-fuse-1.85
*) I only compiled these two versions.

fuse-2.9.2-10.el7.x86_64
fuse-libs-2.9.2-11.el7.x86_64

5.1.5-1.el7.elrepo.x86_64

CentOS Linux 7 (Core)

```
s3fs niki-bj /data/test -o url=http://s3.cn-north-1.amazonaws.com.cn -o passwd_file=/data/.passwd-s3fs -o dbglevel=info -o curldbg,use_path_request_style,allow_other -o retries=1 -o multipart_size=8 -o multireq_max=8 -o parallel_count=32
```
```
```
```
```

gaul:
Previously s3fs issued multipart uploads when the object size was
twice the part size.  Conjoining this with the part size was confusing
and s3fs should add a separate tunable for this if needed, similar to
`singlepart_copy_limit`.  Fixes #1058.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -1194,7 +1194,7 @@ int FdEntity::Load(off_t start, off_t size)
       off_t over_size = iter->bytes - need_load_size;
 
       // download
-      if(2 * S3fsCurl::GetMultipartSize() <= need_load_size && !nomultipart){ // default 20MB
+      if(S3fsCurl::GetMultipartSize() <= need_load_size && !nomultipart){
         // parallel request
         // Additional time is needed for large files
         time_t backup = 0;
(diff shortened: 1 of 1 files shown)


ggtakec:
This was a code that specified twice the size in the old days for efficiency.
This modification is correct because it is unnecessary now.
Thank you.

PR #1080, opened 2019-07-06
Add simple XML parsing wrapper

gaul:
Also simplify `check_region_error`.


ggtakec:
Thank you for remaking function for common processes.

PR #1081, opened 2019-07-06
Break FdEntity locks into data and metadata

### LINKED ISSUE #928: The upload element blocking subsequent transfers
### Additional Information

Amazon Simple Storage Service File System V1.82(commit:unknown) with GnuTLS(gcrypt)

_example: 2.9.4_

fuse/bionic,now 2.9.7-1ubuntu1 amd64 [installed]
libfuse2/bionic,now 2.9.7-1ubuntu1 amd64 [installed]

4.15.0-1021-aws

~$ cat /etc/os-release
NAME="Ubuntu"
VERSION="18.04.1 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.1 LTS"
VERSION_ID="18.04"

na

s3fs#mybucket/sterling/mybucketfuse _netdev,allow_other,iam_role=myawsrole,parallel_count=50,uid=1001,umask=0077,url=http
s://s3-eu-west-1.amazonaws.com 0 0

```
```
### Details about issue

We have s3fs hung off a couple of HA'd SFTP servers. The SFTP servers are used almost exclusively for uploads. Uploads typically go to different locations, thus no real issues with synchronization etc. SFTP servers do of course like to do an ls at the end of transfers etc, but there is very little actual fetching of files.

A number of file transfers will come in at the same time. The file sizes range from 1-40GB (but transfers are over a performant direct connect connection to AWS). When the transfers all kick off at the same time they transfer up to the SFTP server just fine, however when the initial transfer completes and the upload to S3 begins, the other transfers block and wait. Most SFTP clients handle this fine, but we have one IBM product that is a fussy customer that doesnt. Is there any way to stop the other uploads from blocking? Is there anything I can do with the parameters to further tune things. The ec2 instances are decent spec network focussed instances (upload to S3 is impressively fast).

Many thanks

Mat

gaul:
Previously long-running data operations like `RowFlush` would block
metadata operations like `GetStats` and thus user `readdir`.  Fixes #928.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -623,14 +623,16 @@ int FdEntity::FillFile(int fd, unsigned char byte, off_t size, off_t start)
 // FdEntity methods
 //------------------------------------------------
 FdEntity::FdEntity(const char* tpath, const char* cpath)
-        : is_lock_init(false), refcnt(0), path(SAFESTRPTR(tpath)), cachepath(SAFESTRPTR(cpath)), mirrorpath(""),
-          fd(-1), pfile(NULL), is_modify(false), size_orgmeta(0), upload_id(""), mp_start(0), mp_size(0)
+        : is_lock_init(false), refcnt(0), path(SAFESTRPTR(tpath)),
+          fd(-1), pfile(NULL), size_orgmeta(0), upload_id(""), mp_start(0), mp_size(0), is_modify(false),
+          cachepath(SAFESTRPTR(cpath)), mirrorpath("")
 {
   try{
     pthread_mutexattr_t attr;
     pthread_mutexattr_init(&attr);
     pthread_mutexattr_settype(&attr, S3FS_MUTEX_RECURSIVE);   // recursive mutex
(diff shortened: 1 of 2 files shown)


gaul:
This likely needs more scrutiny than usual; I made one pass over all the members but discovered some existing missing lock issues.

ggtakec:
If it uses two lock variables, deadlocks can occur.
Therefore, we need to carefully check the sequence (such as function call nesting) too.
Please give me some time, we should try to check the whole.

gaul:
Agree that this is tricky.  We must make sure that when both locks are needed, we acquire `fdent_lock` before `fdent_data_lock`.  Note that during writes s3fs should only hold the latter during transfers.  I opened #1082 with the missing locking I discovered and will follow up with a `PTHREAD_MUTEX_ERRORCHECK` commit which might help discover other locking problems.

ggtakec:
Thanks, and merge this PR.
We will continue to take care for caution regarding deadlocks.

PR #1085, opened 2019-07-10
Remove recursive locking

gaul:
Recursive locking is frowned upon and is incompatible with
`PTHREAD_MUTEX_ERRORCHECK`.  Also clean up `pthread_mutex_lock` error
checking.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ matrix:
         - sudo pip install --upgrade awscli
       script:
         - ./autogen.sh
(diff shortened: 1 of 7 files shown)


gaul:
This commit can detect some kinds of deadlocks.

ggtakec:
This PR is accepted.
However, instead of using #ifdef, it may be better to always use PTHREAD_MUTEX_ERRORCHECK or to make it optional.

PR #1087, opened 2019-07-12
Prefer modern C headers

gaul:
Found and fixed via clang-tidy.


ggtakec:
It seems that there is no compatibility problem with the current code, so we accept this PR.

Issue #1091, opened 2019-07-13
s3fs core due to fd_manager_lock 


liuyongqing:
#### Version of s3fs being used (s3fs --version)
s3fs 1.84, master branch source code compile

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
fuse 2.9.4

#### Kernel information (uname -r)
Linux 7e812c9e54b6 4.4.0-130-generic #156-Ubuntu SMP Thu Jun 14 08:53:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
ubuntu 16.04

### Details about issue
the coredump call backtrace is:
![image](https://user-images.githubusercontent.com/5205797/61167134-52759780-a56c-11e9-9c46-74060f341059.png)

the crash was triggered by:
1. thread A call FdManager::Open and create a FdEntity
2. thread B call FdManager::Open and get  thread A created FdEntity and trying to call ent->Open
3. thread A call FdManager::Close and deleted the FdEntity
4. thread B continue ent->Open execute and then crashed due to "double free or corruption (out)"

liuyongqing:
@ggtakec

ggtakec:
@liuyongqing Thank you for the report. Earlier, your PR was merged.
Please let us know if you notice any problems in the future.

PR #1092, opened 2019-07-13
fix coredump caused by fd_manager_lock locking

liuyongqing:
### issue link
https://github.com/s3fs-fuse/s3fs-fuse/issues/1091

### Details
FdManager::Open function should hold FdManager::fd_manager_lock until the function execute finished.

In the current master version of s3fs，FdManager::Open hold FdManager::fd_manager_lock in a pair of bracket, and call FdEntity::Open without FdManager::fd_manager_lock  lock, this will cause coredump in multithreading, details are as follows:
1. thread A call FdManager::Open and create a FdEntity
2. thread B call FdManager::Open and get the FdEntity and trying to call ent->Open without hold FdManager::fd_manager_lock  
3. thread A call FdManager::Close and delete the FdEntity
4. thread B continue function FdEntity::Open execute and crashed happened due to "double free or corruption (out)"

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -2113,59 +2113,58 @@ FdEntity* FdManager::Open(const char* path, headers_t* pmeta, off_t size, time_t
   }
   bool close = false;
   FdEntity* ent;
-  {
-    AutoLock auto_lock(&FdManager::fd_manager_lock);
 
-    // search in mapping by key(path)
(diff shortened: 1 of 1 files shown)


ggtakec:
@liuyongqing Thanks for this bug fix.
Your explanation is correct. Possible problems with thread A and B.
I merged this PR.

Issue #1093, opened 2019-07-14
Multipart upload not working at all


larsch:
#### Version of s3fs being used (s3fs --version)
```
Amazon Simple Storage Service File System V1.85(commit:unknown) with OpenSSL
```

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
```
% pacman -Qi fuse
Name            : fuse2
Version         : 2.9.9-1
```

#### s3fs command line used, if applicable
```
s3f  -o ensure_diskfree=1500 -o multipart_size=8 -o dbglevel=debug -o curldbg -f bucket /path/to/bucket -o passwd_file=...,url=...
```
#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
_if you execute s3fs with dbglevel, curldbg option, you can get detail debug messages_
```
[INF]       fdcache.cpp:RowFlush(1462): [tpath=][path=/bugfile][fd=8]
[INF]       curl.cpp:MultipartUploadRequest(3909): [upload_id=ODZk...jY2Mx][tpath=/bugfile][fd=8][offset=0][size=8519680]
[INF]       curl.cpp:UploadMultipartPostRequest(3667): [tpath=/bugfile][start=0][size=8519680][part=4]
[INF]       curl.cpp:UploadMultipartPostSetup(3607): [tpath=/bugfile][start=0][size=8519680][part=4]
[INF]       curl.cpp:prepare_url(4504): URL is https://anon/bugfile?partNumber=4&uploadId=ODZ...2Mx
[INF]       curl.cpp:prepare_url(4536): URL changed is https://anon/bugfile?partNumber=4&uploadId=OD...2Mx
[INF]       curl.cpp:insertV4Headers(2587): computing signature [PUT] [/bugfile] [partNumber=4&uploadId=ODZk...TRjY2Mx] [6532...999]
[INF]       curl.cpp:url_to_host(102): url is https://anon
[DBG] curl.cpp:RequestPerform(2212): connecting to URL 
[ERR] curl.cpp:RequestPerform(2376): ###curlCode: 43  msg: A libcurl function was given a bad argument
[ERR] curl.cpp:MultipartUploadRequest(3932): failed uploading part(-5)
[ERR] fdcache.cpp:RowFlush(1564): failed to multipart post(start=0, size=8519680) for file(8).
```
### Details about issue

[...]

PR #1094, opened 2019-07-14
Fix multipart upload not setting curl options

larsch:
Issue #1093


ggtakec:
@larsch Thanks for this PR. 
It is a bug and fixed by you.

PR #1097, opened 2019-07-15
Add Lock() func parameter in RowFlush()

ggtakec:
#1081 

### Details
The Load method is called from the RowFluch method.
Since the Load method call has already locked the fdent_data_lock mutex, specify the parameter addition (which differs from the default value).


ggtakec:
@gaul This is necessary for testing source code changes after this PR.
Therefore, I will merge first.

PR #1101, opened 2019-07-16
Avoid race when using thread-unsafe gmtime

gaul:
Found via ThreadSanitizer.


ggtakec:
@gaul Thanks, this will be merged soon.

PR #1103, opened 2019-07-17
Use consistent default for readwrite_timeout

gaul:
Previously curl threads mutated this shared state without a lock.
Found via ThreadSanitizer.


gaul:
It might be fine to leave the default value of 60 but remove the unneeded racy mutation.  @ggtakec what do you think?

ggtakec:
@gaul Certainly, as you say, changing the default to 120 avoids this useless configuration.
Setting the default value to 120 seconds has no impact on the whole.
I will merge this PR.

PR #1099, opened 2019-07-15
Add log output filter script at test failure

ggtakec:
### Relevant Issue (if applicable)
n/a

### Details
If the test fails, TravisCI outputs the contents of the test-suite.log file.
At the time of test execution, loglevel = info is set to s3fs, and the operation log of s3fs is recorded in test-suite.log.
Recently, if a test fails, the test execution log file will grow and TravisCI will not be able to view everything.
In this case, TravisCI will display "This log is too long to be displayed. Please reduce the verbosity of your build or download the raw log."

This PR filters the log output so that TravisCI displays the log.
Failed tests display all logs, successful tests do not display logs of s3fs information level.

PR #1106, opened 2019-07-22
Fixed filter-suite-log.sh for on osx

ggtakec:
#1099

### Details
It was corrected because it might not be possible to filter when there are not enough new lines in the message.
This converges the phenomenon that occurred in OSX.


ggtakec:
Since this PR is necessary to display errors on other Travis etc, I will merge it.

Issue #1062, opened 2019-06-30
investigate macOS integration test workarounds


gaul:
Several tests including `test_append_file` have a suspicious workaround:

```sh
    if [ `uname` = "Darwin" ]; then
       cat /dev/null > ${TEST_TEXT_FILE}
    fi
```

Without it s3fs returns stale metadata from the stat cache.  Comparing logs confuses me but it appears that the flush happens later than on Linux.  @ggtakec do you have any idea why this is necessary?

ggtakec:
@gaul It was a code added by #630 PR.
At this time, maybe I added it to pass OSX tests, but this may have hidden the bug.
If you cut this code now, will it not cause an error on OSX?
If errors occur, I will try to investigate.
If it does not occur, we should remove this code immediately.

gaul:
Removing these causes the tests to fail.  I looked into this but I don't understand the failure.

ggtakec:
I remembered this reason, I think this depends on the behavior of osxfuse.
I did not pass the test(test_external_modification) at hand now, then I investigated and remembered.

The test fails if the file is written and read immediately after startup s3fs. This is because s3fs_read() handler is not called by osxfuse.

You can do same issue:
```
$ echo "test1" > s3fs-integration-test/testfile
$ cat s3fs-integration-test/testfile            <--- something wrong(empty)
$ echo "test2" > s3fs-integration-test/testfile
$ cat s3fs-integration-test/testfile            <--- ok
  test2
```
This seems to be related to the handling of "struct fuse_file_info * fi" inside osxfuse, but the details are unknown.
Issues that seem to be related: https://github.com/osxfuse/osxfuse/issues/203

Now I read and write only once to avoid this(cat /dev/null > file).
However, this workaround is not good, so we need to find a way to escape this(or osxfuse issue).

ggtakec:
@gaul 
In the case of osxfuse, it seems that read system call is not called when it is cached (by OS or fuse).
I could not find the same phenomenon with libfuse for linux.

The explanation of fuse (osxfuse) is as follows.
```
if kernel_cache option is not specified and neither direct_io, 
data is still cached after the open(2), so a read(2) system call 
will not always initiate a read operation.
```
So I tried the `kernel_cache` and` direct_io` options and confirmed that the read system call was called.
Note that if user specify the `kernel_cache` option, when user change the file from the outside (by the aws command etc.), the change can not be detected.
[...]

PR #1119, opened 2019-08-06
Added direct_io option for test on osx

ggtakec:
#1062 

### Details
In the osxfuse, when updating or reading a file, the update of stats information may be delayed.
To avoid this, specify the direct_io option.
This removes the useless test code pointed out in #1062 .

When using the direct_io option, osxfuse 3.8.0 or later is required. (There is a bug of redirect append in the previous version)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ matrix:
     - os: osx
       osx_image: xcode9.2
       before_install:
-        - HOMEBREW_NO_AUTO_UPDATE=1 brew tap caskroom/cask
+        - brew tap caskroom/cask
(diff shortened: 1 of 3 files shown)


ggtakec:
@gaul 
In order to avoid osx errors occurring in other PRs and merges, this PR is applied first.

And when using osxfuse 3.8.0 or earlier, direct_io causes a problem with append redirect.
( https://github.com/osxfuse/osxfuse/issues/497 )
So that I update brew in osx tests, but the osx test time exceeds 12 minutes.
If you have an idea to reduce test time, please let me know.

PR #1124, opened 2019-08-07
Remove unnecessary test sleeping

gaul:


gaul:
Apparently `-o use_cache` needs these.

ggtakec:
@gaul 
In osxfuse, when stat cache is enabled, it seems to have to wait for stat cache to become invalid.
Therefore, this sleep was added to make it wait for more than 1 second.
Please note that there is a difference between osxfuse and libfuse (linux) and the behavior is different.
If there is a problem, please let me know, and I will check again.

PR #1125, opened 2019-08-11
Fixed simple copy with nocpoyapi and modify flag in fdcache

ggtakec:
#1098 

### Details
On OSX, when the nocopyapi option was specified, the simple cp command failed.
It was detected by adding a simple cp command with a test case.

Unlike writing to a file with dd, when using the cp command, a callback such as chmod is called.
During the write callback process, the chmod callback was called and the file size was not updated, causing a problem.
This problem was caused by overwriting the size when opening an existing fd with FdManager::Open().

In addition to this modification, this PR also deletes FdEntity::is_modify flag.
This is_modify flag which indicates cache modification, should not be  managed by the FdEntity class(object), but by each element of FdEntity::pagelist.
In other words, add a modify flag to fdpage instead and manage modify status for individual fdpages.
As a result, there is a performance degradation for checking the modify state, but I believe it is insignificant.

This additional fix is ​​the same as that included in the #1098 PR.
(#1098 PR will be rebased)

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -244,9 +244,16 @@ void PageList::FreeList(fdpage_list_t& list)
   list.clear();
 }
 
-PageList::PageList(off_t size, bool is_loaded)
+PageList::PageList(off_t size, bool is_loaded, bool is_modified)
 {
-  Init(size, is_loaded);
+  Init(size, is_loaded, is_modified);
+}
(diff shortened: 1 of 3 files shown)


ggtakec:
@gaul First, this PR is applied and #1098 is corrected.
Please point out any problems.

PR #1126, opened 2019-08-11
Fixed multipart copy and its retry handler

ggtakec:
#1102

### Details
The concurrent write test failed when multipart copy failed and retry is executed.
This function and retry handler was incorrect, those were a mistake in the processing, and those implementation was insufficient.
If you change the permissions of a file larger than 5GB, multipart copy processing will be performed, but these will fail in the retry processing.


ggtakec:
@gaul This PR is needed for #1098, so do the merge.
If you find a problem, please point out.

PR #1127, opened 2019-08-12
Added command line parameter ak_sk to support accessKeyId and secretA…

renxingliang:
…ccessKey commands

### Details

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -5036,6 +5036,21 @@ static int my_fuse_opt_proc(void* data, const char* arg, int key, struct fuse_ar
       return 0;
     }
 
+    if(0 == STR2NCMP(arg, "ak_sk=")){
+      std::string ak_sk = strchr(arg, '=') + sizeof(char);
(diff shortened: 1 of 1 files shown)


gaul:
Specifying the secret key on the command line is insecure since it leaks in `ps` listings.  Instead can you use the `AWSACCESSKEYID` and `AWSSECRETACCESSKEY` environment variables?

ggtakec:
@renxingliang I'm very sorry for my late reply.

The accessKeyId and secretAccessKey options are now obsolete.(If specified, an error will occur)
The reason is to avoid security risks.

The ak_sk option included in this PR has the same definition as the above two options, so it is difficult to support.

Can you tell me why you want to pass these with command line options?
And can it be replaced with environment variables as suggested by @gaul?
(At present, only the environment variable is left, but even this environment variable contains a little risk)

Finally, it is recommended that you avoid calling printf for secret key in this PR code.

gaul:
Closing due to lack of activity.

PR #1128, opened 2019-08-12
Fixed MultiRead method to exit after waiting other parts at error

ggtakec:
n/a

### Details
In multipart processing, when a part reaches the retry count limit due to get some error, an error(EIO) was returned at reaching limit.
However, this is wrong, and the S3fsMultiCurl::MultiRead() method must wait for receiving all part results before returning an error(EIO).


ggtakec:
**NOTE** This PR has not passed the linux-ppc64le test yet.
(The apt-get update command has failed)
Later, try linux-ppc64le again and merge if it passes.

PR #1130, opened 2019-08-13
Strict processing of MultiRead method

ggtakec:
#1128 

### Details
In S3fsMultiCurl::MultiRead() processing, I found a possibility of a problem, so I will correct it again.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -4152,14 +4152,17 @@ int S3fsMultiCurl::MultiRead()
 {
   int result = 0;
 
-  for(s3fscurllist_t::iterator iter = clist_req.begin(); iter != clist_req.end(); ++iter) {
+  for(s3fscurllist_t::iterator iter = clist_req.begin(); iter != clist_req.end(); ){
     S3fsCurl* s3fscurl = *iter;
 
     bool isRetry = false;
-
+    bool isPostpone = false;
     long responseCode = -1;
     if(s3fscurl->GetResponseCode(responseCode, false)){
-      if(400 > responseCode){
+      if(-1 == responseCode){
+        // This is a case where the processing result has not yet been updated (should be very rare).
+        isPostpone = true;
+      }else if(400 > responseCode){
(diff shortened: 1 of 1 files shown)


ggtakec:
@gaul If you find a problem, please let me know.

gaul:
I think we will need to use a fault injection library like [libfiu](https://blitiri.com.ar/p/libfiu/) if we really want to exercise these paths.

src/curl.cpp
```diff
@@ -4152,14 +4152,17 @@ int S3fsMultiCurl::MultiRead()
...
+  for(s3fscurllist_t::iterator iter = clist_req.begin(); iter != clist_req.end(); ){
     S3fsCurl* s3fscurl = *iter;
 
     bool isRetry = false;
-
+    bool isPostpone = false;
     long responseCode = -1;
     if(s3fscurl->GetResponseCode(responseCode, false)){
-      if(400 > responseCode){
+      if(-1 == responseCode){
```
gaul: To me it seems that if `GetResponseCode` ever returns -1, it will always return -1.  Since the above calls with `from_curl_handle` = false this will never update `LastResponseCode`.  Do I understand this correctly?

ggtakec: @gaul Thanks for your help.

After s3fscurl->LastResponseCode is initialized, it remains -1 until curl_easy_getinfo() is called in s3fscurl->RequestPerform().
And since this RequestPerform() is running in a separate thread from this S3fsMultiCurl::MultiRead(), it may return -1.
Therefore, this judgment is necessary.

But, this PR code was not enough.
S3fsCurl::RequestPerform() may return without setting LastResponseCode on error.

Thus if it just check "-1 == LastResponseCode", the problem remains.
It must be possible to distinguish whether the curl_easy_perform() function has not been called yet or whether the processing is an error.
These will be fixed after this comment.

ggtakec: #1136 fixes this.

PR #1134, opened 2019-08-21
Verify file size via metadata as well as data

gaul:


gaul:
Suggested by @kahing.

ggtakec:
@gaul and @kahing Thanks.

PR #1136, opened 2019-08-22
Refixed strict processing of MultiRead method

ggtakec:
#1130 

### Details
S3fsCurl::LastResponseCode has been changed so that the value is set when curl_easy_getinfo() is called and completes normally or reaches the maximum number of retries.
And, the code of the S3fsCurl::RequestPerform() method has been revised.


ggtakec:
@gaul Please check this PR.

PR #1137, opened 2019-08-26
use correct content-type when overwriting and copying files

liuyongqing:
When overwriting and copying a file, we need to use /etc/mime.types to decide the file Content-Type, rather than cached meta Content-Type。

### Details
When overwriting and copying files, s3fs use the original file Content-Type meta cache，the cached Content-Type may be not correct, for example the file upload by a machine doest not have file /etc/mine.types。So in order to allow user fixed the wrong Content-Type,  we need to use /etc/mime.types to decide the file Content-Type。

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -2961,13 +2961,14 @@ int S3fsCurl::PutHeadRequest(const char* tpath, headers_t& meta, bool is_copy)
   responseHeaders.clear();
   bodydata.Clear();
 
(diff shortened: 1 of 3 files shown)


liuyongqing:
It seems this mac platform doesn't exist file /etc/mime.types, so the test case run failed
image

liuyongqing:
@gaul ，hi，gaul，I have add a check to file /etc/mime.types  to decide run check_content_type test case.

gaul:
Thank you for your contribution @liuyongqing!

PR #1144, opened 2019-09-05
Use close-on-exec where possible

gaul:
This prevents theoretical security issues with untrusted child
processes.  Also remove two unneeded void non-parameters.  Found by
clang-tidy.


gaul:
Something is wonky with macOS `F_DUPFD_CLOEXEC` behavior; investigating...

gaul:
Abandoning.

PR #1146, opened 2019-09-07
fix deadlock due to fdmanager_lock and fdent_data_lock

liuyongqing:
### Details
Deadlocks will appear in the following situations：
1. thread A call s3fs_read, already get the fdent_data_lock and cache_cleanup_lock（for clean up cache dir）, trying to get fd_manager_lock
2. thread B call s3fs_open， already get  fd_manager_lock  and fdent_lock, trying to get fdent_data_lock
3. the deadlock occured between thread A and thread B, because each thread wants to get the lock of another thread
so we should use pthread_mutex_trylock for fdmanager when clean up cache

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -2143,7 +2143,8 @@ FdEntity* FdManager::GetFdEntity(const char* path, int existfd)
   return NULL;
 }
 
-FdEntity* FdManager::Open(const char* path, headers_t* pmeta, off_t size, time_t time, bool force_tmpfile, bool is_create, bool no_fd_lock_wait)
+FdEntity* FdManager::Open(const char* path, headers_t* pmeta, off_t size, time_t time, bool force_tmpfile,
+                          bool is_create, bool no_fd_lock_wait, bool no_manager_lock_wait)
 {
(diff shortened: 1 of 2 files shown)


ggtakec:
@liuyongqing Thanks for your kindness.
@gaul This fix seems no problem for me, please give us your opinion.

gaul:
@liuyongqing Could you share how to reproduce these symptoms?  I would like to verify the problem with ThreadSanitizer.

liuyongqing:
@gaul ,the reproduce step like this:
step 1:  use -ouse_cache=/cache_dir  -oensure_diskfree=1000  to mount s3fs, the option ensure_diskfree  is in order to quickly trigger clean up cache, you can adjust it according cache dir disk space.

step 2: several thread read a lot of data from multiple files, then one thread  will trigger clean up cache in FdEntity::Read with hold fdent_data_lock but not hold fd_manager_lock。(in order to increase the possibility of deadlock, we could add a sleep in CleanupCacheDirInternal function, before FdManager::get()->Open(next_path.c_str(), NULL, -1, -1, false, true, true) code line )

step 3: another thread open the same file as the thread who clean up cache, then it will get the fd_manager_lock, trying to get fdent_data_lock.

liuyongqing:
this fix still have problem when close the file

PR #1151, opened 2019-09-11
fix deadlock in clean up cache

liuyongqing:
@gaul ，the previous pull request is：https://github.com/s3fs-fuse/s3fs-fuse/pull/1146, still have some situations not fully considered, so I closed the previous pull request.

Deadlocks will appear in the following situations also：
thread A call s3fs_write, already get the fdent_data_lock and cache_cleanup_lock（for clean up cache dir）, after cleaning one file, and trying to get fd_manager_lock for Close the FdEntity.
thread B call s3fs_open for the same file，already get fd_manager_lock and fdent_lock, trying to get fdent_data_lock
the deadlock occured between thread A and thread B, because each thread wants to get the lock of another thread
so when cleanup a cache file, we should hold fd_manager_lock until it finished, if we cannot hold it, we can ignore it temporarily.

Another deadlock  condition is in NoCacheLoadAndPost, because it trying to hold fd_manager_lock after hold fdent_data_lock

In short，We should lock in the order fd_manager_lock->fdent_lock->fdent_data_lock， if we trying to hold fd_manager_lock after hold fdent_lock or fdent_data_lock，we should use try_lock. The NoCacheLoadAndPost deadlock condition I didn't find a nice way to solve it，but it will only triggered if didn't have enough disk after cleaned up cache.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -2223,22 +2223,6 @@ ssize_t FdEntity::Write(const char* bytes, off_t start, size_t size)
   return wsize;
 }
 
-void FdEntity::CleanupCache()
-{
-  AutoLock auto_lock(&fdent_lock, AutoLock::NO_WAIT);
-
-  if (!auto_lock.isLockAcquired()) {
-    return;
-  }
-
-  if(pagelist.IsModified()){
-    // cache is not committed to s3, cannot cleanup
-    return;
-  }
-
-  FdManager::DeleteCacheFile(path.c_str());
-}
-
 //------------------------------------------------
 // FdManager symbol
 //------------------------------------------------
@@ -2794,19 +2778,16 @@ void FdManager::CleanupCacheDirInternal(const std::string &path)
     if(S_ISDIR(st.st_mode)){
       CleanupCacheDirInternal(next_path);
     }else{
-      FdEntity* ent;
-      if(NULL == (ent = FdManager::get()->Open(next_path.c_str(), NULL, -1, -1, false, true, true))){
-        S3FS_PRN_DBG("skipping locked file: %s", next_path.c_str());
+      AutoLock auto_lock(&FdManager::fd_manager_lock, AutoLock::NO_WAIT);
+      if (!auto_lock.isLockAcquired()) {
+        S3FS_PRN_ERR("could not get fd_manager_lock when clean up file(%s)", next_path.c_str());
         continue;
       }
-
-      if(ent->IsMultiOpened()){
-        S3FS_PRN_DBG("skipping opened file: %s", next_path.c_str());
-      }else{
-        ent->CleanupCache();
+      fdent_map_t::iterator iter = fent.find(next_path);
+      if(fent.end() == iter) {
         S3FS_PRN_DBG("cleaned up: %s", next_path.c_str());
+        FdManager::DeleteCacheFile(next_path.c_str());
       }
-      Close(ent);
     }
   }
   closedir(dp);
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -656,7 +656,36 @@ function test_write_multiple_offsets {
     rm_test_file ${TEST_TEXT_FILE}
 }
 
+function test_clean_up_cache() {
+    describe "Test clean up cache"
+
+    dir="many_files"
+    count=256
+    mkdir -p $dir
+
+    for x in $(seq $count); do
+        dd if=/dev/urandom of=$dir/file-$x bs=1048576 count=1
+    done
+
+    file_cnt=$(ls $dir | wc -l)
+    if [ $file_cnt != $count ]; then
+        echo "Expected $count files but got $file_cnt"
+        rm -rf $dir
+        return 1
+    fi
+    CACHE_DISK_AVAIL_SIZE=`get_disk_avail_size $CACHE_DIR`
+    if [ "$CACHE_DISK_AVAIL_SIZE" -lt "$ENSURE_DISKFREE_SIZE" ];then
+        echo "Cache disk avail size:$CACHE_DISK_AVAIL_SIZE less than ensure_diskfree size:$ENSURE_DISKFREE_SIZE"
+        rm -rf $dir
+        return 1
+    fi
+    rm -rf $dir
+}
+
 function add_all_tests {
+    if `ps -ef | grep -v grep | grep s3fs | grep -q ensure_diskfree`; then
+        add_tests test_clean_up_cache
+    fi
     add_tests test_append_file 
     add_tests test_truncate_file 
     add_tests test_truncate_empty_file
--- a/test/small-integration-test.sh
+++ b/test/small-integration-test.sh
@@ -17,7 +17,15 @@ CACHE_DIR="/tmp/s3fs-cache"
 rm -rf "${CACHE_DIR}"
 mkdir "${CACHE_DIR}"
 
+#reserve 200MB for data cache
+source test-utils.sh
+CACHE_DISK_AVAIL_SIZE=`get_disk_avail_size $CACHE_DIR`
+ENSURE_DISKFREE_SIZE=$((CACHE_DISK_AVAIL_SIZE - 200))
+
+export CACHE_DIR
+export ENSURE_DISKFREE_SIZE 
 FLAGS=(
+    "use_cache=${CACHE_DIR} -o ensure_diskfree=${ENSURE_DISKFREE_SIZE}"
     enable_content_md5
     enable_noobj_cache
     nocopyapi
@@ -25,13 +33,12 @@ FLAGS=(
     notsup_compat_dir
     sigv2
     singlepart_copy_limit=$((10 * 1024))  # limit size to exercise multipart code paths
-    use_cache="${CACHE_DIR}"
     #use_sse  # TODO: S3Proxy does not support SSE
 )
 
 start_s3proxy
 
-for flag in ${FLAGS[*]}; do
+for flag in "${FLAGS[@]}"; do
     echo "testing s3fs flag: $flag"
 
     start_s3fs -o $flag
--- a/test/test-utils.sh
+++ b/test/test-utils.sh
@@ -258,6 +258,11 @@ function check_content_type() {
     fi
 }
 
+function get_disk_avail_size() {
+    DISK_AVAIL_SIZE=`BLOCKSIZE=$((1024 * 1024)) df $1 | awk '{print $4}' | tail -n 1`
+    echo ${DISK_AVAIL_SIZE}
+}
+
 function aws_cli() {
     AWS_ACCESS_KEY_ID=local-identity AWS_SECRET_ACCESS_KEY=local-credential aws $* --endpoint-url "${S3_URL}" --no-verify-ssl 
 }
--- a/src/fdcache.h
+++ b/src/fdcache.h
@@ -166,7 +166,6 @@ class FdEntity
 
     void Close(void);
     bool IsOpen(void) const { return (-1 != fd); }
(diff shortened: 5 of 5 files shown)


src/fdcache.cpp
```diff
@@ -2364,19 +2364,16 @@ void FdManager::CleanupCacheDirInternal(const std::string &path)
...
-      FdEntity* ent;
-      if(NULL == (ent = FdManager::get()->Open(next_path.c_str(), NULL, -1, -1, false, true, true))){
-        S3FS_PRN_DBG("skipping locked file: %s", next_path.c_str());
+      AutoLock auto_lock(&FdManager::fd_manager_lock, AutoLock::NO_WAIT);
+      if (!auto_lock.isLockAcquired()) {
+        S3FS_PRN_ERR("could not get fd_manager_lock when clean up file(%s)", next_path.c_str());
         continue;
       }
-
-      if(ent->IsMultiOpened()){
```
gaul: Can we remove the definition of `IsMultiOpened`?

src/fdcache.cpp
```diff
@@ -2364,19 +2364,16 @@ void FdManager::CleanupCacheDirInternal(const std::string &path)
...
-        S3FS_PRN_DBG("skipping opened file: %s", next_path.c_str());
-      }else{
-        ent->CleanupCache();
+      fdent_map_t::iterator iter = fent.find(next_path);
+      if(fent.end() == iter) {
         S3FS_PRN_DBG("cleaned up: %s", next_path.c_str());
+        FdManager::DeleteCacheFile(next_path.c_str());
```
gaul: If I understand this correct, we do not need to check `FdEntity::pagelist.IsModified` since there are no opens and thus no modified data?

test/test-utils.sh
```diff
@@ -258,6 +258,11 @@ function check_content_type() {
     fi
 }
 
+function get_disk_avail_size() {
+    DISK_AVAIL_SIZE=`df $1 --output=avail |tail -n 1|tr -dc '0-9'`
```
gaul: This does not work on macOS.  Instead try:

```bash
BLOCK_SIZE=$((1024 * 1024)) df $1 | awk '{print $4}' | tail -n 1
```

test/integration-test-main.sh
```diff
@@ -567,6 +567,30 @@ function test_open_second_fd {
...
+
+    dir="many_files"
+    count=256
+    mkdir -p $dir
+
+    for x in $(seq $count); do
+        dd if=/dev/urandom of=$dir/file-$x bs=1048576 count=1
+    done
+
+    file_cnt=$(ls -1 $dir | wc -l)
```
gaul: The `-1` is unnecessary in non-interactive shells/pipelines.

test/integration-test-main.sh
```diff
@@ -597,6 +621,10 @@ function add_all_tests {
     add_tests test_concurrency
     add_tests test_concurrent_writes
     add_tests test_open_second_fd
+    ENSURE_DISKFREE_SIZE=`ps -ef|grep  ensure_diskfree|awk '{print $NF}' |tr -dc '0-9'`
+    if [ ! -z "$ENSURE_DISKFREE_SIZE" ]; then
```
gaul: Could you simplify this to:

```bash
if pidof s3fs | grep -q ensure_diskfree; then
```

gaul:
Please add `pidof` to the `brew install` section of `.travis.yml` to address macOS failures.

test/integration-test-main.sh
```diff
@@ -597,6 +621,9 @@ function add_all_tests {
     add_tests test_concurrency
     add_tests test_concurrent_writes
     add_tests test_open_second_fd
+    if pidof s3fs | xargs -I {} ps -o cmd -fp {} | grep -q ensure_diskfree; then
```
gaul: macOS Travis complains that `cmd` is not found -- is there another way to do this?

https://travis-ci.org/s3fs-fuse/s3fs-fuse/jobs/585372211#L3704

ggtakec:
@liuyongqing #1169 will solve the build error on OSX. Please try it.

ggtakec:
@liuyongqing I'm sorry that #1171 issue makes build error now, please wait for fixing it.

liuyongqing:
@ggtakec ,the -oensure_diskfree param seems have some problem in mac os，the simplest ensure_diskfree mount test code can't pass test case：https://github.com/s3fs-fuse/s3fs-fuse/pull/1170/commits

liuyongqing:
@ggtakec ，do you know how to fix ensure_diskfree mount can't pass test case in MacOS?

gaul:
@liuyongqing do we have a path forward on this pull request?

liuyongqing:
@gaul ，can we ignore the check for mac os temporarily，according to this test pull request：https://github.com/s3fs-fuse/s3fs-fuse/pull/1170/commits，the failed test case not related to this commit.

test/integration-test-main.sh
```diff
@@ -567,6 +567,30 @@ function test_open_second_fd {
...
+    mkdir -p $dir
+
+    for x in $(seq $count); do
+        dd if=/dev/urandom of=$dir/file-$x bs=1048576 count=1
+    done
+
+    file_cnt=$(ls $dir | wc -l)
+    if [ $file_cnt != $count ]; then
+        echo "Expected $count files but got $file_cnt"
+        return 1
```
ggtakec: It is better to call "rm -rf $dir" before returning, because of next testing.

test/integration-test-main.sh
```diff
@@ -567,6 +567,30 @@ function test_open_second_fd {
...
+    fi
+    CACHE_DISK_AVAIL_SIZE=`get_disk_avail_size $CACHE_DIR`
+    if [ "$CACHE_DISK_AVAIL_SIZE" -lt "$ENSURE_DISKFREE_SIZE" ];then
+        echo "Cache disk avail size:$CACHE_DISK_AVAIL_SIZE less than ensure_diskfree size:$ENSURE_DISKFREE_SIZE"
```
ggtakec: It is better to call "rm -rf $dir" before returning, because of next testing, too.

test/small-integration-test.sh
```diff
@@ -17,6 +17,11 @@ CACHE_DIR="/tmp/s3fs-cache"
 rm -rf "${CACHE_DIR}"
 mkdir "${CACHE_DIR}"
 
+#reserve 200MB for data cache
+source test-utils.sh
+CACHE_DISK_AVAIL_SIZE=`get_disk_avail_size $CACHE_DIR`
+ENSURE_DISKFREE_SIZE=$((CACHE_DISK_AVAIL_SIZE - 200))
```
ggtakec: test_clean_up_cache function uses 256MB, so here I think it should be "- 256".

test/test-utils.sh
```diff
@@ -258,6 +258,11 @@ function check_content_type() {
     fi
 }
 
+function get_disk_avail_size() {
+    DISK_AVAIL_SIZE=`BLOCK_SIZE=$((1024 * 1024)) df $1 | awk '{print $4}' | tail -n 1`
```
ggtakec: Please change the environment variable name to "BLOCKSIZE" instead of "BLOCK_SIZE".
This is the environment variable name common to MacOS and other OS.

ggtakec:
@liuyongqing Sorry for my late review.

One of reason for this test failure is:
```
DISK_AVAIL_SIZE=`BLOCK_SIZE=$((1024 * 1024)) df $1 | awk '{print $4}' | tail -n 1`
```
"BLOCK_SIZE" is not valid in Mac OS df.
But we can use "BLOCKSIZE" instead.
The environment variable "BLOCKSIZE" can be used on Mac OS and other OS.
In Mac OS df, the only acceptable environment variable name is "BLOCKSIZE".

And the other is the following part.
```
ENSURE_DISKFREE_SIZE=$((CACHE_DISK_AVAIL_SIZE - 256))
```
200MB is defined as free space, but the test_clean_up_cache test uses 256MB.
So this part seems to be correct for "-256".

Please confirm these points in the conversation of the source code.
(I made the same fix to master as you and confirmed that TravisCI was no longer a problem.)
Please try rebasing this PR code on the latest master.

liuyongqing:
@ggtakec ，hi，I changed the code according to the above advice, but it still failed in macos

ggtakec:
@liuyongqing Thanks for your help.
When I checked it correctly again, I got the same your error.

I think the reason which is that we run some tests and fter that we run **"use_cache = $ {CACHE_DIR} -o ensure_diskfree = $ {ENSURE_DISKFREE_SIZE}"** final test.
And it seems that s3fs got the following error at that time.
```
"There is no enough disk space for used as cache(or temporary) directory by s3fs."
```
This is because we determined the ensure_diskfree size before the test started, but after some tests it seems that the disk usage is different.
(For example, disk space is used in s3proxy, cache directory, etc.)
So I think we should clear the disk before running the ensure_diskfree test or run this test right after calculating ENSURE_DISKFREE_SIZE.

What about running the ensure_diskfree test first to solve it?
To do this, change FLAGS in the small-integration-test.sh file as follows:
```
FLAGS=(
    "use_cache=${CACHE_DIR} -o ensure_diskfree=${ENSURE_DISKFREE_SIZE}"
    enable_content_md5
    enable_noobj_cache
    nocopyapi
    nomultipart
    notsup_compat_dir
    sigv2
    singlepart_copy_limit=$((10 * 1024))  # limit size to exercise multipart code paths
    #use_sse  # TODO: S3Proxy does not support SSE
)
```
I have confirmed that this procedure does not cause any errors.

Please retry to change your code and run test again.

liuyongqing:
@ggtakec ，the test still failed, complain:The job exceeded the maximum log length, and has been terminated.

ggtakec:
@liuyongqing Thanks for your reports.
_My test (the source codes probably are the same) did not put any error, but the source code probably is as same as yours._

I found following error in the TravisCI logs.
```
./integration-test-main.sh: line 677: [: : integer expression expected
test_clean_up_cache passed
```
The test_clean_up_cache test passed, but an error occured in that test script.
Due to this error, it seems that 256 files have not been deleted and there is not enough disk space for s3fs.
So it seems that some tests(maybe writing files) have failed since then.

I think we need to find out why this script error is occurring.

ggtakec:
@liuyongqing I found the reason of following error message. 
```
./integration-test-main.sh: line 677: [: : integer expression expected
```
The reason is not clear, but on MacOS, the variables CACHE_DIR and
ENSURE_DISKFREE_SIZE is not found in the function of integration-test-main.sh.
So it seems that you can work around this problem by exporting them in the definition of each variable in small-integration-test.sh.

In addition to this problem, some test units(functions) forgot to delete test files.
It has been merged into master in #1230.

Even after solving the above two problems, the test after test_chown still seems to fail.
I will continue to investigate, but I'm sorry because it is not in time for the release of v1.86.

ggtakec:
@liuyongqing I merged #1232.
Please export CACHE_DIR and ENSURE_DISKFREE_SIZE variables at where these are defined(please see in my previous comment), and try PR build again.
I hope that you should have a successful build of this PR.

liuyongqing:
@ggtakec ，hi，the macos run test success, but ppc64Ie platform run test case failed

ggtakec:
@liuyongqing It was "The build has been terminated".
Then I would try to rebuild ppc64le now.
Please wait, and if the build on travis would success, please squash the PR code.
After that, I will merge this PR.

liuyongqing:
@ggtakec ，all test pass，commits are squashed to one commit.

ggtakec:
@liuyongqing Thank you for your cooperation.
I'll merge this PR.
We apologize for taking a long time.
But thanks to it, s3fs has gotten better.

Issue #1016, opened 2019-04-18
Trouble Disabling Cache Once Set


jamiejackson:
#### Version of s3fs being used (s3fs --version)
Amazon Simple Storage Service File System V1.85(commit:unknown) with OpenSSL

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
Name        : fuse
Version     : 2.9.2
Release     : 11.el7
Architecture: x86_64

#### Kernel information (uname -r)
3.10.0-957.1.3.el7.x86_64

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"

#### s3fs command line used, if applicable
```
```
#### /etc/fstab entry, if applicable
```
s3fs#files-dev.mysite.com:/legacy_media_share/ /mnt/s3/files-dev fuse _netdev,nonempty,iam_role,allow_other,use_path_request_style,use_cache='',url=http://s3.amazonaws.com,uid=root,gid=hudx_admin,umask=0002,rw 0 0
/dev/mapper/rhel-swap swap swap defaults 0 0
```
#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
_if you execute s3fs with dbglevel, curldbg option, you can get detail debug messages_
```
```
### Details about issue

What are the minimum steps needed to go from using cache (I'd had `use_cache=/tmp`) to disabling it? I tried:

* Both removing the `use_cache` option altogether and having `use_cache=""`
* `sudo umount /mnt/s3/files-dev`
* `sudo rm -rf /tmp/files-dev.mysite.com`
* `sudo mount /mnt/s3/files-dev`

[...]

PR #1165, opened 2019-10-07
Fix use_cache='' handling

olfway:
Do not use "/''" as cache directory when mounting with fstab

 #1016 

### Details
When mounting from fstab it actually called with `use_cache=''` argument and it creates and uses `/''` directory as cache

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -2178,7 +2178,7 @@ off_t           FdManager::free_disk_space = 0;
 //------------------------------------------------
 bool FdManager::SetCacheDir(const char* dir)
 {
-  if(!dir || '\0' == dir[0]){
+  if(!dir || '\0' == dir[0] || 0 == strcmp(dir, "''") || 0 == strcmp(dir, "\"\"")){
     cache_dir = "";
   }else{
     cache_dir = dir;


gaul:
Can the user just provide `use_cache=` instead?  I worry about inconsistently trying to handle quoting issues across all the parameters.

olfway:
I've tried on centos 7
if I try to mount it with use_cache= parameter in fstab, I got error:
fuse: unknown option `use_cache'

fuse-2.9.2-11.el7.x86_64
s3fs-fuse-1.85-1.el7.x86_64

Also, in every man pages on the internet:
-o use_cache (default="" which means disabled)

ggtakec:
@olfway I'm sorry for late reply.

I am hesitant about this proposal for two reasons.
- Common handling of optional parameters  
This is the same opinion as @gaul.  
There is no definition of quotes and double quotes in the current option specification method, and the specified all string is handled as an option parameter.  
In other words, if we supports like this PR code, it is rather than just using the use_cache option, I think that all options should handle parameters in the same way.
- How to make cache unspecified  
s3fs-fuse runs and disables the cache unless the use_cache option is specified.  
In other words, there is no need to specify use_cache="" option and its parameter.  
If we need the option to disable caching, we should add options such as "not_use_cache"(dedicated to disable).  
However, because the caching is disabled by default, this option is redundant and I think it will be confused users.

For the above reasons, we will not merge the contents of this PR.

If you need it for any reason(which we are not yet aware), please re-open it.

PR #1168, opened 2019-10-11
In memory cache for softlinks

AgrawalAmey:
### Details
Adds a hashmap containing mapping of softlink path and destination.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -141,7 +141,9 @@ struct header_nocase_cmp : public std::binary_function<std::string, std::string,
     return (strcasecmp(strleft.c_str(), strright.c_str()) < 0);
   }
 };
+
 typedef std::map<std::string, std::string, header_nocase_cmp> headers_t;
+typedef std::map<std::string, std::string> symlink_cache_t;
 
 //
 // Header "x-amz-meta-xattr" is for extended attributes.
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -113,6 +113,7 @@ std::string instance_name;
 s3fs_log_level debug_level        = S3FS_LOG_CRIT;
 const char*    s3fs_log_nest[S3FS_LOG_NEST_MAX] = {"", "  ", "    ", "      "};
 std::string aws_profile           = "default";
+symlink_cache_t symlink_cache;
 
 //-------------------------------------------------------------------
 // Static variables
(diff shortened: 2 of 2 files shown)


ggtakec:
@AgrawalAmey I'm sorry for my late reply.

Below are my opinions and questions.

First of all, I don't think it is necessary to have only a symbolic link in a separate cache.
The pseudo-symlink created by s3fs on S3(or other distributed storage) is the destination path saved in the object(file) corresponding to the path.
In other words, I believe that the current Stats cache and file object cache(fdcache) can be enough to cache for symbolic link.

In this PR code, there is a dedicated cache for symbolic links in s3fs process, but there seems to be no expiration time or upper limit for updating this cache.
Thus, s3fs once cached links, it will remain cached unless you update the symbolic link through this s3fs process.
Also, if the number of caches is large, memory will be squeezed.

In my opinion based on these, it seems that the reason for the performance improvement you described is due to the fact that this dedicated cache does not cause cache out.

I think that the performance will be improved in the same way by specifying options such as use_cache, max_stat_cache_size, stat_cache_expire, stat_cache_interval_expire, enable_noobj_cache, etc. with the current s3fs.

Could you try a performance comparison with these options?

AgrawalAmey:
Hi @ggtakec ,

I had tried that setting the other cache options, but it does not help. Sorry, I am new to both C++ world and system programming, please correct if I am wrong. But the following two snippets seem responsible,
```c++
time_t mtime         = (!S_ISREG(stobj.st_mode) || S_ISLNK(stobj.st_mode)) ? -1 : stobj.st_mtime;
```
```c++
    if(stat(cachepath.c_str(), &st) == 0){
      if(st.st_mtime < time){
        S3FS_PRN_DBG("cache file stale, removing: %s", cachepath.c_str());
        if(unlink(cachepath.c_str()) != 0){
          return (0 == errno ? -EIO : -errno);
        }
      }
    }
```
I agree that we do not need to introduce a new cache here, but I felt it had two benefits, first that it it is an in-memory cache rather than an on-disk cache. Second, it did not disturb any prior code flow. I agree that the cache must be bounded in time/size.

ggtakec:
@AgrawalAmey 
I'm sorry, my opinion was incomplete.
Your indication is correct.

The cache file by use_cache cannot support symbolic links.
I think there will be a problem with the operation as it is, then I will check this problem and investigate how it can be cached using the current stats and fdcache.

I'm sorry, please wait a little longer.
Thank you for pointing out.

AgrawalAmey:
Thanks a lot @ggtakec

ggtakec:
@AgrawalAmey 
I used your this PR as a reference and generated a PR( #1209 ) that implemented cashout.
Your basic logic remains the same, but the code has changed significantly from your PR because you have incorporated cache-out into the Stats cache mechanism.
Please forgive me for creating another PR.

Since the logic remains the same, the contents of symbolic links can be cached, so the performance is as good as your PR.

If possible, I would appreciate it if you checked the operation with the PR code.

gaul:
Superseded by #1209.

PR #1170, opened 2019-10-14
just test ensure_diskfree param

stealflowercow:
just test ensure_diskfree param

### DIFF
--- a/src/fdcache.h
+++ b/src/fdcache.h
@@ -70,7 +70,6 @@ typedef std::list<struct fdpage> fdpage_list_t;
 
 class FdEntity;
 
-//
 // Management of loading area/modifying
 //
 // cppcheck-suppress copyCtorAndEqOperator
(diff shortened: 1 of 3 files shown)


gaul:
@stealflowercow could you explain this pull request more fully?  I can guess what you are doing but I'm still not sure.  Also can you remove the unneeded formatting changes?

ggtakec:
@stealflowercow
I think it should be divided into tests that specify ensure_diskfree and tests that do not specify(current test codes).
This is because this PR code makes it impossible to perform tests without specifying ensure_diskfree.
(Also, please undo unnecessary code changes.)
Last, Now that your OSX test may have failed, try rebase to the latest master branch.

@gaul
If you add this test, the test will take about +2 minutes.

Thanks for your help.

ggtakec:
The contents of this PR will be included in #1151 and will be withdrawn.

Issue #1171, opened 2019-10-14
Error test_write_multiple_offsets test on OSX


ggtakec:
#### Version of s3fs being used (s3fs --version)
SHA1: db338b3

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
n/a

#### Kernel information (uname -r)
OSX

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
n/a

#### s3fs command line used, if applicable
s3fs-fuse test on travis
```
testing s3fs flag: singlepart_copy_limit=10240
stdbuf -oL -eL ../src/s3fs s3fs-integration-test s3fs-integration-test -o use_path_request_style -o url=https://127.0.0.1:8080 -o no_check_certificate -o ssl_verify_hostname=0 -o use_xattr=1 -o createbucket -o passwd_file=passwd-s3fs -o direct_io -o auto_cache -o stat_cache_expire=1 -o stat_cache_interval_expire=1 -o dbglevel=info -o retries=3 -f -o singlepart_copy_limit=10240
```
#### /etc/fstab entry, if applicable
n/a

#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
https://travis-ci.org/s3fs-fuse/s3fs-fuse/jobs/597622065

### Details about issue
In the TravisCI test of s3fs-fuse, if singlepart_copy_limit=10240, an error occurred in test_write_multiple_offsets test.
This is the same as pointed out by @gaul in #1160.

The error contents are as follows.
https://gist.github.com/ggtakec/8394e005ec39f89a43083dd6a86b0cff

ggtakec:
This bug was found to occur in the test_rename_before_close test.
In "rename before close file", there was a problem with the number of references to the file discriptor.
When renaming the file, the file cache and file status cache for that file in s3fs are not closed, and file initialization(create, truncate) of test_write_multiple_offsets test cannot be performed normally.
Therefore, test_write_multiple_offsets test file at the start of testing cannot be initialized normally, so an error occurs in this test.

I posted PR #1181 that fixed this problem.

PR #1173, opened 2019-10-14
Fixed test_write_multiple_offsets error on osx

ggtakec:
#1171 

### Details
The cause of #1171 is not clear yet.
However, from the travis log, it seems the file size was not 0 even the file was opened with truncating.
(Further inspection is required for file cache clearing errors or file cache size calculation errors.)
In order to avoid test_write_multiple_offsets test issue, I will use another file name for this test.

After merging this fix, investigation continues for #1171


ggtakec:
@gaul Since #1171 affects other PRs, this PR will be merged.

PR #1181, opened 2019-10-23
Fixed a rename bug

ggtakec:
#1171 #1173 

### Details
The bug #1171 is due to forgetting to close the file descriptor in s3fs_rename() function.
This caused errors in tests after the test_rename_before_close test.

By correcting this forgetting to close, I discovered a problem in the rename processing.
During the rename processing, s3fs did not rename the file cache and file status cache.
In addition, the target file objects(FdEntity) were not detected properly when the cache directory was not specified.

Therefore, these two problems are corrected by this PR.

Also, the test code that temporarily avoided the error in #1173 has been restored.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -141,6 +141,48 @@ bool CacheFileStat::DeleteCacheFileStatDirectory()
   return delete_files_in_dir(top_path.c_str(), true);
 }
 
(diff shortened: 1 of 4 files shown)


ggtakec:
@gaul Please review this PR.

ggtakec:
@gaul This must first be merged because it is likely to be the cause of other PR test errors.
If you find a problem, please let me know. I will make corrections as soon as possible.

Issue #1184, opened 2019-10-26
prepare_url screws up the uri if the bucket name and endpoint prefix match


bryceml:
#### Version of s3fs being used (s3fs --version)
Amazon Simple Storage Service File System V1.84(commit:unknown) with GnuTLS(gcrypt)
Copyright (C) 2010 Randy Rizun <rrizun@gmail.com>
License GPL2: GNU GPL version 2 <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
Version: 2.9.9-1ubuntu1

#### Kernel information (uname -r)
5.3.0-19-generic

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
Ubuntu19.10

#### s3fs command line used, if applicable
```
s3fs -f /tmp/test -o url=http://s3.repo.saltstack.com -o use_path_request_style -o public_bucket=1 -o bucket=s3
```
#### /etc/fstab entry, if applicable
```
```
#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
```
[INF] s3fs.cpp:s3fs_init(3379): init v1.84(commit:unknown) with GnuTLS(gcrypt)
[CRT] s3fs.cpp:s3fs_check_service(3780): bucket not found(host=http://s3.repo.saltstack.com) - result of checking service.
```
### Details about issue
What I can determine from the issue, is that if the beginning of the url before the first period matches the bucket name, it screws up calculating the bucket.

What works:
s3fs -f /tmp/test -o url=http://dezb4fersljmf.cloudfront.net -o use_path_request_style -o public_bucket=1 -o bucket=s3

If I do a wireshark packet capture, I see it is checking /s3/ as the uri

What doesn't work:
s3fs -f /tmp/test -o url=http://s3.repo.saltstack.com -o use_path_request_style -o public_bucket=1 -o bucket=s3

If I do a wireshark packet capture, I see it is trying to fetch /s3//s3/.repo.saltstack.com/s3/

I re-ran it with debug level info and got the following:
```
[CRT] s3fs.cpp:set_s3fs_log_level(273): change debug level from [CRT] to [INF] 
[...]

PR #1185, opened 2019-10-26
handle edge case when bucket name and first part of endpoint match

bryceml:
https://github.com/s3fs-fuse/s3fs-fuse/issues/1184

### Details
Fixes when the bucket name and first part of endpoint match by skipping the first part, either `https://` or `http://`


ggtakec:
@bryceml Thanks for your help.

Issue #723, opened 2018-02-23
/usr/src/s3fs-fuse# ./autogen.sh


Amritmatti:
./configure: line 4972: syntax error near unexpected token `common_lib_checking,                                                                                        '
./configure: line 4972: `PKG_CHECK_MODULES(common_lib_checking, fuse >= ${min_fu                                                                                        se_version} libcurl >= 7.0 libxml-2.0 >= 2.6 )'

i'm facing this problem plz help

nmgeek:
Try this:

```
sudo apt-get update
sudo apt-get install automake
rm configure
./autogen.sh
./configure
```

This error happened to me also.  While I can't completely back up to see how it happened it seems like I had an old apt repo configuration (because I had not run apt-get update in months) and an old version of automake was installed.  It was missing one program which caused a broken instance of the configure script to be created.  The broken configure script has the error shown in the first comment, above.

Fyb3roptik:
I am getting the same thing. This did not fix it for me

gaul:
Could you share your operating system?  This should work for Ubuntu 16.04 and newer.

bbxiao1:
I’m seeing this also on Manjaro.

bbxiao1:
FWIW my issue was solved by installing `pkgconf`

gaul:
Please try the suggested workaround and reopen the issue if symptoms persist.

PR #1197, opened 2019-11-18
Specfiy pkg-config as a necessary dependency for compilation

bbxiao1:
https://github.com/s3fs-fuse/s3fs-fuse/issues/723

### Details
I don't usually compile from source so having dependencies spelled out helps newbies like me. I have the comment `or your OS equivalent` because it was `pkgconf` on Manjaro.


ggtakec:
@bbxiao1 Thanks for your kindness.

PR #1199, opened 2019-11-18
Adds requester_pays option to enable requests involving Requester Pays buckets

ahuarte47:
This PR adds `requester_pays` option, false by default, to enable requests involving Requester Pays buckets.

Related issue:
https://github.com/s3fs-fuse/s3fs-fuse/issues/635


ggtakec:
@ahuarte47 @3DTOPO I'm sorry for late reply.
I'll merge this PR soon.

ahuarte47:
Thank you very much @ggtakec !

cprieger:
What version will this be available in? 

Sorry this may be a beginner question. But I am also processing some imagery and have come across the same use case.

gaul:
1.86 should include this feature.

PR #1200, opened 2019-11-18
Use travis cache for brew on osx

ggtakec:
n/a

### Details
The build for OSX was very slow and I felt it needed to be improved.

The reason is that the following processes are long.
- brew tap homebrew / homebrew-cask
- brew install awscli cppcheck truncate

Thus I changed to cache homebrew using TravisCI cache.
If a cache exists, over 800 seconds required for the above two processes can be shortened to about 30 seconds.

The correct way to cache homebrew is to leave it to TravisCI, but I could not shorten by that way.
So that, I add manually setting cache in .travis.yml, and it works fine for now.
And, if you want to create a new TravisCI Cache, you can delete it from the WebUI and recreate it.

I hope that merging this PR will reduce the time stress on the developer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,11 +43,40 @@ matrix:
 
     - os: osx
       osx_image: xcode9.2
+      cache:
+        directories:
+          - $HOME/Library/Caches/Homebrew
+          - /usr/local/Homebrew
+          - $HOME/.osx_cache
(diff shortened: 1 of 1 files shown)


ggtakec:
@gaul This has been merged. If you find a problem, please let me know.

Issue #1164, opened 2019-10-01
segfault mounting through /etc/fstab


sbonasu:
#### Version of s3fs being used (s3fs --version)
s3fs --version
Amazon Simple Storage Service File System V1.85 (commit:1db94a0) with OpenSSL

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
Name        : fuse
Version     : 2.9.2
Release     : 11.el7
Architecture: x86_64
Install Date: Fri 27 Sep 2019 11:30:35 PM UTC
Group       : System Environment/Base
Size        : 223297
License     : GPL+
Signature   : RSA/SHA256, Mon 06 Aug 2018 01:13:08 PM UTC, Key ID 199e2f91fd431d51
Source RPM  : fuse-2.9.2-11.el7.src.rpm
Build Date  : Tue 24 Jul 2018 02:58:28 PM UTC
Build Host  : x86-039.build.eng.bos.redhat.com
Relocations : (not relocatable)
Packager    : Red Hat, Inc. <http://bugzilla.redhat.com/bugzilla>
Vendor      : Red Hat, Inc.
URL         : https://github.com/libfuse/libfuse
Summary     : File System in Userspace (FUSE) utilities

#### Kernel information (uname -r)
3.10.0-1062.1.2.el7.x86_64

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
cat /etc/redhat-release
Red Hat Enterprise Linux Server release 7.7 (Maipo)

#### s3fs command line used, if applicable
works fine
s3fs bucketname /directory -o iam_role="auto" -o url="https://s3-us-east-2.amazonaws.com" -o endpoint=us-east-2 -o dbglevel=info -o curldbg -o allow_other -o use_cache=/tmp

#### /etc/fstab entry, if applicable
s3fs#bucketname /directory fuse fuse _netdev,iam_role=auto,rw,nosuid,nodev,allow_other,nonempty,endpoint=us-east-2,url=http://s3-us-east-2.amazonaws.com 0 0

Doesn't work
s3fs: unable to access MOUNTPOINT xxx: Transport endpoint is not connected

#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
s3fs.cpp:set_s3fs_log_level(297): change debug level from [CRT] to [INF]
[...]

PR #1209, opened 2019-11-26
In memory cache for softlinks with cache out - #1168

ggtakec:
#1168

### Details
This PR is #1164 suggested by @AgrawalAmey.
It incorporates a symbolic link cache, and this PR is added the cash out mechanism.

The symlink caching method is the same as suggested by @AgrawalAmey.
The contents including cash out are shown below.

#### Cache contents
Cache the link destination path using the file path of the symbolic link as a key.
The cache time and reference count are also stored as cache data.
These additional elements are just like the Stats cache.

#### Cache size, expiration date, cash out
I decided to use Stats cache options for this cache size, expiration date, and cashout.
In other words, the symbolic link cache coexists with Stats cache and exists in the class that implements the Stats cache.

This is because s3fs implements symbolic links as S3 objects(files), but it does not exist as a file system.
Therefore, it is easier to implement a simple cache than to modify fdcache (or prepare an equivalent).
Also, when considering the mechanism of cashout, the same implementation as Stats cache makes it easier to implement and show to users.

The Stats cache settings max_stat_cache_size, stat_cache_expire, and stat_cache_interval_expire are shared with this symbolic link cache.
(The maximum number of Stats caches and symbolic links is the number specified by max_stat_cache_size, respectively)

#### Cash in, out
A symbolic link cache is created when a symbolic link is created. (At this time, the Stats cache is cached out)
The Stats cache for symbolic link files is not synchronized with the creation and access of symbolic link files. (Same as before)
When deleting a symbolic link, both the symbolic link and the Stats cache are deleted.

Thanks to @AgrawalAmey for great suggestions and PR.


AgrawalAmey:
@ggtakec sorry for the delayed response. For some reason, I am unable to mount with this PR.  Not sure if it is because of this commit or some other commit in master. The  following line is being logged in a endless loop, 

```
[INF]       curl.cpp:GetIAMCredentials(2855): [IAM role=qa]
[INF] curl.cpp:GetHandler(296): Pool empty: force to create new handler```
```

Issue #1196, opened 2019-11-12
Mounting S3 Bucket using IAM Role


aric49:
### Additional Information
In our AWS account, we have two EC2 instances configured with an IAM role which allows read and write access to our S3 bucket.   One of our instances is able to successfully mount this instance using the IAM role while the other instance cannot.   Both instances were working successfully until we took one of the instances down for a reboot earlier today.

#### Version of s3fs being used (s3fs --version)
On both instances: 
Amazon Simple Storage Service File System V1.85 (commit:cc4a307) with OpenSSL  

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
On both instances: 
```
rpm -qi fuse
Name        : fuse
Version     : 2.9.4
Release     : 1.18.amzn1
Architecture: x86_64
Install Date: Mon 26 Aug 2019 12:22:57 AM UTC
Group       : System Environment/Base
Size        : 213917
License     : GPL+
Signature   : RSA/SHA256, Thu 18 Apr 2019 06:20:01 PM UTC, Key ID bcb4a85b21c0f39f
Source RPM  : fuse-2.9.4-1.18.amzn1.src.rpm
Build Date  : Fri 12 Apr 2019 05:49:04 PM UTC
Build Host  : build-64001.build
Relocations : (not relocatable)
Packager    : Amazon.com, Inc. <http://aws.amazon.com>
Vendor      : Amazon.com
URL         : http://fuse.sf.net
Summary     : File System in Userspace (FUSE) utilities
Description :
With FUSE it is possible to implement a fully functional filesystem in a
userspace program. This package contains the FUSE userspace tools to
mount a FUSE filesystem.

```

#### Kernel information (uname -r)
```
4.14.146-93.123.amzn1.x86_64
```

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
```
NAME="Amazon Linux AMI"
VERSION="2018.03"
ID="amzn"
ID_LIKE="rhel fedora"
VERSION_ID="2018.03"
PRETTY_NAME="Amazon Linux AMI 2018.03"
ANSI_COLOR="0;33"
CPE_NAME="cpe:/o:amazon:linux:2018.03:ga"
HOME_URL="http://aws.amazon.com/amazon-linux-ami/"
```

#### s3fs command line used, if applicable
```
[...]

PR #1233, opened 2020-01-29
break recursion when calling GetIAMCredentials

rallister:
https://github.com/s3fs-fuse/s3fs-fuse/issues/1196

### Details
Added a flag to not insert authHeaders when calling GetIAMCredentials, because they are not needed in this call but also to break recursion which occurs in issue 1196
CheckIAMCredentialUpdate() -> 
   GetIAMCredentials() -> 
       RequestPerform() ->
          insertAuthHeaders() ----> back to CheckIAMCredentialUpdate()

You might come up with a better way but this quickly got me out of a bind.


ggtakec:
@rallister Thanks for your great help.
I think that this PR code can fix the problem (loop).
I'd like to make minor changes to the code, so I'll comment and check it out. (Trivial description change)

src/curl.cpp
```diff
@@ -2384,7 +2384,9 @@ int S3fsCurl::RequestPerform()
   long responseCode;
   int result        = S3FSCURL_PERFORM_RESULT_NOTSET;
 
-  insertAuthHeaders();
+  if(!dontAddAuthHeaders)
+     insertAuthHeaders();
+
```
ggtakec: Please enclose the condition sentence in curly braces.

ggtakec:
@rallister Maybe you might fail with MacOS test on Travis.
Now I merged into master to remove the error.
If you would get an error, please build again.

rallister:
commits simply refuse to squash.

rallister:
@ggtakec ok it's done.

ggtakec:
@rallister Thanks and I merged this PR.

Issue #1203, opened 2019-11-21
Mounting non-AWS bucket?


RalfJung:
We are hosting our own S3-compatible storage and I would like to use s3fs with that. Unfortunately, s3fs insists on connecting to Amazon, and I was unable to find any hint in the documentation how one would specify a different provider. I tried `-o endpoint=OUR_SERVER` to no avail, and I also tried using something like `https://OUR_SERVER/BUCKET_NAME`, which did not work either.

I am sure there is something very simple I am doing wrong here, but alas, I can't figure out what, and the docs I found so far are not helping either.

#### Version of s3fs being used (s3fs --version)
_Amazon Simple Storage Service File System V1.80(commit:unknown) with GnuTLS(gcrypt)_

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
_2.9.7-1+deb9u2_

#### Kernel information (uname -r)
_4.9.0-8-amd64_

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
_Debian GNU/Linux 9 (stretch)_

#### s3fs command line used, if applicable
```
s3fs https://OUR_SERVER/fp-science-ci s3 -d -f
```
#### s3fs output
```
[CRT] s3fs.cpp:set_s3fs_log_level(253): change debug level from [CRT] to [INF] 
[INF]     s3fs.cpp:set_moutpoint_attribute(4151): PROC(uid=0, gid=0) - MountPoint(uid=0, gid=0, mode=40755)
[CRT] s3fs.cpp:s3fs_init(3334): init v1.80(commit:unknown) with GnuTLS(gcrypt)
[INF] s3fs.cpp:s3fs_check_service(3708): check services.
[INF]       curl.cpp:CheckBucket(2860): check a bucket.
[INF]       curl.cpp:prepare_url(4175): URL is http://s3.amazonaws.com/https//OUR_SERVER/fp-science-ci/
[INF]       curl.cpp:prepare_url(4207): URL changed is http://https.s3.amazonaws.com//OUR_SERVER/fp-science-ci/
[INF]       curl.cpp:insertV4Headers(2237): computing signature [GET] [//OUR_SERVER/fp-science-ci/] [] []

```
This is clearly nonsense, it is still using Amazon AWS.

RalfJung:
[...]

PR #1236, opened 2020-02-02
Document host and servicepath

gaul:
The various bits of `host`, `url`, and `servicepath` seem to overlap.
References #1203.


gaul:
Tests seem flaky on macOS?

ggtakec:
@gaul I also noticed, a temporary workaround is to flush the travis cache.
I will investigate to fix it correctly.

Issue #1217, opened 2019-12-18
Everything Uploaded To S3 Is Content-Type application/octet-stream


mydigia:
## My environment info:

$ s3fs --version
```
Amazon Simple Storage Service File System V1.85(commit:unknown) with OpenSSL
Copyright (C) 2010 Randy Rizun <rrizun@gmail.com>
License GPL2: GNU GPL version 2 <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
$ pkg-config --modversion fuse
`2.9.2`

$ rpm -qi fuse, dpkg -s fuse
```
package fuse, is not installed
package dpkg is not installed
Name        : fuse
Version     : 2.9.2
Release     : 11.el7
Architecture: x86_64
Install Date: Mon 16 Dec 2019 15:51:58 UTC
Group       : System Environment/Base
Size        : 223297
License     : GPL+
Signature   : RSA/SHA256, Mon 12 Nov 2018 14:25:34 UTC, Key ID 24c6a8a7f4a80eb5
Source RPM  : fuse-2.9.2-11.el7.src.rpm
Build Date  : Tue 30 Oct 2018 21:32:35 UTC
Build Host  : x86-01.bsys.centos.org
Relocations : (not relocatable)
Packager    : CentOS BuildSystem <http://bugs.centos.org>
Vendor      : CentOS
URL         : https://github.com/libfuse/libfuse
Summary     : File System in Userspace (FUSE) utilities
Description :
With FUSE it is possible to implement a fully functional filesystem in a
userspace program. This package contains the FUSE userspace tools to
mount a FUSE filesystem.
normal        /etc/fuse.conf
normal        /usr/bin/fusermount
normal        /usr/bin/ulockmgr_server
normal        /usr/sbin/mount.fuse
normal        /usr/share/doc/fuse-2.9.2
normal        /usr/share/doc/fuse-2.9.2/AUTHORS
normal        /usr/share/doc/fuse-2.9.2/COPYING
normal        /usr/share/doc/fuse-2.9.2/ChangeLog
normal        /usr/share/doc/fuse-2.9.2/FAQ
normal        /usr/share/doc/fuse-2.9.2/Filesystems
normal        /usr/share/doc/fuse-2.9.2/NEWS
[...]

PR #1237, opened 2020-02-02
Test that s3fs automatically detects Content-Type

gaul:
References #1217.


ggtakec:
@gaul I will modify the MacOS test (I plan to bypass some of it), but some parts of this PR did not pass the test.
In case of MacOS, it seems the Content-Type of test.txt is application/octet-stream. (mime.types file may be different.)

Issue #1241, opened 2020-02-04
Disk usage increases sharply with MacOS test


ggtakec:
### Version of s3fs being used (s3fs --version)
924eeb3

### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
n/a

### Kernel information (uname -r)
Darwin(MacOS)

### Details about issue
A test failed in MacOS testing(current master branch)

When the ensure_diskfree option is specified, the cache disk size that s3fs can use is fixed.
On MacOS, 1GB of disk space may be used by other processes or systems during the test.
So that,  as test_clean_up_cache fail because they check for Disk Available before and after the test.
Other tests may also generate an error indicating that the disk is full.

I guess the cause is related to Timemachine, but I haven't verified it.
The error sometimes continued even if it was disabled by the tmutil command.

For this reason, testing of the ensure_diskfree option on MacOS will be temporarily bypassed.

This issue will be investigated continuously.

ggtakec:
#1239 is included in this Issue.

PR #1242, opened 2020-02-04
Avoid test errors on MacOS

ggtakec:
#1151 #1237

### Details
Since a test error has occurred on MacOS, we have responded.
#### ensure_diskfree test
Testing the code merged in #1151 will fail on MacOS.  
Please refer to Issue #1241 for details.
#### content-types
In content-types test, Mime.type seemed to be different on MacOS, resulting in an error.  
See also #1237.
#### Test time
Ensure_diskfree test is bypassed because ensure_diskfree test takes too much time.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -749,7 +749,7 @@ function test_content_type() {
 }
 
 function add_all_tests {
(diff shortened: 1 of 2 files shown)


ggtakec:
@gaul Sorry for no-review, because we need to fix MacOS test problem, then merge quickly.
Please see if there are any problems.

PR #1244, opened 2020-02-08
use correct content-type when complete multipart upload

liuyongqing:
Content-Type means：‘The Media type of the body of the request (used with POST and PUT requests).', so we should change complete multipart  content-type to 'application/xml'.


gaul:
Agree that your change is correct for CompleteMultipartUpload.  Note that CreateMultipartUpload stores the actual object Content-Type.  Does this pull request fix a compatibility issue or is it just good hygiene?

PR #1253, opened 2020-03-18
Generic compilation instructions: explain that /etc/mime.type is needed

juliogonzalez:

#1217

### Details

I would say having `/etc/mime.types` is a runtime requirement.

Without it s3fs-fuse does not fail but uploads all files as "application/octect-stream".

NOTE1: I don't know how this works at MacOS as I don't have a system to test it. Is it `/etc/mime.types` as well?
NOTE2: If this PR is accepted, we should probably check for `/etc/mime.types` at `configure.ac` and break the build if it's not available.

### DIFF
--- a/COMPILATION.md
+++ b/COMPILATION.md
@@ -15,6 +15,7 @@ Keep in mind using the pre-built packages when available.
 * libcurl
 * libxml2
 * openssl
+* /etc/mime.types (the package providing depends on the OS)
 * pkg-config (or your OS equivalent)
 
 2. Then compile from master via the following commands:


ggtakec:
@juliogonzalez I'm always grateful for your help.

Currently, the test is bypassed in case of macOS by merging #1242.
This was my fault.

The reason that the txt file is determined to be application/octet-stream is that the /etc/mime.types file does not exist on macOS.
But macOS 10.15 Cataline comes with Apache2.4 pre-installed by default, so we may be able to use /etc/apache2/mime.types.
Or, I think that /usr/share/cups/mime.types of CUPS exists.
In any case, the problem is that s3fs looks only /etc/mime.types.

I will modify this mime.types file to be optionally passed to s3fs after merging your PRs.
(If not specified this option, it will use /etc/mime.types as default)
I will make s3fs that will exit with an error if the mime.types specified by the option or the default file does not exist (because s3fs operation cannot be guaranteed on this case).

Regarding **NOTE2**, the s3fs source code does not include the mime.types file for building, and s3fs refers to the mime.types file only at run time.
so I don't think it will cause a build error, and we should put a runtime error as described above.

I'll merge this PR soon.

PR #1254, opened 2020-03-19
Added mime option for strict checking of mime types file

ggtakec:
#1253 #1217 

### Background
The issue #1217 is that when s3fs cannot load the /etc/mime.types file, it uploads the Content-Type of all objects as application/octet-stream.
As pointed out in PR #1253, s3fs should not be able to build or start up.
This allows the user to notice that there is a problem with mime.types.

#### Added mime option
Added this option so that the user can specify the mime.types file path instead of /etc/mime.types(default).
I considered that the mime.types file path may be different depending on the environment by each user.
On non-macOS, the default mime.types path is /etc/mime.types. (this path is used when this option is not specified)
On macOS after Cataline, apache2.4 is pre-installed , so /etc/apache2/mime.types is the default for macOS.
This option gives user's environment flexibility.

### Exit if mime.types loading fails
s3fs exits with an error if the mime.types file(both default and user specified) fails to load.
This allows the user to know that there is problem with mime.types path.
Also, if s3fs can be started, mime.types can be loaded correctly, so it will solve #1217.

### Undo test evasion
I restored the content-types test for macOS that was avoided in #1237.
The error was occurred when testing code #1242 because /etc/mime.types did not exist on macOS.
My response and fixing to #1237 was temporary and wrong.
Therefore, I restored the original test to confirm that content-types can be set correctly on macOS.

### DIFF
--- a/COMPILATION.md
+++ b/COMPILATION.md
@@ -15,7 +15,11 @@ Keep in mind using the pre-built packages when available.
 * libcurl
 * libxml2
 * openssl
-* /etc/mime.types (the package providing depends on the OS)
+* mime.types (the package providing depends on the OS)
+	* s3fs tries to detect `/etc/mime.types` as default regardless of the OS
+	* Else s3fs tries to detect `/etc/apache2/mime.types` if OS is macOS
+	* s3fs exits with an error if these files are not exist
+	* Alternatively, you can set mime.types file path with `mime` option without detecting these default files
 * pkg-config (or your OS equivalent)
 
 2. Then compile from master via the following commands:
--- a/doc/man/s3fs.1
+++ b/doc/man/s3fs.1
@@ -326,6 +326,11 @@ Otherwise an error is returned.
 \fB\-o\fR requester_pays (default is disable)
 This option instructs s3fs to enable requests involving Requester Pays buckets (It includes the 'x-amz-request-payer=requester' entry in the request header).
 .TP
+\fB\-o\fR mime (default is "/etc/mime.types")
+Specify the path of the mime.types file.
+If this option is not specified, the existence of "/etc/mime.types" is checked, and that file is loaded as mime information.
+If this file does not exist on macOS, then "/etc/apache2/mime.types" is checked as well.
+.TP
 \fB\-o\fR dbglevel (default="crit")
 Set the debug message level. set value as crit (critical), err (error), warn (warning), info (information) to debug level. default debug level is critical.
 If s3fs run with "-d" option, the debug level is set information.
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -343,6 +343,20 @@ static const long S3FSCURL_RESPONSECODE_NOTSET      = -1;
 static const long S3FSCURL_RESPONSECODE_FATAL_ERROR = -2;
 static const int  S3FSCURL_PERFORM_RESULT_NOTSET    = 1;
 
+// [NOTE] about default mime.types file
+// If no mime.types file is specified in the mime option, s3fs
+// will look for /etc/mime.types on all operating systems and
+// load mime information.
+// However, in the case of macOS, when this file does not exist,
+// it tries to detect the /etc/apache2/mime.types file.
+// The reason for this is that apache2 is preinstalled on macOS,
+// and the mime.types file is expected to exist in this path.
+// If the mime.types file is not found, s3fs will exit with an
+// error.
+//
+static const char* DEFAULT_MIME_FILE        = "/etc/mime.types";
+static const char* SPECIAL_DARWIN_MIME_FILE = "/etc/apache2/mime.types";
+
 // [NOTICE]
 // This symbol is for libcurl under 7.23.0
 #ifndef CURLSHE_NOT_BUILT_IN
@@ -397,7 +411,7 @@ bool             S3fsCurl::requester_pays      = false;          // default
 //-------------------------------------------------------------------
 // Class methods for S3fsCurl
 //-------------------------------------------------------------------
-bool S3fsCurl::InitS3fsCurl(const char* MimeFile)
+bool S3fsCurl::InitS3fsCurl()
 {
   pthread_mutexattr_t attr;
   pthread_mutexattr_init(&attr);
@@ -411,9 +425,6 @@ bool S3fsCurl::InitS3fsCurl(const char* MimeFile)
     return false;
   }
   if(0 != pthread_mutex_init(&S3fsCurl::curl_share_lock[SHARE_MUTEX_SSL_SESSION], &attr)){
-    return false;
-  }
-  if(!S3fsCurl::InitMimeType(MimeFile)){
     return false;
   }
   if(!S3fsCurl::InitGlobalCurl()){
@@ -618,15 +629,39 @@ int S3fsCurl::CurlProgress(void *clientp, double dltotal, double dlnow, double u
   return 0;
 }
 
-bool S3fsCurl::InitMimeType(const char* MimeFile)
+bool S3fsCurl::InitMimeType(const std::string& strFile)
 {
-  if(!MimeFile){
-    MimeFile = "/etc/mime.types";  // default
+  string MimeFile;
+  if(!strFile.empty()){
+    MimeFile = strFile;
+  }else{
+    // search default mime.types
+    string errPaths = DEFAULT_MIME_FILE;
+    struct stat st;
+    if(0 == stat(DEFAULT_MIME_FILE, &st)){
+      MimeFile = DEFAULT_MIME_FILE;
+    }else if(compare_sysname("Darwin")){
+      // for macos, search another default file.
+      if(0 == stat(SPECIAL_DARWIN_MIME_FILE, &st)){
+        MimeFile = SPECIAL_DARWIN_MIME_FILE;
+      }else{
+        errPaths += " and ";
+        errPaths += SPECIAL_DARWIN_MIME_FILE;
+      }
+    }
+    if(MimeFile.empty()){
+      S3FS_PRN_ERR("Could not find miime.types files, you have to create file(%s) or specify mime option for existing mime.types file.", errPaths.c_str());
+      return false;
+    }
   }
+  S3FS_PRN_DBG("Try to load mime types from %s file.", MimeFile.c_str());
 
   string line;
-  ifstream MT(MimeFile);
+  ifstream MT(MimeFile.c_str());
   if(MT.good()){
+    S3FS_PRN_DBG("The old mime types are cleared to load new mime types.");
+    S3fsCurl::mimeTypes.clear();
+
     while(getline(MT, line)){
       if(line[0]=='#'){
         continue;
@@ -647,6 +682,10 @@ bool S3fsCurl::InitMimeType(const char* MimeFile)
         S3fsCurl::mimeTypes[ext] = mimeType;
       }
     }
+    S3FS_PRN_INIT_INFO("Loaded mime information from %s", MimeFile.c_str());
+  }else{
+    S3FS_PRN_ERR("Could not load mime types from %s, please check the existence and permissions of this file.", MimeFile.c_str());
+    return false;
   }
   return true;
 }
--- a/src/curl.h
+++ b/src/curl.h
@@ -355,7 +355,6 @@ class S3fsCurl
     static bool DestroyCryptMutex(void);
     static int CurlProgress(void *clientp, double dltotal, double dlnow, double ultotal, double ulnow);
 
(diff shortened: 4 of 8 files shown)


ggtakec:
@juliogonzalez You are correct, I think we should add comments about mime.types in `COMPILATION.md`.

After this build, I will try to make additional corrections. (And squash when merging)
If you could, please check it after I change it.

It seems that there is no description of macOS in Installation Notes, so I try to add it.
The environment on travis is one of example for macOS, so I refer to that summarize the procedure.

COMPILATION.md
```diff
@@ -15,7 +15,10 @@ Keep in mind using the pre-built packages when available.
 * libcurl
 * libxml2
 * openssl
-* /etc/mime.types (the package providing depends on the OS)
+* mime.types (the package providing depends on the OS)
+	* /etc/mime.types is default for not macOS
+	* /etc/apache2/mime.types is default for macOS
+	* If there is a file in a path other than the default, specify it with the mime option.
```
juliogonzalez: Shouldn't it be
>	* If there is a file in a path other than the default, specify it with the `--mime` option.

?

COMPILATION.md
```diff
@@ -15,7 +15,10 @@ Keep in mind using the pre-built packages when available.
 * libcurl
 * libxml2
 * openssl
-* /etc/mime.types (the package providing depends on the OS)
+* mime.types (the package providing depends on the OS)
+	* /etc/mime.types is default for not macOS
+	* /etc/apache2/mime.types is default for macOS
```
juliogonzalez: This doesn't seem to match what you say on the PR description:
> On non-macOS, the default mime.types path is /etc/mime.types. (this path is used when this option is not specified)
> On macOS after Cataline, apache2.4 is pre-installed , so /etc/apache2/mime.types is the default for macOS.

So if this is true, I suggest:

> 	* /etc/mime.types is default GNU/Linux, BSD and macOS until Catalina (10.15)
> 	* /etc/apache2/mime.types is default for macOS Catalina (10.15) and newer

COMPILATION.md
juliogonzalez: Well right, not `--mime` but just `mime` :-)

ggtakec: Is below description better than it?
- Alternatively, the mime.types file path can be specified with the `mime` option.

COMPILATION.md
ggtakec: I'll fix it because your description is better.
Thanks a lot.

COMPILATION.md
juliogonzalez: Yes, but on a second look, I'd use `/etc/mime.types`, `/etc/apache2/mime.types` and `mime.types` in all cases.

doc/man/s3fs.1
```diff
@@ -326,6 +326,10 @@ Otherwise an error is returned.
 \fB\-o\fR requester_pays (default is disable)
 This option instructs s3fs to enable requests involving Requester Pays buckets (It includes the 'x-amz-request-payer=requester' entry in the request header).
 .TP
+\fB\-o\fR mime (default is "/etc/mime.types")
+Specify the path of the mime.types file.
+If not specified, use the "/etc/mime.types" file ("/etc/apache2/mime.types" for macOS) as the default value.
```
juliogonzalez: Are you sure you want `/etc/apache2/mime.types` for macOS in all cases?

src/curl.cpp
```diff
@@ -343,6 +343,16 @@ static const long S3FSCURL_RESPONSECODE_NOTSET      = -1;
 static const long S3FSCURL_RESPONSECODE_FATAL_ERROR = -2;
 static const int  S3FSCURL_PERFORM_RESULT_NOTSET    = 1;
 
+// [NOTE]
+// Default path to mime.types file.
+// For macOS 10.15 Cataline, Apache2.4 is pre-installed.
+//
+#if defined(__APPLE__)
+static const std::string MIMETYPE_DEFAULT_PATH = "/etc/apache2/mime.types";
```
juliogonzalez: Same here.

doc/man/s3fs.1
ggtakec: I would like to discuss this point. That's because I'm still worried and still can't judge.
Details are described in the comments below.

ggtakec:
The current PR is that s3fs will try to use /etc/apache2/mime.types on all macOS if the mime option is not specified.
I don't think this is good, but I think there are three ways to determine the default path.
- (1) Search mime.types file with priorities (only for macOS)  
/etc/mime.types> /etc/apache2/mime.types
- (2) Search mime.types file by priority (all in all)  
/etc/mime.types> /etc/apache2/mime.types
- (3) Only /etc/apache2/mime.types (this is currently PR codes)

In the case of (1) and (2), I am concerned that there is a possibility of loading a file that the user does not intend.
That's why we chose method (3).
If case (3), s3fs will not start if user use Catalina or earlier. 
That is, the user will have to use the `mime` option.
Thus I expected it.
However, implementation of (1) and (2) is also possible easy.

@juliogonzalez @gaul 
So I would love to hear your opinion.

juliogonzalez:
What is strange to me is: why the new macOS catalina has `mime.types` at `/etc/apache2/`? Are we sure this is always the case? Can any macOS Catalina user confirm?

If that is correct, then I'd go with option 2. Can't tell you for BSD (can start at least a FreeBSD VM I guess), but on Debian, Ubuntu, CentOS, openSUSE and SLE, I can tell you the file is by default at `/etc/mime.types`, either installed with the system (openSUSE/SLE), either installed by an optional package (`mailcap`/`mine-types`).

And in fact, it's the correct place FMPOV to look for the file system-wide. On GNU/Linux you also have `/etc/apache2/mime.types` but at it's just a symlink to `/etc/mime.types`

This is a SLE12 with Apache installed.

```bash
$ find /etc|grep 'mime.types'
/etc/apache2/mime.types
/etc/mime.types

$ ls -l /etc/apache2/mime.types
lrwxrwxrwx 1 root root 13 feb 25 15:09 /etc/apache2/mime.types -> ../mime.types
```

This is Debian 10 with Apache installed:
```bash
$ find /etc/|grep 'mime.types'
/etc/mime.types

$ grep -r '/etc/mime.types' /etc/apache2/
/etc/apache2/mods-available/mime.conf:  TypesConfig /etc/mime.types
/etc/apache2/mods-available/mime.conf:  # es is ecmascript in /etc/mime.types
/etc/apache2/mods-available/mime.conf:  # tr is troff in /etc/mime.types
```

Looks strange that Apache is installed by default on macOS, but I admit last time I had contact with macOS was maybe 12 years ago. Maybe now it's installed to show local documentation or who-knows-what. But even so, are we sure it's not a link to `/etc/mime.types`?

ggtakec:
@juliogonzalez Thanks for your help.
I comfirmed /etc/apache2/mime.types with macOS 10.15 (Catalina) at my macbook.
(I have not installed apache2 on this macOS with homebrew etc.)

To confirm the preinstall status of apcahe2 on each macOS, I tested using TravisCI.
(NOTE: I believe the osx image provided by TravisCI (xcode is installed) is the same as the user's environment, but there is no certainty that it is correct)

As a result, the presence of /etc/apache2/mime.types in all environments listed in [macOS Version](https://docs.travis-ci.com/user/reference/osx/#macos-version) was confirmed.

In other words, this file seems to exist from macOS 10.10 to 10.15.
(This file is linked to actual file "/private/etc/apache2/mime.types")

If you have installed apache2 with Homebrew on macOS, there will be files under /usr/share, too.
But in that case, the user will be able to use the `mime` option.

Small later, I will change the code of this PR to the following specification.

- Search path priority of mime.types
(1) If the `mime` option is specified, only check the specified file
(2) If the `mime` option is not specified, /etc/mime.types is checked at first.
(3) If the `mime` option is not specified, /etc/apache2/mime.types is checked next.

If the file is not found, s3fs will not start with an error.

Along with this, I will also re-change COMPILATION.md and man.

COMPILATION.md
ggtakec: Since the search path and priority were changed, the code was squashed again and the description was almost changed.
Please check the new wording.

COMPILATION.md
ggtakec: Same as above.

doc/man/s3fs.1
ggtakec: Same as COMPILATION.md.

src/curl.cpp
ggtakec: Here, stop the judgment in APPLE, and change it so that it is detected in order of /etc/apach2/mime.typs from /etc/mime.types.

ggtakec:
The code was modified.
- the default mime.types file search with the priority from /etc/mime.typs to /etc/apache2/mime.types.
- And the `mime` option has the same specifications as when PR was made.
-The man page, COMPILATION.md, etc. have been modified accordingly.

@juliogonzalez @gaul 
Please check it.

src/curl.cpp
```diff
@@ -343,6 +343,23 @@ static const long S3FSCURL_RESPONSECODE_NOTSET      = -1;
...
 static const int  S3FSCURL_PERFORM_RESULT_NOTSET    = 1;
 
+// [NOTE]
+// This is the search path array for the mime.types file, ordered
+// by priority.
+// On most operating systems(Debian, Ubuntu, CentOS, openSUSE and
+// etc.), /etc/mime.types is expected. However, this file is almost
+// absent on macOS, but /etc/apache2/mime.types exists instead of
+// it. So, when s3fs loads the mime.types file, it tries to find
+// mime file in the order of this array.
```
juliogonzalez: I wonder if there could be some corner cases on GNU/Linux where `/etc/apache2/mime.types` exists, but `/etc/mime.types` does not.

If such situation is possible, then this would be wrong, as s3fs-fuse would be depending on Apache instead of depending on the system package that provides `/etc/mime.types`. So if at some point Apache is removed, s3fs-fuse will stop working.

Personally, I'd look for for `/etc/mime.types for all OS, and if it doesn't exist only for `/etc/apache2/mime.types` on macOS.

ggtakec: @juliogonzalez Thanks for your help.
I could understand your concern. (please point out if wrong)
Certainly, It is not good that s3fs depends on apache.
I'll try to change that it only treat macOS as special(apache preinstalled OS).

The final specification will be changed as follows:
For macOS, s3fs try to detect /etc/apache2/mime.types first, and if it doesn't exist, look for /etc/mime.types.
For non-macOS, it try to detect only /etc/mime.types.
If any file cannot be found, both s3fs will fail to start.
However, if user specify the `mime` option, s3fs will try to detect only the file specified by that option.

I try making the code changes again, please wait a few days.

juliogonzalez: > For macOS, s3fs try to detect /etc/apache2/mime.types first, and if it doesn't exist, look for /etc/mime.types.
> For non-macOS, it try to detect only /etc/mime.types.

Wouldn't it be easier:

1. s3fs try to detect `/etc/mime.types` regardless of the OS
2. Else s3fs try to detect `/etc/apache2/mime.types` if OS is macOS
3. Else exit with an en error.

?

Benefit: you treat `/etc/apache2/mime.types` as an exception at macOS, and only if the standard file doesn't exist. If the standard file exists on past or future versions, then it will be used.

ggtakec: Thanks, I will up codes with squashing soon.

ggtakec:
@juliogonzalez @gaul 
I just uploaded new modified code.
Please review if you like.

COMPILATION.md
```diff
@@ -15,7 +15,9 @@ Keep in mind using the pre-built packages when available.
 * libcurl
 * libxml2
 * openssl
-* /etc/mime.types (the package providing depends on the OS)
+* mime.types (the package providing depends on the OS)
+	* As default, the file is searched in the order /etc/mime.types to /etc/apache2/mime.types, and first detected file is loaded as mime information.
```
juliogonzalez: It seems this needs an update:
> A `mime.types` file (the package providing it depends on the OS)
> * As default, `/etc/mime.types` is checked (ff this file does not exist on macOS, then `/etc/apache2/mime.types` is checked as well)
> * Alternatively, Set `mime.types` file path with `mime` option, it is loaded instead of default search files.

ggtakec: I'm sorry that I forgot to modify this file.
I updated this file.

juliogonzalez: I'd recommend highlighting filenames and paths as I did at the quote above.

ggtakec: Thanks for review.
I set highlighting paths and updated file(with squash)

ggtakec:
@gaul @juliogonzalez I'll merge this PR. If you are still worried, please point out.
Thanks,

PR #1265, opened 2020-04-11
Fixed a bug of stats cache compression

ggtakec:
n/a

### Details
There was a bug in the stats cache compression process.
Due to this bug, stats that were not scheduled to be deleted may have been removed from the cache, which may have affected performance.
This bug has been fixed by this PR.

Thanks to 天空 for reporting this bug.

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -543,20 +543,23 @@ bool StatCache::TruncateCache()
   // 3) erase from the old cache in order
   size_t            erase_count= stat_cache.size() - CacheSize + 1;
   statiterlist_t    erase_iters;
-  for(stat_cache_t::iterator iter = stat_cache.begin(); iter != stat_cache.end(); ++iter){
+  for(stat_cache_t::iterator iter = stat_cache.begin(); iter != stat_cache.end() && 0 < erase_count; ++iter){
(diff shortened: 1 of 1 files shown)


src/cache.cpp
```diff
@@ -543,19 +543,20 @@ bool StatCache::TruncateCache()
...
+    }else{
+      // iter is not have notruncate flag
+      erase_iters.push_back(iter);
+      sort(erase_iters.begin(), erase_iters.end(), sort_statiterlist());
     }
-    // iter is not have notruncate flag
-    erase_iters.push_back(iter);
-    sort(erase_iters.begin(), erase_iters.end(), sort_statiterlist());
-    if(erase_count < erase_iters.size()){
+    while(erase_count < erase_iters.size()){
```
gaul, line 559: It would be more efficient to only sort the array once, e.g.,

```c++
if(erase_count < erase_iters.size()){
    sort(erase_iters.begin(), erase_iters.end(), sort_statiterlist());
    while(erase_count < erase_iters.size()){
        erase_iters.pop_back();
    }
}
```

ggtakec: You are correct, I'll fix it soon.

PR #1266, opened 2020-04-11
Test cache eviction

gaul:


gaul:
I worry that this PR make increase test run-time too much.  Maybe we should combine some of the flags?

ggtakec:
@gaul I think that the elapsed time of the test is likely to increase slightly, but it should not be a problem.
This is a good test for cache in and out.
However, it is unlikely that a test script will be able to determine if the cashout was successful.
That is because even if the stats are cached out, they are immediately re-acquired.
But I agree this PR.
Thanks.

PR #1267, opened 2020-04-12
Remove truncate brew package

gaul:
coreutils now includes this functionality.  Addresses Travis errors of
the form:

> Error: Cannot install truncate because conflicting formulae are installed.
>  coreutils: because both install `truncate` binaries
>
>Please `brew unlink coreutils` before continuing.


gaul:
This seems more complicated.  Continuing in #1268.

PR #1268, opened 2020-04-12
Upgrade Travis macOS builder

gaul:
Addresses warnings of the form:

> Warning: You are using macOS 10.12.
>
> We (and Apple) do not provide support for this old version.
>
> You will encounter build failures with some formulae.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,7 +79,6 @@ matrix:
           else
             exit 1;
           fi
-        - sudo ln -s /usr/local/opt/coreutils/bin/gstdbuf /usr/local/bin/stdbuf
       script:
         - ./autogen.sh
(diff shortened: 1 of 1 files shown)


gaul:
Apparently this cannot work due to travis-ci/travis-ci#10017.  Earlier I was seeing some strange errors related to truncate and homebrew.  Hopefully they were transient.  Even though we cannot upgrade to macOS 10.13 this PR did surface a few cleanups: 48a872e2853852e66651f25afb648c4ed4fda71d, a44fc1103d4b8a161d3c686224500626046c822d, and feafb44bae005f4fe2df94a0c8fe66f9adbdc4d9.

gaul:
It seems like all the Homebrew packages compile from source now, slowing run-time, perhaps because macOS 10.12 is no longer supported:

https://brew.sh/2019/11/27/homebrew-2.2.0/

gaul:
I don't think we can move forward here until Travis improves its support.

Issue #1245, opened 2020-02-14
Increased “Transport endpoint is not connected” rate with 1.86


acdha:
### Additional Information

I've noticed what appears to be an increased rate of “transport endpoint is not connected” errors with 1.86, although that could be a coincidence due to load — this has been a long-running problem which I was hoping 1.86 would improve. 

#### Version of s3fs being used (s3fs --version)

Amazon Simple Storage Service File System V1.86 (commit:unknown) with OpenSSL

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)

2.9.7-1ubuntu1

#### Kernel information (uname -r)

4.15.0-1058-aws

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)

```
NAME="Ubuntu"
VERSION="18.04.4 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.4 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic
```

#### /etc/fstab entry, if applicable

```
s3fs#bucket /srv/bucket fuse _netdev,umask=0022,uid=5000,gid=5000,noexec,nosuid,nodev,noatime,allow_other,nonempty,ensure_diskfree=1024,retries=8,max_stat_cache_size=1500000,multireq_max=500,parallel_count=40,dbglevel=warning,use_cache=/var/cache/bucket 0 0
```

#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)

<details>

```
Feb 14 14:01:04 ip-172-31-37-246 s3fs[1903]: curl.cpp:RequestPerform(2459): HTTP response code 500, returning EIO. Body Text:
[...]

Issue #1191, opened 2019-11-05
s3fs crash during cache read


Regressor:
#### Version of s3fs being used (s3fs --version)
Amazon Simple Storage Service File System V1.85 (commit:0e815c2) with OpenSSL

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
Version: 2.9.4-1ubuntu3.1

#### Kernel information (uname -r)
4.4.0-1092-aws

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
Description:    Ubuntu 16.04.4 LTS

#### s3fs command line used, if applicable
s3fs media.xxxxxxxx.com /opt/media -o rw,allow_other,nonempty,use_cache=/cache,multireq_max=100,ensure_diskfree=10000,use_path_request_style,endpoint=eu-west-1,url=http://s3-eu-west-1.amazonaws.com,dev,suid

#### /etc/fstab entry, if applicable
s3fs#media.xxxxxxxx.com /opt/media fuse _netdev,allow_other,nonempty,use_cache=/cache,multireq_max=100,ensure_diskfree=10000,use_path_request_style,endpoint=eu-west-1,url=http://s3-eu-west-1.amazonaws.com 0 0

#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
Nov 02 22:41:09 ip-xxx-xx-xx-x sudo[7998]:     root : TTY=pts/3 ; PWD=/var/crash ; USER=root ; COMMAND=/usr/bin/apport-unpack /var/crash/_usr_bin_s3fs.0.crash s3fscrash
Nov 02 22:45:05 ip-xx-xx-xx-x sudo[8579]:     root : TTY=pts/3 ; PWD=/var/crash ; USER=root ; COMMAND=/usr/bin/apport-unpack /var/crash/_usr_bin_s3fs.0.crash s3fscrash

### Details about issue
Hi. We have a problem with s3fs. Our webserver distribute media files mounted from s3 bucket. We use caching and sometimes s3fs process crashed and no files can be read or write. We have script checking s3 mount availability that do remounts when problem detected.

When crash happens apport create report. Here is the backtrace:

#0  0x00007fab676b7428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007fab676b902a in __GI_abort () at abort.c:89
[...]

PR #1271, opened 2020-04-14
Fixed the truncation bug of stat file for cache file

ggtakec:
#1245 #1191 

#### Truncate cache page information file(stat)
When serializing the page information file(stat) of the cache file, the file was not truncated.
#### Wrap s3fs_strtoofft for PageList 
s3fs_strtoofft function was throwing an exception, but PageList::Serialize method did not catch it.
Therefore, an intermediate wrap_strtoofft function was added that catches this exception and is called from PageList::Serialize to wrap the s3fs_strtoofft function.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -63,6 +63,33 @@ static const int MAX_MULTIPART_CNT = 10 * 1000;  // S3 multipart max count
 #endif
 
 //------------------------------------------------
+// Local utility function
+//------------------------------------------------
+// Wrapped s3fs_strtoofft()
+//
+// PageList::Serialize method uses s3fs_strtoofft function,
+// but s3fs_strtoofft throws an exception, so wrap it in
+// this function.
+// PageList::Serialize works as 0 when the read data is invalid.
+//
+static off_t wrap_strtoofft(const char* str)
+{
+  off_t result;
+  if(str){
+    try{
+      result = s3fs_strtoofft(str);
+    }catch(std::exception &e){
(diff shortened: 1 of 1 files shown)


src/fdcache.cpp
```diff
@@ -843,6 +870,10 @@ bool PageList::Serialize(CacheFileStat& file, bool is_output)
     }
 
     string strall = ssall.str();
+    if(-1 == ftruncate(file.GetFd(), strall.length())){
```
gaul, line 873: Would it be safer to truncate to zero?  This would ensure that if the subsequent write (partially?) fails that there is not garbage in the file.

gaul: Actually it seems like we should probably write to a temp file and rename onto the existing cache file for atomicity reasons.  What do you think?

ggtakec: I understand, I will change to truncate the file size to 0.
And I will check the source code again for the operation linked with the cache file.
please wait a moment.

ggtakec: I modified the source code so that the size of truncate is 0.
I also comment outside of this conversation about this PageList file conflict.
Then, apply for the correction in another PR different from this PR.

ggtakec:
I will investigate the collision of the stats file(which is serialized file of PageList object) that holds the state of the cache file, and its timing. After that I will post new PR after this PR merge.
The important thing is to ensure that the cache and stats files are definitely paired.
In the current master code, there are some uncertainties, so I will throw a PR to dispel it.
Please wait for the next PR on this matter.

For the time being, I will merge this PR, if you have any problem, please let me know.

PR #1273, opened 2020-04-18
Add install instructions for Arch Linux

migueldvb:
### Details

Add install instructions for Arch Linux

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,12 @@ Many systems provide pre-built packages:
   sudo yum install s3fs-fuse
   ```
 
+* Arch Linux:
+
+  ```
+  sudo pacman -S s3fs-fuse
+  ```
+
 * Debian 9 and Ubuntu 16.04 or newer:
 
   ```


gaul:
Thank you for your contribution @migueldvb!

juliogonzalez:
Should we maybe add installation from source at https://github.com/s3fs-fuse/s3fs-fuse/wiki/Installation-Notes ?

It's useful if a user needs to test something from master, a branch, etc.

gaul:
Does https://github.com/s3fs-fuse/s3fs-fuse/wiki/Installation-Notes#from-github-repository cover this?  If not let's continue discussing doc improvements in a separate issue.  At least we can remove distributions that are end of life like Ubuntu 10.04 - 12.04.

juliogonzalez:
@gaul I was talking about https://github.com/s3fs-fuse/s3fs-fuse/wiki/Installation-Notes#notes-for-specific-operating-systems, as it's only there were the packages that are required to build and run are listed.

In fact https://github.com/s3fs-fuse/s3fs-fuse/wiki/Installation-Notes#general-instructions should be probably part of https://github.com/s3fs-fuse/s3fs-fuse/blob/master/COMPILATION.md

gaul:
@juliogonzalez Ah I understand, it would be nice to have Arch compilation instructions as well.  Agreed that this makes more sense as part of COMPILATION.md.  I also made some edits to remove the stale Ubuntu 10.04 and 12.04 instructions although we should also refresh 14.04 to 16.04 given that the former is EOL.

PR #1274, opened 2020-04-19
Improved strictness of cache file stats(file)

ggtakec:
#1271 

### Details
This PR code was modified to strictly manage the file(cache file stats) that holds the state of the s3fs cache file.

If the user has caching turned on, s3fs will load the cache file status(information on the range where the file contents are cached) from this file when opening it.
And if user edits the file, the cache will change and the cache file stats will change as well.
This cache file stats is written out when the file is closed(and flushed).
That is, the cache file stats(file) is loaded when user opens the file and writes when user closes it.

Access to the cache file stats(file) is exclusively controlled within the s3fs process.
Therefore, input/output to this file is performed simultaneously with the cache file with exclusive control.

However, s3fs must ensure that the cache file and its status file (cache file stats) are paired.
In this PR, it is modified to write the cache file's inode number to the cache file stats(file).
And the inode number is checked when loading this file.

When loading the cache file stats(file) output by s3fs before PR, the inode is not checked and the access to cache file is allowed.
However, when the cache status changes and the file(object) is closed, the cache file stats(file) is written out in the new format.
This maintains backward compatibility.
Be careful, an error will occur if the cache file stats(file) created by this PR code is loaded by old s3fs version.
In this case, that cache won't load, and s3fs will get the file(object) from S3 server.

Please note that multiple s3fs processes can not use the same cache directory at the same time.
The s3fs cache does not support multiple s3fs processes.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -853,7 +853,7 @@ bool PageList::ClearAllModified()
   return Compress();
 }
 
-bool PageList::Serialize(CacheFileStat& file, bool is_output)
+bool PageList::Serialize(CacheFileStat& file, bool is_output, ino_t inode)
 {
   if(!file.Open()){
     return false;
@@ -863,7 +863,7 @@ bool PageList::Serialize(CacheFileStat& file, bool is_output)
     // put to file
     //
     ostringstream ssall;
-    ssall << Size();
+    ssall << inode << ":" << Size();
 
     for(fdpage_list_t::iterator iter = pages.begin(); iter != pages.end(); ++iter){
       ssall << "\n" << iter->offset << ":" << iter->bytes << ":" << (iter->loaded ? "1" : "0") << ":" << (iter->modified ? "1" : "0");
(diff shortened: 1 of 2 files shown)


src/fdcache.cpp
```diff
@@ -908,13 +908,46 @@ bool PageList::Serialize(CacheFileStat& file, bool is_output)
...
+        if(0 == cache_inode){
+          S3FS_PRN_ERR("wrong inode number in parsed cache stats.");
+          delete[] ptmp;
+          return false;
+        }
+      }
+    }
+    // check inode number
+    if(0 != cache_inode && cache_inode != inode){
+      S3FS_PRN_ERR("differ inode and inode number in parsed cache stats.");
```
gaul, line 947: I'm trying to understand what the strictness helps with -- do you worry about an external process modifying the cache file while s3fs is running?

ggtakec: Although Including editing by an external process, I am concerned about problems(cache inconsistency) due to other factors, too.

The cache file is opened by s3fs when the file(object) is opened.
At that time, load the stats file of the cache file and s3fs knows the contents of the cache file(which area is loaded).
This may not be good, but s3fs uses two files to manage one piece of information.
Thus I believe these need to be consistent.
Therefore, the inode number was added to cache stats file for that it pointed to correct cache file.
If this inode number is not as same as the cache file's, s3fs should not load the cache file and its stats.

Issue #1220, opened 2019-12-20
EntityTooSmall in multipart upload


cnanakos:
`[ERR] curl.cpp:RequestPerform(2423): HTTP response code 400, returning EIO. Body Text: <Error><Code>EntityTooSmall</Code><Message>Your proposed upload is smaller than the minimum allowed size</Message><ProposedSize>528384</ProposedSize><$inSizeAllowed>5242880</MinSizeAllowed><PartNumber>1</PartNumber><ETag>480294879a56052a5937e78cb358580c</ETag><RequestId>AA1936A5569E43CF</RequestId><HostId>1pnskFIU9nckO2A1FqBhqCAV/92ZJaMbSixIEy+6HgS1BwyUN5PDhu4tLn81hxx229H+iSK2yzjN</Hos$Id></Error>`

The size of the file is 1.6GiB and the multipart_size is 16MiB. Everything else is default.

The part size is set as:

`-o multipart_size=16
`

Amazon Simple Storage Service File System V1.85 (commit:cc4a307) with OpenSSL
Copyright (C) 2010 Randy Rizun <rrizun@gmail.com>
License GPL2: GNU GPL version 2 <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

sqlbot:
Will you reformat that error as a code block, please, and verify that it's pasted correctly?  It appears to be corrupted or some of the XML tags are being interpreted or stripped. 

Also please show exactly how you're configuring the part size when mounting.

cnanakos:
Of course, updated the initial comment. Thanks.

cnanakos:
`[INF]       curl.cpp:insertV4Headers(2746): computing signature [POST] [/backups/test.vbk] [uploadId=fF_BMap7QeyO7dyTRX9k5iAw-LEVb3L3TqcihvagDwt2qh82EzsN324LuB15LpkU6xdCE4ORCGC6tM0fV3gOiKpBN
jx-MdlSSoqdDC7kUY9QClarHT7lmebFMdHBSroV] [7d75563ff4aed76905da416ef6911ef7a4217f80a7319bed33135f94f36a1ba5]
[INF]       curl.cpp:url_to_host(99): url is https://s3.xxxx.com
[...]

PR #1276, opened 2020-04-20
Do not create page when writing to an existing one

gaul:
This works around issues with random writes and the mixupload
optimization where an unloaded range could be created.  Also do not
retry requests on HTTP 416 since this is not a transient error.
References #1220.


gaul:
@ggtakec This PR allows the modified test to succeed but I still have doubts about `PageList::Parse`.  Unfortunately we do not have any unit tests so this is hard to debug.

gaul:
Superseded by #1277.

PR #1277, opened 2020-04-20
Fixed insufficient upload size for mix multipart upload

ggtakec:
#1220 #1276

### Details
This PR is a fix for the cause of #1220.

The bug of #1220 is that the size of one part does not meet the required size of multipart when uploading as mix multipart upload.
If the size of one part does not satisfy the required size of multipart, s3fs will download the data around that part and try to satisfy the required size.
And s3fs will upload with a part size that is larger than the requested size.
However, the code contained a bug, and although s3fs downloaded it, s3fs uploaded it without including that downloaded area.
This PR fixes this bug.

Although not included in this PR, the 416 HTTP response code will be corrected in the PR of #1276.
This PR and #1276 will be the correction of the defect of #1220.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -336,8 +336,10 @@ void PageList::Clear()
 bool PageList::Init(off_t size, bool is_loaded, bool is_modified)
 {
   Clear();
-  fdpage page(0, size, is_loaded, is_modified);
-  pages.push_back(page);
+  if(0 < size){
+    fdpage page(0, size, is_loaded, is_modified);
+    pages.push_back(page);
+  }
   return true;
 }
 
@@ -391,7 +393,7 @@ bool PageList::Parse(off_t new_pos)
       // nothing to do
       return true;
     }else if(iter->offset < new_pos && new_pos < iter->next()){
-      fdpage page(iter->offset, new_pos - iter->offset, iter->loaded, false);
+      fdpage page(iter->offset, new_pos - iter->offset, iter->loaded, iter->modified);
       iter->bytes -= (new_pos - iter->offset);
       iter->offset = new_pos;
       pages.insert(iter, page);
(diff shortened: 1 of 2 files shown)


ggtakec:
There was a bug(which was found by the test code #1276) yet. I will update the code soon.

src/fdcache.cpp
```diff
@@ -336,8 +336,12 @@ void PageList::Clear()
 bool PageList::Init(off_t size, bool is_loaded, bool is_modified)
 {
   Clear();
-  fdpage page(0, size, is_loaded, is_modified);
-  pages.push_back(page);
+  if(0 < size){
+    fdpage page(0, size, is_loaded, is_modified);
+    pages.push_back(page);
+  }
+  Dump();
```
gaul, line 343: Needed?

src/fdcache.cpp
```diff
@@ -1999,7 +2003,7 @@ int FdEntity::RowFlush(const char* tpath, bool force_sync)
           return -1;
         }
         for(fdpage_list_t::const_iterator iter = dlpages.begin(); iter != dlpages.end(); ++iter){
-          if(0 != (result = Load(iter->offset, iter->bytes, true))){
+          if(0 != (result = Load(iter->offset, iter->bytes, true, true))){  // set loaded and modified flag
```
gaul, line 2006: Maybe use a pseudo-keyword argument, e.g., `/*is_modified_flag=*/ true`?  clang-tidy checks these: https://clang.llvm.org/extra/clang-tidy/checks/bugprone-argument-comment.html

src/fdcache.cpp
ggtakec: Without this judgment, I was creating a useless fdpage object with offset 0 and size 0.
It doesn't lead to any trouble, but I think it is better not to create it.

gaul: Sorry I mean is the call to `Dump` needed?

src/fdcache.cpp
ggtakec: Thanks! I will add it and update soon.

ggtakec:
I try to make idea for a test that inspects the stat file of a cache file.
Maybe it works only when the use_cache option is specified and I think it would be a way to directly check the cache file and its stat file.

src/fdcache.cpp
gaul: The call to `Dump` only logs the list so I don't think it's necessary.

ggtakec: I'm sorry for my fault, this line is for my debugging.
It will be removed soon.

ggtakec: I fixed this, and rebase this PR from the current master.

PR #1280, opened 2020-04-22
Add test for cache file stat content

ggtakec:
#1277 

### Details
We needed to test the cache file and its stat file(fdcache.cpp), it was requested in #1277.
I added a test to check the contents of the cache file stat(file).
Notes: In this test, the size of the downloaded area may differ depending on the minimum upload unit, etc., so the total size of the cache file is checked.


ggtakec:
I merge this PR.
If you find a problem please let me know.

PR #1282, opened 2020-04-29
Warn about missing MIME types instead of exiting

### LINKED ISSUE #1270: Document mime types depenency
Someone on StackOverflow recently asked about the mime types dependency:

https://stackoverflow.com/questions/61207571/s3fs-could-not-load-mime-types-for-curl-library/61208047#61208047

Should this be a hard dependency?  Instead we could just warn about its absence since most users not require Content-Type.

gaul:
s3fs uses the MIME types file to set Content-Type for uploaded
objects.  Most distribution packages should install this via
recommended (not required) dependencies.  Users compiling from source
may not have this installed and s3fs should not prevent launching
since most users do not care about Content-Type.  Instead warn about
MIME types absence.  Fixes #1270.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -650,7 +650,7 @@ bool S3fsCurl::InitMimeType(const std::string& strFile)
       }
     }
     if(MimeFile.empty()){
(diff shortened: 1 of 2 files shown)


ggtakec:
@gaul I have accepted this fix.
Accept that if MIME fails to load, a warning will be printed and it will not prevent it from starting.

PR #1285, opened 2020-05-03
Not abort process by exception threw from s3fs_strtoofft

ggtakec:
#1256 #1191

### Details
Although the s3fs_strtoofft() function throws an exception, the process may abort because s3fs does not catch this exception.(ex. #1256 #1191 )
I prepared a wrap function that catches the exception and returns the result or the value(=default value 0).
The part that called each s3fs_strtoofft function is replaced with these wrap functions.
_The s3fs-fuse option analysis function may be left as it is. But it has been modified according to this change._

This fix also helps prevent unexpected aborts in exceptions, but rather to ease debugging difficulties when aborted.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -1375,13 +1375,17 @@ S3fsCurl* S3fsCurl::UploadMultipartPostRetryCallback(S3fsCurl* s3fscurl)
   string upload_id;
   string part_num_str;
   int    part_num;
+  off_t  tmp_part_num = 0;
   if(!get_keyword_value(s3fscurl->url, "uploadId", upload_id)){
     return NULL;
   }
   if(!get_keyword_value(s3fscurl->url, "partNumber", part_num_str)){
     return NULL;
   }
-  part_num = s3fs_strtoofft(part_num_str.c_str(), /*base=*/ 10);
+  if(!try_strtoofft(part_num_str.c_str(), tmp_part_num, /*base=*/ 10)){
+    return NULL;
+  }
+  part_num = static_cast<off_t>(tmp_part_num);
 
   if(s3fscurl->retry_count >= S3fsCurl::retries){
     S3FS_PRN_ERR("Over retry count(%d) limit(%s:%d).", s3fscurl->retry_count, s3fscurl->path.c_str(), part_num);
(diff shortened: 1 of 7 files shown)


jgmarce:
Is there issue tracking some corrupt file metadata (fsdata) possibly caused by poor thread management (multireq_max) that leads to the s3fs_strtoofft abort?   I'd like to join/contribute to that conversation as I have a repeatable issue in my environment.

ggtakec:
@jgmarce Please tell me the information of the bug you have.
If you can, I would like to know the result of checking the s3fs executable built with the current master.

I read some bug reports, then I opened this PR because s3fs was often aborting due to an exception from s3fs_strtoofft.
The only change to this code is the addition of a function to handle the s3fs_strtoofft exception.(It's just a fix to prevent s3fs from aborting.)
However, the basically cause of s3fs_strtoofft throwing an exception seems to be different for each issue.

If you have a log of bugs that occur due to the poor threads of multipart requests, it would be helpful if you can raise new Issue.

jgmarce:
I don't know that I'm in a position any longer to help. 

s3fs was not stable for my application.   I saw comments about
multireq_max and setting this to less than 5.  This change did seem to
improve stability.   I continued to make changes until I was running
with debug, in the foreground and with the single-threaded option.  
Still 100% availability was not achieved.   I found that for the
reported s3fs_strtoofft, I found that I could remove the offending file
from cache, but learned over time that some files repeated the failure. 
   At that point, if found "yas3fs" and replaced "s3fs". 

To look at the any deeper, I'd like to know if there is a checker "like
a filesystem check" that could examine and validate a s3fs cache area. 

Then, I'd like to know if someone is committed to fixing the instability
when multireq_max is above 1. 

There is nothing magical about my workload, just heavy read of large
files....     

I have the old cache, in place, still and if there were a way to examine
it for corruption, we could start there. 

-joe 

---

On 2020-05-14 10:30, Takeshi Nakatani wrote:

> @jgmarce [1] Please tell me the information of the bug you have.
> If you can, I would like to know the result of checking the s3fs executable built with the current master. 
> 
> I read some bug reports, then I opened this PR because s3fs was often aborting due to an exception from s3fs_strtoofft.
> The only change to this code is the addition of a function to handle the s3fs_strtoofft exception.(It's just a fix to prevent s3fs from aborting.)
> However, the basically cause of s3fs_strtoofft throwing an exception seems to be different for each issue. 
> 
> If you have a log of bugs that occur due to the poor threads of multipart requests, it would be helpful if you can raise new Issue. 
> 
> Thanks in advance for your help. 
> 
> --
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub [2], or unsubscribe [3].
 

Links:
------
[1] https://github.com/jgmarce
[2]
https://github.com/s3fs-fuse/s3fs-fuse/pull/1285#issuecomment-628674041
[3]

gaul:
@ggtakec I think it is reasonable to create some tool to check cache consistency; what do you think?

ggtakec:
@jgmarce
s3fs does not yet have a tool to check consistency about cache file and it's stat.
@gaul
I think it's better to prepare a checker as you say, to solve @jgmarce problem and to stabilize s3fs.

First, we try to set multireq_max to 1 or more, download (read) the object, and check whether the information in the cache file and its stat file is correct.
I have created new issue #1291 for this problem.

ggtakec:
It will continue to consider #1291.
Before that, I will merge this PR.

PR #1296, opened 2020-05-24
Import ossfs tests

gaul:
This turns up a failure in `test_rename_file` when calling `seek(0)` on a
renamed file.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -846,6 +846,12 @@ function test_cache_file_stat() {
     rm_test_file "${BIG_FILE}"
 }
 
+function test_ut_ossfs {
+    echo "Testing ossfs python ut..."
+    export TEST_BUCKET_MOUNT_POINT=$TEST_BUCKET_MOUNT_POINT_1
+    ../../ut_test.py
+}
+
 function add_all_tests {
     if `ps -ef | grep -v grep | grep s3fs | grep -q ensure_diskfree` && ! `uname | grep -q Darwin`; then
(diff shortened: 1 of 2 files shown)


gaul:
I'm not sure we should commit this as-is, but the test appears to surface a real issue.

gaul:
This test fails only with `-o use_cache` and `-o nocopyapi`.

gaul:
`-o usecache` fails with:

```
s3fs: [ERR] fdcache.cpp:RenamePath(1530): failed to link old cache path(/tmp/s3fs-cache/s3fs-integration-test/testrun-31
771/6ae07gg8fa) to new cache path(/tmp/s3fs-cache/s3fs-integration-test/testrun-31771/bbe9ec624f) by errno(17).
s3fs: [ERR] fdcache.cpp:Rename(2791): Failed to rename FdEntity object for /testrun-31771/6ae07gg8fa to /testrun-31771/b
be9ec624f
```

gaul:
These tests now pass.  The Python-based tests are interesting and allow fine-grained control over system calls but do not integrate well into the existing shell-script infrastructure.  We previously worked around this with helpers like `test/write_multiple_offsets.py` but maybe we should look at moving to a framework like this instead?

ggtakec:
If these shell script does not give enough testing, we probabry should use other tools such as python instead.
Do you know a good tool for disk/file accessing as a unit testing framework?
If we need more rigorous testing, I think we also can also prepare your code in C/C++ for testing.

PR #1299, opened 2020-05-24
Remove references to Amazon S3

### LINKED ISSUE #1298: Remove references to Amazon S3
s3fs works with a variety of S3 implementations and it should avoid specifying Amazon S3 when not necessary.

gaul:
s3fs works with a variety of S3 implementations and it should avoid
specifying Amazon S3 when not necessary.  Fixes #1298

### DIFF
--- a/test/integration-test-common.sh
+++ b/test/integration-test-common.sh
@@ -4,14 +4,14 @@
 # Common code for starting an s3fs-fuse mountpoint and an S3Proxy instance 
 # to run tests against S3Proxy locally.
 #
-# To run against an Amazon S3 or other S3 provider, specify the following 
+# To run against a hosted S3 provider, specify the following
 # environment variables:
 #
 # S3FS_CREDENTIALS_FILE=keyfile      s3fs format key file
 # S3FS_PROFILE=name                  s3fs profile to use (overrides key file)
 # TEST_BUCKET_1=bucketname           Name of bucket to use 
 # S3PROXY_BINARY=""                  Specify empty string to skip S3Proxy start
-# S3_URL="https://s3.amazonaws.com"  Specify Amazon AWS as the S3 provider
+# S3_URL="https://s3.amazonaws.com"  Specify the S3 provider
 #
 # Example of running against Amazon S3 using a bucket named "bucket:
 #
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 #
-# s3fs - FUSE-based file system backed by Amazon S3
+# s3fs - FUSE-based file system backed by S3
 #
 # Copyright(C) 2007 Randy Rizun <rrizun@gmail.com>
 #
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 #
-# s3fs - FUSE-based file system backed by Amazon S3
+# s3fs - FUSE-based file system backed by S3
 #
 # Copyright(C) 2007 Randy Rizun <rrizun@gmail.com>
 #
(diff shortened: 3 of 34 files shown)


test/integration-test-common.sh
```diff
@@ -4,14 +4,14 @@
 # Common code for starting an s3fs-fuse mountpoint and an S3Proxy instance 
 # to run tests against S3Proxy locally.
 #
-# To run against an Amazon S3 or other S3 provider, specify the following 
+# To run against a real S3 provider, specify the following
```
gaul, line 7: "real" in this case means not S3Proxy.  I will change this to "hosted".

test/integration-test-common.sh
```diff
@@ -4,14 +4,14 @@
...
 # environment variables:
 #
 # S3FS_CREDENTIALS_FILE=keyfile      s3fs format key file
 # S3FS_PROFILE=name                  s3fs profile to use (overrides key file)
 # TEST_BUCKET_1=bucketname           Name of bucket to use 
 # S3PROXY_BINARY=""                  Specify empty string to skip S3Proxy start
-# S3_URL="https://s3.amazonaws.com"  Specify Amazon AWS as the S3 provider
+# S3_URL="https://example.com"       Specify the S3 provider
```
gaul, line 14: I will revert the URL.

ggtakec:
@gaul 
I have a similar opinion to @sqlbot.
I assume s3fs is a basically tool that supports S3 (which is Amazon's distributed storage and its API).
And I think that it should be expressed that `s3fs can operate if it is a distributed storage that provides an API compatible with S3`.

In some cases, s3fs supports another distributed storage which does not provide a fully compatible API.
So what about how to provide a list of distributed storage on which s3fs worked?
(I already know the documentation for https://github.com/s3fs-fuse/s3fs-fuse/wiki/Non-Amazon-S3 that @gaul has prepared.)
Alternatively, how about having a list other than S3 that we confirmed the operation in README as an item?

Thus, I don't think it is necessary to delete the `Amazon` string of this PR.

ggtakec:
Isn't it possible to use `Amazon S3 and S3 compatible`(for example) without deleting` Amazon`?
I believe that the main support for s3fs(the primary target for proof and testing) is Amazon S3.
We try to add and support S3 compatible distributed storage(which has S3 API) as well.
I think this is derived support as a result of the main target. (Guaranteing operation with S3 API is the first goal of s3fs)

So you know, even so it's not fully compatible, s3fs will also prepare options to support them.
Thus I would want the list to be prepared so that the user won't be disappointed.

Then I think it is not necessary to delete Amazon from the comment and display, but how about expressing it as `Amazon S3 and S3 compatible` etc.(Other English is okay ...)

gaul:
Deleting Amazon is the point of this PR.  s3fs works across many different implementations and recommending using s3fs is difficult for those vendors if their largest competitor receives advertising through the program.  Personally, I use s3fs with Amazon but it's easy to see how other vendors say S3-compatible without mentioning Amazon.

gaul:
We lack consensus on this change.  I will close the PR for now and interested parties should follow up in the open issue.

ggtakec:
@gaul I thought about this request.

I can agree with your idea of ​​making s3fs a standard tool for `S3`.
However, I think that the main target of s3fs should be Amazon S3.
Of course, we should support another distributed storages which provide S3 I/F, and prepare the environment for testing.

This is the intention of that the README and other documentation have a list of distributed storage which s3fs can support.
I think that those information is needed to allow s3fs to be listed as a candidate when the user is looking for a FUSE tool that supports distributed storage which provides S3 I/F.

Based on this assumption, let's change `by Amazon S3` embedded in the header such as source code to `by S3`

If there is no discrepancy between your opinion and mine, I would like this PR to be revived.

PR #1303, opened 2020-05-31
Relink cache stats file atomically via rename

gaul:
The new file may already exist so link may fail.  Further `link`/`unlink`
is not atomic.  Addresses an error when renaming an open with with
`use_cache`.  References #1296.


gaul:
@ggtakec Please check this carefully -- I may not understand the cache file logic.

Issue #1291, opened 2020-05-18
Checking consistency of cache file and its stat


ggtakec:
#### Version of s3fs being used (s3fs --version)
v1.86

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
n/a

#### Kernel information (uname -r)
n/a

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
n/a

#### s3fs command line used, if applicable
Set multireq_max to 2 or more

#### /etc/fstab entry, if applicable
n/a

#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
n/a

### Details about issue
This is an issue derived from #1285 

Start s3fs with multireq_max set to 2 or more.
When a large object (file) is read (downloaded), the consistency between the cache file and the stat information of the cache file may not match.

We would like to prepare tools so that we can investigate the possibility and cause of this inconsistency.

gaul:
@ggtakec does this issue need further work?

ggtakec:
@gaul I'm trying to add debug-only code to this issue.
Checking cache files from outside s3fs cannot be exact, so I'm trying to prepare an option and write code that can only work when the option is enabled.
So please wait a while and I'll ask you to review them.

gaul:
@ggtakec Is this resolved?

qianyi-sourse:
@ggtakec Is this resolved? I met the similar problem. After I cp a large file to the mount dir , the cache file size and source file size did not match.

ggtakec:
The issue is likely due to forgetting to close(because of waiting for a bug report after merging the related PR).
A lot of code has changed since when this problem occurred, so it is unknown if it still occurs.
I think there have been reports of similar events, but I think they are always checked.
(In addition, since the cache file is a sparse file, it is necessary to compare it with the cache information file contents.)

Since this issue has changed a lot of code since I made, I think we should close this issue for now.

@qianyi-sourse 
Could you please post the phenomenon you are currently detecting as a new issue?
_If you write #1291 as the related issue in your issue, it will be easier to track down the issue._
At that time, it would be helpful if you could write detailed information together.
In addition, since the file size may be different, please let me know the options you are using and the free disk space.

Thanks in advance for your assistance.

PR #1312, opened 2020-06-19
Fixed a bug about serializing from cache file

ggtakec:
#1274 

### Details
There was a problem in the part that serialized from the stats information file of the cache file changed in PR of #1274.
It seems that the cache file was not functioning effectively from #1274 onwards to this PR.
This bug has been fixed by this PR.

In addition, I'm preparing the code corresponding to #1291, and it is planed to provide a function that can check the consistency of the cache file with that code.


ggtakec:
@gaul 
This bug is simple, but fatal and we believe it should be applied early.
The fix is ​​simple and does not affect other operations. So I will do a merge once the PR CI passes.
If you found any additional issues, please let me know.
Sorry for the urgent response.

Issue #1311, opened 2020-06-19
s3fs error 400 due to part size exceeding 5GB


xrefft:
To reproduce this problem，you just need:

```shell
touch tmp
truncate -s5368709120 ./tmp
```
or

Just copy a 5G file from local file system to s3fs.

Note: the maximum part size allowed by our s3 backend is 5GB.

gaul:
s3fs supports objects larger than 5 GB.  How do you mount s3fs?  Also can you run via `s3fs -d -f` and share the debug logs?

xrefft:
S3fs does support 5GB files. When the file grows slowly by `s3fs_write`, everything is ok. But when the file grows fast by truncate, it will be different.

The error occurred when calling the `s3fs_truncate` function. 
```log
[INF] s3fs.cpp:s3fs_getattr(971): [path=/big_file]
[INF] s3fs.cpp:s3fs_truncate(2220): [path=/big_file][size=5368709120]
```
`s3fs_truncate` called `pagelist.Resize` , And `pagelist.Resize` just append a large ”part“ into the pagelist.
```c++
bool PageList::Resize(off_t size, bool is_loaded, bool is_modified) {
        //...
	} else if (total < size) {
		// add new area
		fdpage page(total, (size - total), is_loaded, is_modified);
		pages.push_back(page);
        //..
```
Next, when `RowFlush` is called, the error happens naturally. Inside multipart upload, only 1 part will be uploaded.
```log
[INF]       fdcache.cpp:RowFlush(1860): [tpath=][path=/big_file][fd=9]
[INF]       curl.cpp:ParallelMixMultipartUploadRequest(1472): [tpath=/big_file][fd=9]
[INF]       curl.cpp:PreMultipartPostRequest(3442): [tpath=/big_file]
[INF]       curl.cpp:prepare_url(4624): URL is http://s3v2-qos.storage.wanyol.com/the-test-bkt/big_file?uploads
[INF]       curl.cpp:prepare_url(4657): URL changed is http://s3v2-qos.storage.wanyol.com/the-test-bkt/big_file?uploads
[INF]       curl.cpp:insertV4Headers(2694): computing signature [POST] [/big_file] [uploads] []
[INF]       curl.cpp:url_to_host(97): url is http://s3v2-qos.storage.wanyol.com
[INF]       curl.cpp:RequestPerform(2345): HTTP response code 200
[...]

PR #1319, opened 2020-06-27
Fixed upload error about mixuploading sparse file and truncating file

ggtakec:
#1311 #1318

### Details
There were cases where an error occurred when uploading a mix of files.
It occurs when the file is directly created by truncate(size larger than one entity size(5MB or more) of multipart upload).
In addition to truncate, it could also occur if the existing file size was changed significantly.

These bug has been fixed by this PR.
The fix is ​​to set the modify flag to ON in the stretched area when performing an open(internal processing) that specifies the file size.
This bug is due to a missing fix when supporting the mix upload process.

And I also added test cases(I also changed some existing code comments).


ggtakec:
@gaul Thanks.

PR #1322, opened 2020-07-03
Interim fix for Travis macOS Build

ggtakec:
#1268 

### Overview
Travis macOS Build uses Travis cache to reduce build time.  
And I confirmed that the Travis macOS build was successful because the cache exists, but it failed when the cache was deleted.  
And I also found other issues, this PR is an interim fix for those.  

### Background about macOS build on Travis-CI
The build of macOS uses macOS 10.12 which is the image of Travis CI(xcode9.2).  
The reason for using this image is that it's a version that doesn't require an OS reboot after installing osxfuse.  
OS images of 10.13 or later require reboot, and we cannot be built and tested s3fs.  
So we need to continue to use 10.12.  

And Homebrew is not currently offering this version of the binary package as Apple has dropped 10.12 support.  
However, Homebrew offers build and install from source code instead of binary packages.  

#### (1) If call brew tap without HOMEBREW_NO_AUTO_UPDATE=1, python3 cannot be installed
The failure reason is because Homebrew tries to install sqlite which is a python3 dependency, and fails.  
Currently, Travis macOS image (xcode9.2) fails to download the source package from sqlite.org site due to `No route` or `Timeout`.  
_The site has A and AAAA IP addresses, but both are in error._  

When there is no cache on travis, `HOMEBREW_NO_AUTO_UPDATE=1` was not attached, but `HOMEBREW_NO_AUTO_UPDATE=1` was provisionally specified in new code.  
Thus we can install python3.  

#### (2) awscli cannot be installed via Homebrew
For 10.12, Homebrew doesn't seem to install aswcli.  
So I changed it to install using pip.  

#### (3) truncate package
s3fs travis build had installed Homebrew's truncate package, which has been replaced by gnu's gtruncate provided by the coreutils package.  
Therefore, we do not install truncate.  

#### (4) cppcheck compatible
If we install cppcheck after running brew tap without `HOMEBREW_NO_AUTO_UPDATE=1`, cppcheck 2.1 will be installed.  
I found some errors detected in this version, so I fixed it(in curl.cpp).  
_In this PR code, HOMEBREW_NO_AUTO_UPDATE=1 is set, so 1.81 of cppcheck will be installed. However, the correction is added as a preventive measure._

#### (5) Modification of travis.yml
Incorporated some debugging messages and comments into .travis.yml.  
Moreover, the part where the warning was detected by the validation of Travis (matrix->jobs etc.) was corrected.

#### (6) Modification of test code
If testing without python3, skip the `write_multiple` test cases.  
I also fixed an incorrect call in `test_ut_ossfs` case.   

### Note
When merging this PR, please **clear the cache of master** on Travis-CI before merging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,10 @@
 #
 
 language: cpp
+dist: xenial
+os: linux
 
-matrix:
+jobs:
   include:
     - os: linux
       dist: trusty
@@ -29,7 +31,7 @@ matrix:
         - sudo apt-get update -qq
         - sudo apt-get install -qq attr cppcheck libfuse-dev openjdk-7-jdk
         - sudo update-alternatives --set java /usr/lib/jvm/java-7-openjdk-amd64/jre/bin/java
-        - sudo pip install --upgrade awscli
+        - sudo -H pip install --upgrade awscli
       script:
         - ./autogen.sh
         - ./configure CPPFLAGS='-I/usr/local/opt/openssl/include' CXXFLAGS='-std=c++03 -DS3FS_PTHREAD_ERRORCHECK=1'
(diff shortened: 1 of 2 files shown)


ggtakec:
This PR has been rebase after being modified by #1326.
And `(6)Modification of test code` is no longer needed, so it is no longer modified.

.travis.yml
```diff
@@ -57,28 +59,55 @@ matrix:
...
+            echo "python3 cannot be installed, so this is added as a temporary workaround.";
+            echo "If it is xcode 9.4 or higher, clear this patch.";
+            HOMEBREW_NO_AUTO_UPDATE=1 brew tap homebrew/homebrew-cask;
           else
+            echo "==> [Found cache] HOMEBREW_NO_AUTO_UPDATE=1 brew tap homebrew/homebrew-cask";
             HOMEBREW_NO_AUTO_UPDATE=1 brew tap homebrew/homebrew-cask;
           fi
         - HOMEBREW_NO_AUTO_UPDATE=1 brew cask install osxfuse
-        - S3FS_BREW_PACKAGES='awscli cppcheck truncate';
+        - S3FS_BREW_PACKAGES='cppcheck python3';
```
gaul, line 73: Needed?

.travis.yml
```diff
@@ -57,28 +59,55 @@ matrix:
...
+                HOMEBREW_NO_AUTO_UPDATE=1 brew upgrade ${s3fs_brew_pkg};
+              fi
             else
+              echo "==> Try to install ${s3fs_brew_pkg}";
               HOMEBREW_NO_AUTO_UPDATE=1 brew install ${s3fs_brew_pkg};
             fi;
           done
+        - pip3 --version;
+          if [ $? -eq 0 ]; then
+            echo "==> Try to install aswscli by pip3";
```
gaul, line 89: Spelling: awscli

.travis.yml
```diff
@@ -57,28 +59,55 @@ matrix:
...
+              if [ $? -eq 0 ]; then
+                echo "==> Try to upgrade ${s3fs_brew_pkg}";
+                HOMEBREW_NO_AUTO_UPDATE=1 brew upgrade ${s3fs_brew_pkg};
+              fi
             else
+              echo "==> Try to install ${s3fs_brew_pkg}";
               HOMEBREW_NO_AUTO_UPDATE=1 brew install ${s3fs_brew_pkg};
             fi;
           done
+        - pip3 --version;
```
gaul, line 87: Not needed with python2?

src/curl.cpp
```diff
@@ -4590,7 +4593,10 @@ void* S3fsMultiCurl::RequestPerformWrapper(void* arg)
 {
   S3fsCurl* s3fscurl= static_cast<S3fsCurl*>(arg);
   void*     result  = NULL;
-  if(s3fscurl && s3fscurl->fpLazySetup){
+  if(!s3fscurl){
+    return (void*)(intptr_t)(-EIO);
+  }
+  if(s3fscurl->fpLazySetup){
```
gaul, line 4599: Is this part of a separate change?

.travis.yml
ggtakec: This is because I want to install the python3 code with awscli install if possible.
I think awscli will not support python2 either soon.

.travis.yml
ggtakec: Sorry. I will fix it.

.travis.yml
ggtakec: Since 10.12 has python2 by default, I think it's fine to check only pip3.

src/curl.cpp
ggtakec: My judgment may be wrong, but I got an error if I did not fix it like this with cppcheck 2.1.
I noticed an error when cppcheck2.1 may be installed in some cases.
(But this PR code uses cppcheck1.8x)

gaul: I broken this into a separate PR #1329, preserving authorship.

ggtakec: OK, Thanks.
I will remove this after merging #1329. And re-rebase this PR.

ggtakec:
Again, this PR has been rebase after being modified by #1329.
And `(4)cppcheck compatible` is no longer needed, so it is no longer modified.

Issue #1323, opened 2020-07-03
Upgrade Travis macOS Builder


ggtakec:
I'll post this issue because we need to keep watching it in the future.  
This issue was first raised by @gaul in #1268.  
In addition, as a content related to this Issue, Travis CI build will be temporarily modified by #1322.

#### osxfuse and travis's image
The reason we are forced to use macOS 10.12 is that osxfuse changes an extension of macOS, and macOS image after 10.13 provided by travis  requires a reboot after osxfuse installation.  
We have to wait for the Travis macOS image to accommodate this.  

#### Homebrew supports
Homebrew's binary support for macOS 10.12 is already over, and we have to use 10.13(travis ci image: xcode9.4) or later.  
Also, awscli for macOS 10.12 can no longer be installed from Homebrew, so we must install it using pip(3).  

#### Python3
And if s3fs build run Homebrew tap without `HOMEBREW_NO_AUTO_UPDATE=1`, python3 cannot be installed(This may be temporary).  
The reason why it cannot be installed is because Homebrew fails to install sqlite, which has a dependency on python3.  
Currently, Travis macOS image(xcode9.2) fails to download the source package from sqlite.org site due to `No route` or `Timeout`.  
_The site has A and AAAA IP addresses, but both are in error._  

#### Test code by python3
The s3fs test code has tests that require python3.  
Currently some test cases only run when python3 is installed.  
However, this is an exceptional condition and we want to run it always.

gaul:
Related issue: travis-ci/travis-ci/issues/10017.

gaul:
Closing since we are moving to GitHub actions in #1488.

PR #1326, opened 2020-07-07
Use Python 2 for write_multiple_offsets

gaul:
This aligns with `ut_test.py`.  Using an older Python also allows
compatibility with the older macOS 10.12 Travis CI.  References #1323.


ggtakec:
Thanks! I'll merge this code.
After that, #1323 is rebased, the part where the test code has been skipped is restored, and PR is performed again.

However, we should remove the code for python2 when it's available for 10.13 and above.
Support for python2 has already ended.

Issue #1321, opened 2020-07-02
Stop recommending `s3fs#bucketname` in fstab


pxeger:
### Details about issue
In the README and other bits of documentation, you recommend to use this format of `fstab` listing:
```none
s3fs#bucketname /path/to/mountpoint fuse option1,option2 0 0
```
However, according to [`man 5 fstab`](https://helpmanual.io/man5/fstab), this is deprecated:
>  mount(8) and umount(8) support filesystem subtypes.  The subtype is defined by '.subtype' suffix.  For example 'fuse.sshfs'. It's recommended to use subtype notation rather than add any  pre‐fix to the first fstab field (for example 'sshfs#example.com' is deprecated).

Instead you should use
```none
bucketname /path/to/mountpoint fuse.s3fs option1,option2 0 0
```

gaul:
@pxeger Could you submit a pull request for this?

PR #1327, opened 2020-07-07
`s3fs#bucketname ... fuse` -> `bucketname ... fuse.s3fs` (#1321)

pxeger:
#1321

### Details

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ s3fs mybucket /path/to/mountpoint -o passwd_file=${HOME}/.passwd-s3fs -o dbgleve
 You can also mount on boot by entering the following line to `/etc/fstab`:
 
 ```
-s3fs#mybucket /path/to/mountpoint fuse _netdev,allow_other 0 0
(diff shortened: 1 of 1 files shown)


gaul:
Does this work on macOS?  I don't see it in the man page:

https://www.unix.com/man-page/osx/5/fstab/

pxeger:
That was kind of the point of me opening the issue rather than directly a PR, I (rashly?) expected someone to correct me if I was wrong, and it didn't seem to be so. And I don't have a Mac to test it. But I think very few people will be using this on macOS because it's kind of a server thing, so maybe they can figure it out for themselves?

ggtakec:
I tried it on my Mac, but mounting using /etc/fstab with osxfuse and s3fs seems difficult.
(Actually, I gave up on doing this with the mount command)
The path to osxfuse needed to execute is refused by SIP(System Integrity Protection).
So for the time being, I think that this wording is good because it is general, without considering the correspondence in osxfuse.

PR #1333, opened 2020-07-12
Put similar processing together into method GetCacheFileStatTopDir

ggtakec:
n/a 

### Details
Since the same process was repeated, I summarized it in a method.
(This is the method you will need for the subsequent PR.)


ggtakec:
This PR does not contain any logic changes, so we will merge it.
After this merge, another PR will be published. (This PR needed to be coded for later PR.)
@gaul  If there is a problem, I will correct it soon, so please comment.

PR #1341, opened 2020-07-25
Change default stat_cache_expire

gaul:
Previously s3fs cached files forever which confused users with
creating objects using another client.


gaul:
I think we could pick a different default but unlimited seems wrong to me.

gaul:
> My impression is that it feels small at 60 seconds.
> Considering the case of actually operating a file via only s3fs, I think that cache out will occur frequently at 60 seconds.
> This value is application dependent, so I think that maybe the first change should be better to be around 15 minutes.

I agree; updated PR.

PR #1345, opened 2020-07-31
repair write file corruption when the disk space is insufficient

liuyongqing:
https://github.com/s3fs-fuse/s3fs-fuse/issues/1346

Version: S3FS 1.8.6
Os：CentOS
Description:
When the disk space is insufficient, we use mv command move one file to s3fs mount point, the final data exist in s3 is zero length object, we can use -oensure_diskfree to easily reproduce it
image

Analysis:
the following is what happened in disk space sufficient case:
1. the process create a zero byte object in s3fs
2. the process write some data to local disk space, due to disk space is insufficient, triggerd init upload and upload several part 
3. the process utimesat function to set file time, which will call put object copy api and generate another zero byte object according to step 1 generated zero object
4. the process call close function, which will complete upload part
5. finally according to s3 timing, the zero object in step 3 will remain.

according to above analysis,  a non expect zero object exist in s3，in other case, the data is corrupt. 
image

Solution:
1. If disk space is insuffient, we return error, let user free more disk space for use, the short of this solution is：utimesat is not effective in new file, and can't upload several big file in same time：https://github.com/s3fs-fuse/s3fs-fuse/pull/1345/commits
2. change chmod、chown and utimesat and other s3fs api, if the path's FdEntity exist and uploadid exist, only add metadata to cache

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -2497,40 +2497,25 @@ ssize_t FdEntity::Write(const char* bytes, off_t start, size_t size)
   int     result = 0;
   ssize_t wsize;
 
-  if(0 == upload_id.length()){
-    // check disk space
-    off_t restsize = pagelist.GetTotalUnloadedPageSize(0, start) + size;
-    if(ReserveDiskSpace(restsize)){
-      // enough disk space
+  // check disk space
+  off_t restsize = pagelist.GetTotalUnloadedPageSize(0, start) + size;
+  if(ReserveDiskSpace(restsize)){
+    // enough disk space
 
-      // Load uninitialized area which starts from 0 to (start + size) before writing.
-      if(!FdEntity::mixmultipart){
-        if(0 < start){
-          result = Load(0, start, /*lock_already_held=*/ true);
-        }
+    // Load uninitialized area which starts from 0 to (start + size) before writing.
+    if(!FdEntity::mixmultipart){
+      if(0 < start){
+        result = Load(0, start, /*lock_already_held=*/ true);
       }
+    }
 
-      FdManager::FreeReservedDiskSpace(restsize);
-      if(0 != result){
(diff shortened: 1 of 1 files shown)


ggtakec:
@liuyongqing Thanks for your PR! (This bug is very bad)
@gaul I reproduced the problem and investigated the cause.

This problem(bug) seems to be a problem of s3fs_chmod/s3fs_utimens function(s3fs.cpp) and FdEntity::RowFlush().
The FdEntity::Write method seems fine.

If the system have enough free disk space, the process order is as follows:
1) create file
2) update time stamp for file(s3fs_utimens)
3) update file permission(s3fs_chmod)
4) multipart upload
   4-1) start
   4-2) upload each parts
   4-3) finish upload

If there is not enough free disk space, the order is as follows.
1) create file
4-1) multipart upload start
4-2) upload each parts (excluding the last part)
2) update time stamp for file(s3fs_utimens)
3) update file permission(s3fs_chmod)
4-2) upload last part
4-3) finish upload

In other words, multipart upload is started, and time/chmod processing is performed in the middle of it.
As a result, the file size is incorrect (size=0).

We were not aware of this problem.
In S3, if another Copy API is operated on the same file before the Copy API is completed, the process of executing it first and then performing Complete later seems to be ignored.

It seems that the s3fs_chmod/s3fs_utimens functions and FdEntity::RowFlush should be modified instead of modifying the FdEntity::Write method, so please wait for a while(I take a time to fix codes).

ggtakec:
I created new PR #1349.
I think this issue has been resolved without errors and the root cause has been prevented.

@liuyongqing 
If you can, please test with the PR code, and give us any comments.
If #1349 will be merged, this PR will be closed.

gaul:
Superseded by #1349.

Issue #1346, opened 2020-07-31
The moved file length become zero when disk space is insufficient


liuyongqing:
Version: S3FS 1.8.6
Os：CentOS
Description:
When the disk space is insufficient, we use mv command move one file to s3fs mount point, the final data exist in s3 is zero length object, we can use -oensure_diskfree to easily reproduce it
![image](https://user-images.githubusercontent.com/5205797/89029037-14f2d080-d360-11ea-9e60-6cb6f0408a13.png)

Analysis:
the following is what happened in disk space sufficient case:
1. the process create a zero byte object in s3fs
2. the process write some data to local disk space, due to disk space is insufficient, triggerd init upload and upload several part 
3. the process utimesat function to set file time, which will call put object copy api and generate another zero byte object according to step 1 generated zero object
4. the process call close function, which will complete upload part
5. finally according to s3 timing, the zero object in step 3 will remain.

according to above analysis,  a non expect zero object exist in s3，in other case, the data is corrupt.
![image](https://user-images.githubusercontent.com/5205797/89030548-48832a00-d363-11ea-94a1-dfd6d5f0f5a5.png)

Solution:
1. If disk space is insuffient, we return error, let user free more disk space for use, the short of this solution is：utimesat is not effective in new file, and can't upload several big file in same time：https://github.com/s3fs-fuse/s3fs-fuse/pull/1345/commits
2. change chmod、chown and utimesat and other s3fs api, if the path's FdEntity exist and uploadid exist, only add metadata to cache

gaul:
Please test with latest master and reopen is symptoms persist.

PR #1349, opened 2020-08-02
Fixed a bug about move file over limit of ensure space

ggtakec:
#1345 

### Details
There is a case where the header(meta) information is put during multipart upload, and the file size becomes 0.
Thus, this PR code is that s3fs updates header(meta) information after multipart upload is completed.

#### About bug
This bug happens when s3fs move a file from another disk(not managed by s3fs) when there is not enough free disk space.
(It seems to be the same with a copy, but it has not been confirmed)
In such a case, we need to PUT each header information etc. after the file upload is completed.

#### About test case
In addition, I considered a test case for confirming this fix, but it was not decided upon completion.


src/fdcache.cpp
```diff
@@ -2577,6 +2583,77 @@ ssize_t FdEntity::Write(const char* bytes, off_t start, size_t size)
...
 
+// [NOTE]
+// The meta(headers) cannot be updated while multipart upload is in progress.
+// Therefore, this method holds the header that will be updated if a multipart
+// upload is in progress and returns true.
+// All pending headers are updated after the multipart upload is complete.
+// Only the specified headers are retained as meta. Other headers are merged
+// when updated.
+//
+bool FdEntity::IsPendingMeta(const headers_t& meta, const metakeys_t& keys)
```
liuyongqing, line 2594: Hi， ggtakec, this method combines two function, judge and append headers to pending_headers，can this method split to two function for easy understand: IsUploading and UpdatePendingData

ggtakec:
Hi, @liuyongqing
Thanks for your advice.

Splitting the methods would be correct for clarity and as a public
method as an FdEntity class.
I also thought about this point.
However, it is a special process of Pending Headers, and there are
things I want to keep as an indivisible process, so I made it one
method like this.
In other words, I didn't want to check that uploading is in progress
and caching the Header as two separate methods.

Regards,
Takeshi

2020年8月3日(月) 22:16 liuyongqing <notifications@github.com>:

src/fdcache.h
```diff
@@ -166,6 +170,7 @@ class FdEntity
     bool SetAllStatus(bool is_loaded);                          // [NOTE] not locking
     //bool SetAllStatusLoaded(void) { return SetAllStatus(true); }
     bool SetAllStatusUnloaded(void) { return SetAllStatus(false); }
+	int UploadPendingMeta(void);
```
gaul, line 173: bad indentation

src/fdcache.h
```diff
@@ -183,6 +188,7 @@ class FdEntity
     bool RenamePath(const std::string& newpath, std::string& fentmapkey);
     int GetFd(void) const { return fd; }
     bool IsModified(void) const { return pagelist.IsModified(); }
+	bool IsPendingMeta(const headers_t& meta, const metakeys_t& keys);
```
gaul, line 191: bad indentation

src/s3fs.cpp
```diff
@@ -3426,11 +3461,18 @@ static int s3fs_removexattr(const char* path, const char* name)
   meta["x-amz-copy-source"]        = urlEncode(service_path + bucket + get_realpath(strpath.c_str()));
   meta["x-amz-metadata-directive"] = "REPLACE";
 
-  if(0 != put_headers(strpath.c_str(), meta, true)){
-    free_xattrs(xattrs);
-    return -EIO;
+  // Since meta data cannot be updated during uploading, it puts with a delay.
```
gaul, line 3464: This change confuses me -- why doesn't locking prevent this situation?  During sync, upload should take the metadata lock, copy the headers it needs, take the data lock, release the metadata lock, then initiate upload.  Is my understanding incorrect?

liuyongqing:
@ggtakec ,hi, ggtakec，there is one data corruption situtaion may happend in multi process, due to judge IsPendingMeta and put_headers is not atomic operation：
1. thread A calling IsPendingMeta -> put_headers, however put_headers not really executed due to process schedual
2. thread B calling Write triggerd init multipart upload
3. thread A put_headers are executed
4. thread A put_headers generate zero object will remain
although the above situation will not exist in sequential posix api call, but may exist in parallel posix api call

src/fdcache.h
ggtakec: Thanks, it will be fixed.

src/fdcache.h
ggtakec: Thanks, it will be fixed.

src/s3fs.cpp
ggtakec: If user moves(mv) the file, Hooks of s3fs are called as follows.
1. s3fs_create
1. s3fs_write (called multiple times for each block size)
1. s3fs_utimene
1. s3fs_chown
1. s3fs_chmod
1. s3fs_flush
1. s3fs_release

If the disk space is safe, a call to s3fs_flush will start and end the multipart upload.
However, if you are running out of disk space, multipart upload is starting during s3fs_write is called multiple times.
And the multipart upload is completed at s3fs_flush.
Thus, during this multipart upload, mtime/mod/owner changes will be called.
If this is blocked(exclusive object: mutex etc), the process itself will stop and s3fs will deadlock.

Therefore, while executing(after starting) multipart upload, the updated contents of the meta data will be accumulated.
I think that the function that may call put_header during multipart upload needs this accumulation processing.

ggtakec:
@liuyongqing Thanks for your comments.
I think this is the case where thread A and thread B are performing file operations at the same time.

thread A is uploading files and updating meta (stats) information.
thread B is also updating meta (stats) information.

There is contention by two separate threads, but s3fs does not support file locking, so this contention cannot be prevented.
In other words, it seems that the meta information of the thread written last will be the last information.

However, it is impossible to judge whether the file size is zero (possibly) and is correct.
However, if thread B was a valid process, I think it may be zero.

ggtakec:
@gaul @liuyongqing This fix doesn't seem to be complete yet.
The root cause of this is the same, but I have confirmed another bad things, so I will also fix it and try to update the code again.

liuyongqing:
@ggtakec ，hi，the corruption case I describled above due to multithread update metadata，may resolved by pushdown the put_headers logic to FdEntity and hold the FdEntity lock：
1. thread A calling utimesat function, holding FdEntity lock to update information(at this moment, no InitUpload triggerd)
2. thread B calling Write，due to FdEntity is locked，so it blocked, will not triggered InitUpload due to lack of disk space
3. thread A calling put_headers and release lock
4. thread B hold lock continue to write data and triggered init upload 
5. the last saved object is Thread B upload object which is a non-zero object, which is user expected
In the above process, due to thread A have the lock, prevent thread B triggered InitUpload before PutObject

ggtakec:
@liuyongqing I think I understand your opinion.
However, there are some challenges.

> thread B calling Write，due to FdEntity is locked，so it blocked, will not triggered InitUpload due to lack of disk space

In this part, thread B continues to receive write data from FUSE.
This can't be blocked because FUSE is passing the data(I think if it is blocked, it means file locking)
And this write data exceeds the free space, so a multipart upload must be started during receiving data.
In other words, no matter how to block inside s3fs, write data will continue to be received, so if we do not start to upload the file, we will have to reliably lose the data or need to generate an EIO error.

There is another concern. (This seems straightforward if the client program was multiprocess rather than multithreaded.)
If we choose to block, it will not only block during put_header processing, but we should certainly block during upload as well.
Suppose a file is open and writing to it.
At this time, if there is a process that calls put_header, the block under this exclusive control will continue to be blocked until the file is closed (flush).
This period is unknown.
If user opens a file, writes to it, but doesn't close(flush) it, updates to that file seem to block forever.

s3fs does not lock files(block operation).
I think that s3fs can block if s3fs is processing a very short time during system call, but I don't think it can block the case where a client program operates on threads/processes simultaneously.

liuyongqing:
> Suppose a file is open and writing to it.
At this time, if there is a process that calls put_header, the block under this exclusive control will continue to be blocked until the file is closed (flush).

Because FdEntity Write hold lock during write，so one Write finished, the process call utimesat、chmod will hold the lock, and according to uploadId whether empty to decide update meta to memory(uploadId is empty) or call PutObjectCopy in server side change meta

liuyongqing:
Because s3fs use a char sequence to find the FdEntity，so there only one FdEntity exist for one path and only one api call will hold FdEntity Lock;  if s3fs was blocked due to lock，user posix api call will be blocked, so no data was written to disk file

ggtakec:
@liuyongqing Thanks for pointing out.
I will change a bit of this PR code.
However, I want to keep the blocking period for IO operations to a minimum.
Of course, there will be file conflicts, so I will clarify the behavior at that time.
Please wait a bit.

ggtakec:
I reviewed the code.
In addition to #1346, defects in #1360 have also been confirmed, so I have fixed them as the same cause.

If meta data updates(such as chmod/chown/utimens) occur during multipart upload, these will be pending and modified after the file is flushed.
The other case, If a meta data update occurs when the file is open(not uploading), the header upload will be performed and the meta data inside the opened FdEntity is also updated.

In any of the above cases, even if file operations conflict, those are regarded as operations on one file object and are processed (including pending) on ​​a first-come-first-served basis.
If the file is opened and the multipart upload is started before flush system call(most likely due to lack of disk space), then the completion of the upload has top priority.

##### Related additional changes:
- In relation to the addition of utility functions in s3fs_util.cpp, some utility function variables should be const, so that is fixed.
- The test_cache_file_stat test is now run first  
This is because the use_cache and ensure_diskspace options are specified at the same time, then  in some cases the file is uploaded without using cache file, and the test fails(mainly on OSX).


This update is rebased and squashed.

PR #1356, opened 2020-08-11
Prefer %zd and %zu for ssize_t and size_t

gaul:
This removes unnecessary casting.


gaul:
Sorry, this isn't right.  I misunderstood `%j`.

PR #1362, opened 2020-08-15
Add script to run tests using sanitizers

gaul:
These currently show several kinds of errors.


gaul:
We cannot run these in CI since they are too slow.

Issue #1354, opened 2020-08-09
RHEL 6 (Linux 2.6.32) compatibility


gaul:
Recently a StackOverflow user reported an issue compiling s3fs using RHEL 6:

https://stackoverflow.com/questions/63303833/problem-installing-s3fs-on-the-rhel-6-system

This is due to the use of `SEEK_HOLE` and `SEEK_DATA` in `PageList::GetSparseFilePages`.  Linux 3.8 introduced these for ext4.  It appears that this works on macOS, at least modern versions.

s3fs support for RHEL 6 has always been complicated due its use of an older FUSE.  I don't think that we should do anything to support such an ancient distribution since users can compile an older version like 1.86, but I would like this to be an intentional change instead of accidental.  Setting a minimum requirement of RHEL 7 and Ubuntu 16.04 (or equivalent) allows s3fs freedom to modernize its dependencies, e.g., C++11, AWS SDK (#1068).

juliogonzalez:
Just a side note, RHEL 6 es its Maintenace Phase on November 30, 2020: https://access.redhat.com/articles/4665701

From that point the Extended Life Phase applies, and that means (https://access.redhat.com/support/policy/updates/errata#Extended_Life_Cycle_Phase): "Red Hat will provide limited ongoing technical support. No bug fixes, security fixes, hardware enablement or root-cause analysis will be available during this phase, and support will be provided on existing installations only."

CentOS6 goes EoL on November 30, 2020 as well: https://wiki.centos.org/About/Product

In short: documenting that RHEL6/CentOS6 are not supported should be fine, specially because I understand that 1.87 will still support it.

I suggest that a note is added on 1.87 release notes. Something like "WARNING: This is the last s3fs release supporting RHEL6/CentOS6 and Ubuntu <= 16.04). Please consider migrating to a newer OS."

I also suggest you consider the impact on other distributions when bumping dependencies, such as Debian or SLE (AFAIK both provide gcc with C+11 support, but that's to be checked).

In the end it's a matter of finding a balance, as using an old s3fs version can be a problem (since older versions do not have maintenance).

gaul:
Small correction, 1.86 was the last release that compiled on RHEL 6.  I would certainly merge a PR which provides a workaround for older Linux distributions but I cannot justify spending my own time to work on this.  Managing the backlog of issues for more modern environments is a large enough task already!

juliogonzalez:
[...]

PR #1365, opened 2020-08-16
Dynamically determine whether lseek extended options are supported

ggtakec:
#1354 

### Details
This PR is a code to avoid that s3fs cannot be built when whence SEEK_HOLE/SEEK_DATA of the lseek function is undefined.
Also, in the runtime environment where lseek cannot recognize these whence, s3fs will not start up with an error if the set_check_cache_sigusr1 option is specified.

This fix will broaden the build and run environments.

However, I think that we do not want to change the target OS supported by s3fs by this fix does.
The environment that can be built/executed is increase to wide, but the OS environment that can be officially supported follows the contents discussed in #1354.


ggtakec:
@gaul This PR is merged to proceed with future merges.

PR #1385, opened 2020-09-07
close FdEntity to avoid leakage

liuyongqing:
close FdEntity to avoid leakage

### Details
close opened FdEntity in s3fs_chmod、s3fs_chown、s3fs_utimens、s3fs_setxattr and s3fs_removexattr

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1601,10 +1601,12 @@ static int s3fs_chmod(const char* _path, mode_t mode)
                 // allow to put header
                 // updatemeta already merged the orgmeta of the opened files.
                 if(0 != put_headers(strpath.c_str(), updatemeta, true)){
+                    FdManager::get()->Close(ent);
                     return -EIO;
                 }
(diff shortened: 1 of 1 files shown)


liuyongqing:
@ggtakec ,hi ggtakec，s3fs_utimens and several other s3fs function call FdManager::get()->ExistOpen, but not close it

ggtakec:
@liuyongqing Thanks for your help, it's my fault.

gaul:
This seems reasonable.  But how can we detect these situations, e.g., Valgrind?

ggtakec:
@gaul I'll consider whether Valgrind can detect this bug.
However, it may be difficult to detect because it is a logic that only increases the reference count in FdEntity, and because file descriptors that remain open until the end of the process are automatically closed.
For that, we may need some logic to check the number of file descriptors left behind internally, with no files open via s3fs.
In any case, we will need to consider it in the future so as not to embed similar defects.

PR #1388, opened 2020-09-10
return 0 after parsing "use_session_token" arg

### LINKED ISSUE #651: Where to put `aws_session_token` information
#### Additional Information

- Version of s3fs being used (s3fs --version): *Amazon Simple Storage Service File System V1.82(commit:f4515b5) with OpenSSL*

- Version of fuse being used (pkg-config --modversion fuse): *2.9.2*

- System information (uname -a): *Linux 3.10.0-693.el7.x86_64 #1 SMP Thu Jul 6 19:56:57 EDT 2017 x86_64 x86_64 x86_64 GNU/Linux*
 - Distro (cat /etc/issue): *Red Hat Enterprise Linux Workstation release 7.4 (Maipo)*

#### Details about issue

We currently use temporary credentials using `aws sts`, from the documentation it is not clear to me how or where I would put the `aws_session_token`. Without setting this, I cannot access aws s3 buckets (i.e. I have to refresh this credential each day to gain access to the s3 buckets again). Can someone please point me to where I need to set this value in s3fs? Thanks!

maresb:
#651 

### Details
I hope this might fix #651.

There's clearly an open bug in #651 confirmed by several users related to the parsing of `-o use_session_token`. Looking for anomalies in the source code, I noticed that there was a return value everywhere except here, so I suspect this may be (at least partially) responsible.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -4285,6 +4285,7 @@ static int my_fuse_opt_proc(void* data, const char* arg, int key, struct fuse_ar
         }
         if (0 == STR2NCMP(arg, "use_session_token")) {
             is_use_session_token = true;
+	    return 0;
         }
         if(0 == STR2NCMP(arg, "ibm_iam_endpoint=")){
             std::string endpoint_url;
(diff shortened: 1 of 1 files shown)


gaul:
Thank you for your contribution @maresb!

PR #1390, opened 2020-09-11
Remove uses of implicit namespace std

gaul:
Fixed via:

```
sed -i '/using namespace std/{N;d}' src/*.cpp
sed -i 's/ string/ std::string/g' src/*.cpp
sed -i 's/(string/(std::string/g' src/*.cpp
sed -i 's/\[string/\[std::string/g' src/*.cpp
sed -i 's/^string/std::string/g' src/*.cpp
sed -i 's/ ifstream/ std::ifstream/g' src/*.cpp
sed -i 's/ istringstream/ std::istringstream/g' src/*.cpp
sed -i 's/ ostringstream/ std::ostringstream/g' src/*.cpp
sed -i 's/ max(/ std::max(/g' src/*.cpp
sed -i 's/ min(/ std::min(/g' src/*.cpp
sed -i 's/ endl/ std::endl/g' src/*.cpp
```


src/curl.cpp
```diff
@@ -106,26 +104,26 @@ sse_type_t       S3fsCurl::ssetype             = sse_type_t::SSE_DISABLE;
 bool             S3fsCurl::is_content_md5      = false;
 bool             S3fsCurl::is_verbose          = false;
 bool             S3fsCurl::is_dump_body        = false;
-string           S3fsCurl::AWSAccessKeyId;
-string           S3fsCurl::AWSSecretAccessKey;
-string           S3fsCurl::AWSAccessToken;
+std::string           S3fsCurl::AWSAccessKeyId;
```
ggtakec, line 107: If possible, please align the indentation.

src/fdcache.cpp
```diff
@@ -93,7 +91,7 @@ pthread_mutex_t FdManager::fd_manager_lock;
 pthread_mutex_t FdManager::cache_cleanup_lock;
 pthread_mutex_t FdManager::reserved_diskspace_lock;
 bool            FdManager::is_lock_init(false);
-string          FdManager::cache_dir;
+std::string          FdManager::cache_dir;
```
ggtakec, line 94: If possible, please align the indentation.

ggtakec (changes requested):
I understand the changes to namespace explicit, I think it's good.
I was worried that the indentation was slightly off.

src/curl.cpp
gaul: Done.

src/fdcache.cpp
gaul: Done.

gaul:
> I understand the changes to namespace explicit, I think it's good.
> I was worried that the indentation was slightly off.

I fixed most of these, at least the ones I found after two reviews.  We can tidy this up more with `clang-format`.

ggtakec (approved):

ggtakec:
Thanks, I'll merge this soon.

PR #1404, opened 2020-09-16
fix dead lock in disk insufficient and optimize code

liuyongqing:
Dead lock will occured in UploadPendingMeta which will call put_headers, the situation like follow:
1. RowFlush get fdent_data_lock, fdent_lock and call put_headers and try to lock fdmanager_lock
2. A another thread call s3fs_getattr， get fd_manager_lock and try get lock fdent_lock to increment refecnt
In step one and step two, the dead lock occured, the solution is not get fdmanager lock to update mtime. In adittion, the file x-cos-meta-xxx can directly update to orgmeta and only need to call put_headers once in UploadPendingMeta.

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -1032,7 +1032,6 @@ int FdEntity::NoCachePreMultipartPost()
     // initialize multipart upload values
     upload_id.erase();
     etaglist.clear();
-    pending_headers.clear();
 
     S3fsCurl s3fscurl(true);
     int      result;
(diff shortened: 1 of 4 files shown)


liuyongqing:
@ggtake，hi， ggtake，I have test the previous fix code about move data to mount point when disk is shortage, there is a dead lock  situation described above.  I optimize the code, please review the code, thanks.

liuyongqing:
@ggtakec ，hi, ggtake，is there any suggestions for the dead lock repair code

ggtakec (commented):
@liuyongqing Thanks for your great help!
I understand the deadlock bug that you have detected.

However, there may be a problem with the fix.
The bug I fixed in #1349 maybe recur. Or I think we need to reconfirm it.(I haven't been able to verify it yet.)

But, instead of the modification code shown by you, I think that you can deal with it by modifying only the FdEntity :: UploadPendingMeta method without changing the logic so far.
I will be able to put a PR later about that code.

ggtakec:
@liuyongqing Could you please check if the bug has been fixed in #1410?

liuyongqing:
@ggtakec ，hi，ggtakec，if one multipart uploading is processing, the above code will not upload pending meta until RowFlush，so the zero object [1349](https://github.com/s3fs-fuse/s3fs-fuse/pull/1349) problem will not appear, I have test the disk shortage  move case ；if user call chmod、chown many times，it seems we only need to update once，no need to try many times.

ggtakec:
@liuyongqing Thanks for your test about #1349.
 As you have shown, the final update will suffice for the call to put_headers.
I will merge this PR soon, and close #1410.

Issue #1402, opened 2020-09-15
 Timestamp not updated on the s3fs mount point of same file ( without content change)


kddon:
### Additional Information
_The following information is very important in order to help us to help you. Omission of the following details may delay your support request or receive no attention at all._
_Keep in mind that the commands we provide to retrieve information are oriented to GNU/Linux Distributions, so you could need to use others if you use s3fs on macOS or BSD_

#### Version of s3fs being used (s3fs --version)
_example: 1.00_
V1.87
#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
_example: 2.9.4_
2.9.2
#### Kernel information (uname -r)
_command result: uname -r_
4.14.192-147.314.amzn2.x86_64
#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
_command result: cat /etc/os-release_
Amazon Linux

#### s3fs command line used, if applicable
```s3fs <s3 bucket> -o use_cache=/tmp,iam_role="iam_role",allow_other,umask=022 <mount point>
```
#### /etc/fstab entry, if applicable
```
```
#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
_if you execute s3fs with dbglevel, curldbg option, you can get detail debug messages_
```
```
### Details about issue
We are using s3fs to mount the s3 bucket on a local drive. When we upload the same file( without content change) on the s3 bucket the timestamp of a file is changed but that file timestamp not changed in the s3fs mount point its show the old timestamp.

If we update the content of file and then upload it to the s3 bucket the timestamp changes on s3 bucket and s3fs mount point as well but not for the file which has same content.

Please kindly update.

gaul:
`integration-test-main.sh` has `test_update_time` which should exercise this.  Can you look at this and provide a complete test that demonstrates your symptoms?

kddon:
how can i execute this integration-test-main.sh? is there any parameter required

gaul:
You can run the tests if you check out the master branch from GitHub then run `make` and `make check -C test`.  (You can find this information in `.travis.yml`)

kddon:
Hi, I have execute make check -C test and getting below error.

make: Entering directory `/root/s3fs-fuse/test'
make  check-TESTS
make[1]: Entering directory `/root/s3fs-fuse/test'
make[2]: Entering directory `/root/s3fs-fuse/test'
[...]

PR #1407, opened 2020-09-17
Rework s3fs pid handling

gaul:
Previously `S3FS_PID` was not set in the correct shell and thus
`ps u $S3FS_PID` showed all programs.  This caused the flag detection to
fail when users ran other instances of s3fs with different flags.
References #1402.


test/integration-test-main.sh
```diff
@@ -318,8 +318,8 @@ function test_external_directory_creation {
     describe "Test external directory creation ..."
     OBJECT_NAME="$(basename $PWD)/directory/${TEST_TEXT_FILE}"
     echo "data" | aws_cli s3 cp - "s3://${TEST_BUCKET_1}/${OBJECT_NAME}"
-    ls | grep directory
-    get_permissions directory | grep ^750$
+    ls | grep -q directory
+    get_permissions directory | grep -q 750$
```
gaul, line 322: This is necessary to match macOS permission of `40750`.

ggtakec (approved):

Issue #1162, opened 2019-09-30
s3fs crashes with segmentation fault error in amazon linux


ggegiya1:
### Additional Information
Unable to mount an s3 bucket locally using s3fs  on Amazon Linux AWS EC2 instance:
```
cannot access /mnt/s3/bucket: Transport endpoint is not connected
```
s3fs mount command exits quitely, although, an error message is reported in dmesg:
```
$ dmesg|tail
[ 6666.320098] s3fs[13661]: segfault at 7f1cbe6dfff8 ip 00007f1cc4bc8c9f sp 00007f1cbe6e0000 error 6 in libc-2.17.so[7f1cc4b43000+1c2000]
```
#### Version of s3fs being used (s3fs --version)
Amazon Simple Storage Service File System V1.85 (commit:1db94a0) with OpenSSL

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
2.9.4

#### Kernel information (uname -r)
4.14.138-89.102.amzn1.x86_64 #1 SMP Thu Aug 15 15:41:44 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
NAME="Amazon Linux AMI"
VERSION="2018.03"
ID="amzn"
ID_LIKE="rhel fedora"
VERSION_ID="2018.03"
PRETTY_NAME="Amazon Linux AMI 2018.03"
ANSI_COLOR="0;33"
CPE_NAME="cpe:/o:amazon:linux:2018.03:ga"
HOME_URL="http://aws.amazon.com/amazon-linux-ami/"

#### s3fs command line used, if applicable
```
/usr/local/bin/s3fs #{bucketName} /mnt/s3/#{bucketName} -o iam_role -o allow_other -o endpoint=#{bucketRegion} -o uid=495  -o gid=493
```
#### /etc/fstab entry, if applicable
```
```
#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
_if you execute s3fs with dbglevel, curldbg option, you can get detail debug messages_
```
$ dmesg|tail
[ 6666.320098] s3fs[13661]: segfault at 7f1cbe6dfff8 ip 00007f1cc4bc8c9f sp 00007f1cbe6e0000 error 6 in libc-2.17.so[7f1cc4b43000+1c2000]
```
[...]

PR #1408, opened 2020-09-17
Ensure environment variable is set when using ECS

gaul:
Previously s3fs crashed when `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`
was not set.  References #1162.


ggtakec (approved):

ggtakec:
@gaul Thanks

PR #1410, opened 2020-09-19
fix dead lock in disk insufficient and optimize code(feat #1404)

ggtakec:
#1404 

### Details
This is a bug fix that @liuyongqing found.
Since the modification of #1404 may have a large impact, it is a code modified with a small impact range.


ggtakec:
It does not merge, it applies #1404.

PR #1413, opened 2020-09-20
not call put headers if not exist pending meta

liuyongqing:
The repair code in [1404](https://github.com/s3fs-fuse/s3fs-fuse/pull/1404)  still have some problem, there is no need to call put_headers method if not exist pending meta，

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -30,6 +30,7 @@
 #include "fdcache_entity.h"
 #include "fdcache.h"
 #include "string_util.h"
+#include "s3fs_util.h"
 #include "autolock.h"
 #include "curl.h"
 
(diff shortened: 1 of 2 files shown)


src/fdcache_entity.cpp
```diff
@@ -1476,23 +1476,28 @@ bool FdEntity::MergeOrgMeta(headers_t& updatemeta)
     if(0 != updatetime){
         SetCtime(updatetime, true);
     }
-    bool is_pending = !upload_id.empty();
+    is_meta_pending = !upload_id.empty();
```
ggtakec, line 1479: If is_meta_pending is true, it may be rewritten to false.
I think it should be: 
`is_meta_pending |= ! upload_id.empty ();`

ggtakec (changes requested):

src/fdcache_entity.cpp
liuyongqing: accept

liuyongqing:
@ggtakec ，hi, ggtakec，the code was changed according to your advice, thanks for your suggestion.

ggtakec:
@liuyongqing Thanks for your changes.
But sorry, we have an error on travis-ci.com ( migrating from travis-ci.org ), please wait merging until fixing it.

ggtakec (approved):

Issue #1415, opened 2020-09-21
Migrate from travis-ci.org to travis-ci.com


ggtakec:
## Details about issue
Travis will close travis-ci.org on December 31st, 2020 as follows:
By then, move to travis-ci.com.

> In an effort to ensure that all of our users - whether you build open-source, public or private repositories - receive regular feature updates, security patches and UX/UI enhancements, we are announcing that travis-ci.org will be officially closed down completely on December 31st, 2020, allowing us to focus all our efforts on bringing new features and fixes to travis-ci.com and all of our awesome users like yourself on the travis-ci.com domain. 

Related URL: https://blog.travis-ci.com/2018-05-02-open-source-projects-on-travis-ci-com-with-github-apps

gaul:
Done: https://travis-ci.com/github/s3fs-fuse/s3fs-fuse

ggtakec:
Thanks!

ggtakec:
I found an error on travis-ci.com, it occurred on ppc64 about java loading.
```
$ sudo update-alternatives --set java /usr/lib/jvm/java-7-openjdk-ppc64el/jre/bin/java
update-alternatives: error: alternative /usr/lib/jvm/java-7-openjdk-ppc64el/jre/bin/java for java not registered; not setting
The command "sudo update-alternatives --set java /usr/lib/jvm/java-7-openjdk-ppc64el/jre/bin/java" failed and exited with 2 during .
```

PR #1418, opened 2020-09-22
Specify Java 8 for Travis

gaul:
The next version of S3Proxy will require at least Java 8 and Java 7
seems to have some incompatibility with travis-ci.com.
References #1415.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,7 @@ jobs:
       cache: apt
       before_install:
         - sudo apt-get update -qq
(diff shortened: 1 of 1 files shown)


gaul:
So this is a little awkward, we can't install Java 7 which doesn't have the jclouds `Closer` race condition but the next version of S3Proxy is waiting on the next version of jclouds to release with the fix.  The easiest solution seems to be to remove the PowerPC build temporarily.

gaul:
Superseded by #1424.

PR #1435, opened 2020-10-02
Remove exceptions from s3fs_strtoofft

gaul:
Explicit return codes are simpler and safer.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -33,7 +33,7 @@ AC_CHECK_HEADERS([sys/xattr.h])
 AC_CHECK_HEADERS([attr/xattr.h])
 AC_CHECK_HEADERS([sys/extattr.h])
 
-CXXFLAGS="$CXXFLAGS -Wall -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2"
+CXXFLAGS="$CXXFLAGS -Wall -fno-exceptions -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2"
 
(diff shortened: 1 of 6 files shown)


configure.ac
```diff
@@ -33,7 +33,7 @@ AC_CHECK_HEADERS([sys/xattr.h])
 AC_CHECK_HEADERS([attr/xattr.h])
 AC_CHECK_HEADERS([sys/extattr.h])
 
-CXXFLAGS="$CXXFLAGS -Wall -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2"
+CXXFLAGS="$CXXFLAGS -Wall -fno-exceptions -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2"
```
ggtakec, line 36: I'm curious about the options.
I think that we don't have to add the -fno-exceptions option, please comment on this.
I feel that we only need to be careful not to use try-catch in our code.
(Are you specifying this to avoid accidentally writing code that handles exceptions?)

ggtakec (commented):
Please comment on the -fno-exceptions option.

configure.ac
gaul: This flag warns when the code throws or catches any exceptions.  The behavior of calling an external function which throws an exception, e.g., `operator new` and `std::bad_alloc`, is identical -- aborting the program since s3fs lacks a top-level catch block.  I can remove this but it is helpful to avoid using exceptions in the future and generates slightly better code.

PR #1436, opened 2020-10-02
Fixed signature error due to case of hex string

ggtakec:
#1433

### Details
The Travis-CI execution result at the time of PR of #1433 was good, but after merging to master, the following error(example) occurred and I could not pass through Travis-CI.
```
s3fs: [ERR] curl.cpp:RequestPerform(2246): HTTP response code 403, returning EPERM. Body Text: <?xml version='1.0' encoding='UTF-8'?><Error><Code>SignatureDoesNotMatch</Code><Message>Forbidden</Message><RequestId>4442587FB7D0A2F9</RequestId></Error>
```
This seems to be due to the difference in case in URL encoding, so I modified the code.

### DIFF
--- a/src/string_util.cpp
+++ b/src/string_util.cpp
@@ -163,7 +163,7 @@ std::string urlEncode(const std::string &s)
             result += c;
         }else{
             result += "%";
-            result += s3fs_hex(&c, 1);
+            result += s3fs_hex(&c, 1, false);
         }
     }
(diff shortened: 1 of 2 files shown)


ggtakec:
@gaul Sorry merging before your confirm.
First and foremost, we need to fix errors in the master branch for other PRs.

PR #1437, opened 2020-10-03
Added atime and Corrected atime/mtime/ctime operations

ggtakec:
n/a

### Details
Added access time (a time) just like a normal file system.  

I also reviewed the atime/ctime/mtime update patterns and made them the same as regular filesystems.  
(Currently, those update patterns are not the same as normal file operations.)  

The test case has been modified and the update of atime/mtime/ctime is regulated as follows according to the command.
- create file or `touch`  
update ctime, atime and mtime with same time value.
- `touch -a`  
update ctime and atime with same time value, mtime is kept.
- write file (ex. `echo *** > file`)  
update ctime and mtime with same time value, atime is kept.
- update attributes (ex. `chmod`)  
update only ctime, atime and mtime are kept.
- copy file with attributes (ex. `cp -p **** ****`)  
update only ctime, atime and mtime are kept.
- move file (ex. `mv **** ****`)  
update only ctime, atime and mtime are kept.

_This fix and current code do not support updating the stat time of the parent directory of the file which is updated._
_Further consideration is needed to support this, and it can be difficult to achieve._
_Because, if operating the file, PUT request etc will be called to update the parent directory(object), which causes problems with performance and the number of requests._

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -771,9 +771,21 @@ bool convert_header_to_stat(const char* path, const headers_t& meta, struct stat
 
     // mtime
     pst->st_mtime = get_mtime(meta);
+    if(pst->st_mtime < 0){
+        pst->st_mtime = 0L;
+    }
 
     // ctime
     pst->st_ctime = get_ctime(meta);
+    if(pst->st_ctime < 0){
+        pst->st_ctime = 0L;
+    }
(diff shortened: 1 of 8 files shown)


gaul:
Thanks for including extensive tests!

Issue #1427, opened 2020-09-29
1.87 - symlinks not working after remount


cris-b:
### Additional Information
_The following information is very important in order to help us to help you. Omission of the following details may delay your support request or receive no attention at all._
_Keep in mind that the commands we provide to retrieve information are oriented to GNU/Linux Distributions, so you could need to use others if you use s3fs on macOS or BSD_

#### Version of s3fs being used (s3fs --version)
1.87
#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
2.9.2

#### Kernel information (uname -r)

3.10.0-1127.18.2.el7.x86_64

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)

NAME="Red Hat Enterprise Linux Server"
VERSION="7.8 (Maipo)"

When using 1.87:

create symlink
ls -l <symlink>
lrwxrwxrwx. 1 crisb crisb       6 Sep 29 10:13 cristest -> data
unmount s3fs
mount s3fs
ls -l <symlink>
lrwxrwxrwx. 1 crisb crisb       6 Sep 29 10:13 cristest -> 

Rolling back to 1.86 solves the issue.

I'm only using _netdev, passwd_file and url parameters to mount.

gaul:
@ggtakec I bisected this to 1f796d432d767bdcfe476052fe890bdb5488a15c but this does not make sense to me.  Any ideas?

ggtakec:
@gaul This is a bug, I probably found its reason in code, please wait PR.

gaul:
Please test with the latest master.

PR #1447, opened 2020-10-10
Fixed a bug that symlink could not be read after restarting s3fs

ggtakec:
#1427 

### Details
When opening the symlink file, it was no longer necessary to load it into a temporary file or cache file.
This PR fixed it.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -688,7 +688,7 @@ static FdEntity* get_local_fent(AutoFdEntity& autoent, const char* path, bool is
     }
 
     // open
-    time_t mtime         = (!S_ISREG(stobj.st_mode) || S_ISLNK(stobj.st_mode)) ? -1 : stobj.st_mtime;
+    time_t mtime         = (!S_ISREG(stobj.st_mode) && !S_ISLNK(stobj.st_mode)) ? -1 : stobj.st_mtime;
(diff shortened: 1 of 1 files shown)


src/s3fs.cpp
```diff
@@ -688,7 +688,7 @@ static FdEntity* get_local_fent(AutoFdEntity& autoent, const char* path, bool is
     }
 
     // open
-    time_t mtime         = (!S_ISREG(stobj.st_mode) || S_ISLNK(stobj.st_mode)) ? -1 : stobj.st_mtime;
+    time_t mtime         = (!S_ISREG(stobj.st_mode) && !S_ISLNK(stobj.st_mode)) ? -1 : stobj.st_mtime;
```
gaul, line 691: Can we exercise this with a test somehow, maybe by limiting the statcache size?

ggtakec: This bug is not related to stats and stats cache.
The stats information(meta) is used the correct value.
However, it sets mtime to -1 in the get_local_fent function and does not read the contents of the symlink file.
( https://github.com/s3fs-fuse/s3fs-fuse/blob/master/src/fdcache_entity.cpp#L482-L490 )
This is the part that prevents the contents of a file from being read except for a normal file, like an object that indicates a directory.
In the case of symlink, it is necessary to read it, but there was a problem with the judgment statement of mtime.

Issue #1445, opened 2020-10-09
Support AWS IMDSv2 for role credentials


nmeyerhans:
As described in AWS [blog posts](https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/) and [documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html), the IMDSv2 is a revision of the EC2 instance metadata service that provides security benefits over the original metadata service.

The [addition of IMDSv2 support to the ec2metadata command](https://github.com/canonical/cloud-utils/pull/1/commits/991719ef079222ce844b27e7ceecee176308aecf) might serve as a useful example to illustrate the API changes. In the case of s3fs, if support for IMDS-compatible services outside of Amazon EC2 is desired, it may be necessary to fall back to IMDSv1 support in the event that the token API returns a 404. Alternatively, it might be reasonable to ask the user to explicitly choose to enable IMDSv2 support if it's desired.

gaul:
@nmeyerhans This seems straightforward to implement; would you like to submit a pull request?

nmeyerhans:
I have not started on the implementation, but I don't mind doing so. I'll try to get a PR to you in the next few days.

gaul:
@nmeyerhans Could you look at https://stackoverflow.com/questions/69031023/how-to-make-s3fs-use-imds-v2-when-mounting-s3-buckets-from-ec2-instance ?

erka:
@gaul it isn't fully support. There is an issue with getting the ec2 role from metadata automatic. It is called without token and failed.

nmeyerhans:
@gaul I've posted an update on that stackoverflow post. It appears that `S3fsCurl::LoadIAMRoleFromMetaData()` was never updated to include tokens, so `iam_role=auto` doesn't work if tokens are required. Explicitly specifying a role name should work.

I can prepare a PR to fix `S3fsCurl::LoadIAMRoleFromMetaData()`

PR #1462, opened 2020-10-29
Implement AWS IMDSv2 support

nmeyerhans:
#1445 

### Details

AWS IMDSv2 is a session oriented method for retrieving instance metadata,
including IAM credentials, in Amazon EC2.  It is enabled by default in
non-enforcing mode in AWS (meaning it retains backwards compatibility with
existing IMDSv1 clients), but can be switched to enforcing mode, in which
clients are required to return API tokens with requests.

With this change, we implement support for IMDSv2 and enable it by default when
IAM roles are our source for authentication credentials.  In the event that
s3fs is running in cloud environment offering an IMDSv1-compatible API, we
support graceful fallback to that mode.  It can also be selected explicitly via
the imdsv1only mount option.

More details on IMDSv2 are available at
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
and
https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/


nmeyerhans:
I've revised this PR to avoid using the C++11 `std::to_string` function when converting from an integer value. This should address the earlier test failure.

src/curl.cpp
```diff
@@ -2701,6 +2756,23 @@ int S3fsCurl::GetIAMCredentials()
         }
         url = std::string(S3fsCurl::IAM_cred_url) + env;
     }else{
+		if(S3fsCurl::IAM_api_version > 1){
```
gaul, line 2759: Fix strange indentation (tabs vs. 4-spaces?)

src/s3fs_help.cpp
```diff
@@ -313,6 +313,13 @@ static const char help_string[] =
     "      \"auto\", s3fs will automatically use the IAM role names that are set\n"
     "      to an instance. If you specify this option without any argument, it\n"
     "      is the same as that you have specified the \"auto\".\n"
+	"\n"
+	"   imdsv1only (default is to use IMDSv2)\n"
```
gaul, line 317: Please add to `doc/man/s3fs.1` as well.

src/curl.h
```diff
@@ -348,6 +356,7 @@ class S3fsCurl
         static void InitUserAgent();
         static bool SetRequesterPays(bool flag) { bool old_flag = S3fsCurl::requester_pays; S3fsCurl::requester_pays = flag; return old_flag; }
         static bool IsRequesterPays() { return S3fsCurl::requester_pays; }
+		static bool SetIMDSVersion(int version);
```
gaul, line 359: Fix indentation.

src/curl.cpp
```diff
@@ -1656,6 +1668,13 @@ bool S3fsCurl::ParseIAMCredentialResponse(const char* response, iamcredmap_t& ke
     return true;
 }
 
+bool S3fsCurl::SetIAMv2APIToken(const char* response)
+{
+    S3FS_PRN_INFO3("Setting AWS IMDSv2 API token to %s", response);
+    S3fsCurl::IAMv2_api_token = std::string(response);
+	return true;
```
gaul, line 1675: Indentation.

gaul (changes requested):
Please address small comments.  Overall this seems fine.

gaul:
Thank you for your contribution @nmeyerhans!

PR #1448, opened 2020-10-12
Flush dirty data after a writing a number of bytes

gaul:
This allows s3fs to write large files without consuming a large amount
of temporary local storage but can slow uploads due to server-side
copies.  References #617.  Fixed #1056.  Fixes #1257.

PR #1467, opened 2020-11-03
Fixed flushing dirty data and compressed the cache size

ggtakec:
#1448 

### Details
Regarding the process when the max_dirty_data option is specified(the process of uploading while writing a file), the following two have been modified.

#### Fixed a bug
Changed the buffer that stores the return value when uploading(Flush) during the processing of s3fs_write.  
The return value of the s3fs_write function must be the number of bytes written, so it returned the wrong value(error code).  
Due to this, there were cases where uploading could not be performed normally.  

#### Cache file compression
In the process by max_dirty_data option, when uploading while writing the file, the contents of the cache file remained as it was.  
If it remains, it will take up disk space.  
After uploading, new code punches a HOLE in the cache file so that it clears blocks on the hard disk for cached data.  
This will minimize disk space pressure when uploading with max_dirty_data option.  

##### Cache file state when max_dirty_data option is **not specified**
- Cache file status(if uploading 66000044 bytes)  
```
644532 -rw------- 1 guest users 660000044 Nov  3 05:20 big.txt
```
- Cache file stat information (`.<Bucketname>/big.txt`)  
```
1351054:660000044
0:660000044:1:0
```
##### Cache file state when max_dirty_data option is **specified**
- Cache file status(if uploading 66000044 bytes)  
```
25012 -rw------- 1 guest users 660000044 Nov  3 05:20 big.txt
```
- Cache file stat information (`.<Bucketname>/big.txt`)  
```
1351054:660000044
0:634388480:0:0
634388480:25611564:1:0
```

In this way, the disk space is not used except for the last uploaded area.  

##### Notes
The fallocate() function is used for this process, and this function is a non-portable Linux-specific system call.  
Therefore, this function does not exist except on Linux base(ex: OSX).  
To avoid this, I check the fallocate function in configure and implement a dummy function if it does not exist.  
The dummy function is always failing and does not compress the cache file.  
This is a limitation on OSX etc.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -32,6 +32,7 @@ AC_PROG_CC
 AC_CHECK_HEADERS([sys/xattr.h])
 AC_CHECK_HEADERS([attr/xattr.h])
 AC_CHECK_HEADERS([sys/extattr.h])
+AC_CHECK_FUNCS([fallocate])
 
 CXXFLAGS="$CXXFLAGS -Wall -fno-exceptions -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2"
 
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -1601,6 +1601,71 @@ int FdEntity::UploadPendingMeta()
     return result;
 }
 
+// [NOTE]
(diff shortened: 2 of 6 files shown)


gaul:
Thanks for completing this!  Now s3fs should be able to handle very large files.

PR #1475, opened 2020-11-13
Fixed .gitignore for src/test_string_util.cpp

ggtakec:
n/a

### Details
src/test_string_util.cpp should be excluded in .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,7 +73,7 @@ missing
 *.trs
 default_commit_hash
 src/s3fs
-src/test_*
+src/test_string_util
 test/s3proxy-*
 
 #


.gitignore
```diff
@@ -74,6 +74,7 @@ missing
 default_commit_hash
 src/s3fs
 src/test_*
+!src/test_string_util.cpp
```
gaul, line 77: Perhaps we should change `src/test_*` to `test_string_util`?

ggtakec: I was wondering which one to use, but is that better?
A new force push. soon

Issue #1454, opened 2020-10-16
test_update_time and test_update_directory_time fail due to access time


gaul:
I see this consistently on my Linux system:

```
touch with -a option expected updated ctime: 1602857824 != 1602857824, atime: 1602857824 != 1602857824 and same mtime: 1602857824 == 1602857824
```

I wonder why we don't see this in Travis?  Likely a regression from #1437.

ggtakec:
The s3fs entry point called by the touch command is `the s3fs_utimens()` function.
Then the value of `mtime`/`ctime`/`atime` are passed to the argument timespec [2] of this function.

The processing inside the `s3fs_utimens()` function differs depending on whether the file is opened or not.
If you were running the s3fs test script, you probably haven't opened the file.
In this case, `merge_headers()` sets each header, and then `put_headers()` is called, which is a simple process.

I think from this error:
It is highly possible that `ts[0].tv_sec` of the argument `timespec[2]` of the `s3fs_utimens()` function is the same as `atime`/`ctime` of the file before executing the touch command.
If you are logging, you can look for the line like below.
```
[INF] s3fs.cpp:s3fs_utimens(1925): [path=/zzz][mtime=1601651879][ctime/atime=1602901289]
```
I think that if the `atime`/`ctime` value in this log is the same as the `atime`/`ctime` value set in the file, then the defect does not seem to be s3fs.
In other words, there seems to be a problem up to the call from fuse.
The test program waits 2 seconds before executing the touch command, so basically I don't think this will happen.
It's a mystery.

gaul:
I reproduced these symptoms on both my Fedora 32 desktop and Debian 10 Raspberry Pi:

```
s3fs: [INF] s3fs.cpp:s3fs_utimens(1925): [path=/testrun-21107/test-s3fs.txt][mtime=1602905290][ctime/atime=1602905290]
...
touch with -a option expected updated ctime: 1602905290 != 1602905290, atime: 1602905290 != 1602905290 and same mtime: 1602905290 == 1602905290
```

ggtakec:
It seems that `s3fs_utimens` is called with this error value as an argument.
That means that as the logic in s3fs, each value is set correctly.
We need to find out why the function was called with this value.

Will the results change when you run the touch(-a) command manually or when you increase the sleep seconds in the test_update_time test?

In integration-test-main.sh, there is a test for the `touch` command before` touch -a`.
[...]

PR #1477, opened 2020-11-18
Added a condition to the atime test

ggtakec:
#1454

### Details
The `atime` test may fail.
It happens because FUSE does not call the `s3fs_utimens` entry function.

The cause of this has not been clarified yet, but it seems that it depends on the options and stats when mounting.
It seems to be options such as relatime/noatime.

With the exception of ubuntu:xenial, the test seems to fail if the relatime/noatime option is given.

Therefore, to avoid test failures, bypass the `touch -a` command test if the relatime/noatime option is set.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -583,6 +583,22 @@ function test_mtime_file {
     rm_test_file $ALT_TEST_TEXT_FILE
 }
 
+# [NOTE]
+# If it mounted with relatime or noatime options , the "touch -a"
+# command may not update the atime.
+# In ubuntu:xenial, atime was updated even if relatime was granted.
+# However, it was not updated in bionic/focal.
+# We can probably update atime by explicitly specifying the strictatime
+# option and running the "touch -a" command. However, the strictatime
+# option cannot be set.
+# Therefore, if the relatime option is set, the test with the "touch -a"
+# command is bypassed.
+# We do not know why atime is not updated may or not be affected by
+# these options.(can't say for sure)
+# However, if atime has not been updated, the s3fs_utimens entry point
+# will not be called from FUSE library. We added this bypass because
+# the test became unstable.
+#
 function test_update_time() {
     describe "Testing update time function ..."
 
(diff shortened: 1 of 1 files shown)


test/integration-test-main.sh
```diff
@@ -656,17 +672,19 @@ function test_update_time() {
...
     # "touch -a" -> update ctime/atime, not update mtime
     #
-    touch -a $TEST_TEXT_FILE
-    atime=`get_atime $TEST_TEXT_FILE`
-    ctime=`get_ctime $TEST_TEXT_FILE`
-    mtime=`get_mtime $TEST_TEXT_FILE`
-    if [ $base_atime -eq $atime -o $base_ctime -eq $ctime -o $base_mtime -ne $mtime ]; then
-       echo "touch with -a option expected updated ctime: $base_ctime != $ctime, atime: $base_atime != $atime and same mtime: $base_mtime == $mtime"
-       return 1
+    if ! mount | grep s3fs | grep -e noatime -e relatime >/dev/null; then
```
gaul, line 675: Please use `ps u $S3FS_PID | grep -q -e noatime -e relatime` which works when a user has multiple s3fs processes.

test/integration-test-main.sh
```diff
@@ -798,17 +820,19 @@ function test_update_directory_time() {
...
     # "touch -a" -> update ctime/atime, not update mtime
     #
-    touch -a $TEST_DIR
-    atime=`get_atime $TEST_DIR`
-    ctime=`get_ctime $TEST_DIR`
-    mtime=`get_mtime $TEST_DIR`
-    if [ $base_atime -eq $atime -o $base_ctime -eq $ctime -o $base_mtime -ne $mtime ]; then
-       echo "touch with -a option expected updated ctime: $base_ctime != $ctime, atime: $base_atime != $atime and same mtime: $base_mtime == $mtime"
-       return 1
+    if ! mount | grep s3fs | grep -e noatime -e relatime >/dev/null; then
```
gaul, line 823: Same as above.

gaul (changes requested):
Let me research this a little more.

test/integration-test-main.sh
ggtakec: @gaul Thanks for a comment.
ps command only knows the parameters which s3fs starts to run with.
You can find the parameters given by the OS by default in the mount command, /etc/mtab, and /proc/mounts.
Therefore, I used those commands or files to see the options that I didn't specify(or were ignored or set) at boot time.
(`cat /proc/mounts` may have been better than the mount command.)

However, it does not correspond to the state of the multiple s3fs process you pointed out, so I will fix it.
Since it cannot be determined by PID, modify it so that it is confirmed by the mount point.

test/integration-test-main.sh
ggtakec: Same as above. Please double check after I change the code.

ggtakec:
@gaul I would like to merge this PR.
The `relatime` option is granted by the OS and cannot be changed in some cases.
In that case, atime change requests do not reach s3fs through FUSE.
Therefore, if we do not apply this patch, the test will not pass on some OS(because `reltime` option depends on that OS).

I will merge this PR assuming that the source code has not been changed by this PR, but if more problems occur in the future, I will create a new issue.

PR #1478, opened 2020-11-19
Fixed errors detected by cppcheck 2.2

ggtakec:
n/a

### Details
Fixed an error detected by cppcheck 2.2.
Errors detected in curl_utils.cpp: nullPointerRedundantCheck is probably a false positive.
I have set the pointer of the variable of etaglist_t to add_etag_list (), but an error was output because it is an auto variable.
For this error, To fix this the memory is allocated and a variable pointer is passed.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -3913,7 +3913,6 @@ int S3fsCurl::MultipartUploadRequest(const char* tpath, headers_t& meta, int fd,
     std::string    upload_id;
     struct stat    st;
     int            fd2;
-    etaglist_t     list;
     off_t          remaining_bytes;
     off_t          chunk;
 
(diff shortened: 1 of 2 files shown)


gaul:
> I have set the pointer of the variable of etaglist_t to add_etag_list (), but an error was output because it is an auto variable.
> For this error, To fix this the memory is allocated and a variable pointer is passed.

I think cppcheck is partially correct and partially wrong here due to an incomplete lifetime analysis.  The shorter-lived object should not hold onto pointers to the longer-lived object and instead the longer-lived object should access a field within the shorter-lived object.  #1479 shows an alternate approach which simplifies the code but actually has a similar issues that cppcheck does not warn about.

ggtakec:
@gaul Thanks for your comment.
I have confirmed that #1479 does not give an error in cppcheck 2.2.
And since #1479 is the more appropriate fix, this PR will be closed.
(I understand the false positives of cppcheck.)

ggtakec:
@gaul Sorry for my mistake, one of them needed to be fixed, so I will issue another PR.

PR #1492, opened 2020-12-10
fix 404 error message

kontrollanten:
### Relevant Issue
* #1377

### Details
AWS can return an explicit 404 message that says that the key doesn't exist:

```
[INF]       curl.cpp:RequestPerform(2441): HTTP response code 404 was returned, returning ENOENT
[ERR] curl.cpp:CheckBucket(3439): Check bucket failed, S3 response: <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Key>my-key/</Key><RequestId>F571C209C9ACA4FC</RequestId><HostId>SBKSxWs3ULB8R8ev....</HostId></Error>
[CRT] s3fs.cpp:s3fs_check_service(3901): bucket not found(host=https://s3.amazonaws.com) - result of checking service.
[
```

It would be great if the message provided by AWS could be printed to the user as well, but I've no idea about how to that in C++.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -3486,7 +3486,7 @@ static int s3fs_check_service()
                 S3FS_PRN_CRIT("invalid credentials(host=%s) - result of checking service.", s3host.c_str());
 
             }else if(responseCode == 404){
-                S3FS_PRN_CRIT("bucket not found(host=%s) - result of checking service.", s3host.c_str());
+                S3FS_PRN_CRIT("bucket or key not found(host=%s) - result of checking service.", s3host.c_str());
 
             }else{
                 // another error


gaul:
@kontrollanten When does this trigger, when mounting a bucket with a path?  Also can you expand on your comment about printing the message provided by AWS?  Do you mean cleaning up the XML message?

kontrollanten:
I'm sorry I was unclear. Yes, this is triggered when mount a bucket with a path.

When not passing the debug flag to s3fs the error message provided to the user is:

```
s3fs.cpp:s3fs_check_service(3901): bucket not found(host=https://s3.amazonaws.com) - result of checking service.
```

So the ideal would be to change that to:

```
s3fs.cpp:s3fs_check_service(3901): bucket or key not found(host=https://s3.amazonaws.com) - result of checking service.
NoSuchKey - The specified key does not exist.
```

Or it would be even better to just write "bucket not found" or "key not found", depending on the `Key` value in the XML provided by AWS.

gaul:
Thank you for your contribution @kontrollanten!  I think that s3fs should echo the actual `<Message>` from the response but your PR is an improvement.

PR #1497, opened 2020-12-14
Update README note on eventual consistency

adamchainz:
n/a

### Details
AWS promise read-after-write consistency now ([announcement](https://aws.amazon.com/about-aws/whats-new/2020/12/amazon-s3-now-delivers-strong-read-after-write-consistency-automatically-for-all-applications/)), but presumably non-AWS S3-compatible services are still eventually consistent.


gaul:
Thank you for your contribution @adamchainz!

adamchainz:
Thanks for maintaining this great tool.

PR #1489, opened 2020-12-02
Added Github Actions workflow

ggtakec:
### Relevant Issue (if applicable)
#1488 

### Details
This PR adds a Github Actions workflow for testing on 11 OS(version).

We have only 2 operating systems on TravisCI, but this PR(Github Actions) can build and test on the following operating systems.
- Ubuntu 20.04/18.04/16.04
- Debian 10(buster)/9(stretch)
- CentOS 8/9
- Fedora 32/31
- openSUSE Leap 15
- macos 10

### Notes
- macos11 is not supported as it requires a reboot after installing osx fuse.
- Added Debian/CentOS/Fedora/OpenSUSE.
- All test use JDK1.8.

@juliogonzalez If you can, please give me comments about OpenSUSE.
@gaul Please check about JDK.

This PR is only tested on Travis CI, not on Github Actions. It will be tested after being merged.
(This PR source code has passed the test on my repository's Github Actions)

Issue #1488, opened 2020-12-02
How about using Github Actions instead of Travis CI?


ggtakec:
I would like to be able to test various operating systems at the same time in the s3fs-fuse test.
Then I was trying to add some OS to my current Travis CI tests.
But I found some problem described below.

### About Travis CI issues
TravisCI is closing travis-ci.org and moving to travis-ci.com.
(Already, we have moved to travis-ci.com.)
At travis-ci.com, not everything is free and it provides `free plan` for public repositories.
This `free plan` allows user to build 1000min/month.
(For public organization, this limit may be higher or even higher. But I have not investigated in detail.)

However, if you run a CI test on multiple OSs, it will take about 20min to build each OS, so it will consume several times that amount(count of OSs).
If you test 5 OSs, it takes 100min for all.
Then If you build 10 times a month, you will exceed this free plan.

And if you exceed that, you will not be able to build individually unless you change to a paid plan instead of a free plan.
(The amount is a minimum of $69 and +$15 if you need macos)
The worst thing for me is that with the free plan I was able to build in 10 concurrent, but I'm limited to running in 1concurrent(or 2).
(In the case of the minimum paid plan, it will be 1 concurrent, and in the case of 5 OS, it will wait 100 minutes.)

Collaborators on s3fs-fuse will consume this build time on travis-ci.com just by forking and building s3fs-fuse.(If they have travis-ci.com settings)
This will happen if collaborators build/test the code in their repository before post the PR.

Perhaps the charge for CI is similar for other CIs such as CircleCI.(I have not confirmed it, but it seems to be the same as far as I can see the documents etc.)

### How about solving it with Github Actions?
I think that s3fs-fuse should pass the CI test of the supported all OSs.
At this time, I think that the above problems(charging for CI) may affect committers and collaborators in no small measure.

Therefore, we are considering running CI(test) on Github Actions.
Currently, I'm trying with 11OS(version), and it can all be done in parallel and seems to be running for free.

### PR for switch to Github Actions
I will soon make PR that uses Github Actions. It has an expanded OS to be tested.(macos10 etc.)
If that PR is merged, I also want to see if I want to stop testing on Travis CI.

gaul:
I think using a different CI is fine as long as FUSE and macOS work.  While Travis was very good 5 years ago when we started using it perhaps other options are better today.

ggtakec:
Yes, that's right.
[...]

PR #1498, opened 2020-12-16
Fixed about Github Actions failure(centos8 and macos)

ggtakec:
#1488 #1489

### Details
There was an error in macos and centos8 of Github Actions added in #1489.
Also, since last week, there was an error in CentOS8, so I fixed these.

#### macos
On macos, I found that the test code running s3fs via stdbuf was in a dead loop(blocked)-like state.
The only workaround I could find was to eliminate via stdbuf and start s3fs directly.
However, since the output of s3fs is processed by stdbuf again, I don't think it will be a problem.
If there is a problem, I used this workaround because it only obscures the log.

Also, macos stdbuf/sed/truncate is not compatible with linux, so I changed them to gnu(coreutils etc).
We've already changed truncate/stdbuf, but I've also added sed.
Instead of replacing this programs in ci.yml(for Github Actions), I've moved those to a test script.

#### centos8
CentOS8 has recently announced CentOS Stream, and it seems that there have been (probably) some changes in line with it.
The `PowerTools` repository name was changed from uppercase to lowercase(`powertools`), which caused the test to fail.
I fixed this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,7 +143,7 @@ jobs:
         run: |
           TAPS="$(brew --repository)/Library/Taps";
           if [ -e "$TAPS/caskroom/homebrew-cask" ]; then rm -rf "$TAPS/caskroom/homebrew-cask"; fi;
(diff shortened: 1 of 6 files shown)


ggtakec:
I will merge this PR.
After that, this will allow the now failing Github Actions to succeed.
@gaul If you find a problem, please let me know.

PR #1502, opened 2020-12-22
fix: miss header when retry

fly3366:
May failed when retry(403、No range eg.).

### Details
Miss header when reset curl handle.


ggtakec (approved):

ggtakec:
@fly3366 Thank you. I think this PR code is correct.
@gaul If there is no problem, merge.

gaul:
Thank you for your contribution @fly3366!

Issue #1159, opened 2019-09-25
Switch to fuse 3.x


Keruspe:
From the release notes of libfuse 2.9.9:

```
Note: This is a maintenance release from the 2.9 branch of libfuse. Users are encouraged to transition to the actively developed libfuse 3.x.
```

gaul:
Can you share what advantages this would provide?  While CentOS 7 EPEL includes fuse3, only Ubuntu 19.04 includes fuse3 which means that both 16.04 and 18.04 LTS users would not be able to use s3fs with fuse3.

juliogonzalez:
CentOS6 will break (but EoL is not that far away).

It would also break:
- SLE12 (all SPs)
- SLE15 (but not SLE15SP1)
- openSUSE Leap 15.0 (but not Leap 15.1)

All **maintained** Fedora versions seem to be have already 3.X

Keruspe:
Most obvious advantage would be to depend on a maintained version of fuse instead of a soon-to-be-deprecated one

gaul:
If there are no functional benefits then I do not see a reason to pursue this.  After Ubuntu 18.04 ends support in 2023 we can re-evaluate this.

juliogonzalez:
As far as I can see at `configure.ac`, s3fs-fuse requires use `2.8.4` as **minimum**

@Keruspe does master compile with 3.X?

If it does, then there's nothing to do at this moment.

If it doesn't, then it's not critical, but at some point (most probably before 2023), this should be addressed, as I am pretty sure newer distributions will start deprecating removing `2.x`, while older distributions will keep it.

ggtakec:
@Keruspe Changes from libfuse 2.x to 3.x include the addition of arguments to some I / F functions.
Therefore, it involves a modification of the s3fs code.
I want these source codes to be compatible with each other (not #ifdef), and to make s3fs source code be one codes.
We will make s3fs correspond to libfuse 3.x, but please take some time including testing and investigation.
Thanks in advance for your kindness.

gaul:
It appears that FUSE 3.6.0 supports 1 MB writes:

https://github.com/libfuse/libfuse/blob/master/ChangeLog.rst#libfuse-360-2019-06-13

Previously s3fs saw a large benefit when moving from 4 KB to 128 KB writes #813.

gaul:
[...]

PR #1503, opened 2020-12-25
Retry failed may load bad page

fly3366:
In MultRead, If Any retry failed. And curl return no OK(such like WRITE ERROR eg.), MultiRead may return 0 and fd may load this page.
But this page is bad. if request include 403、5xx、Partical File(Read Timeout?Connection loss), And next retry is not CURLE_OK, The 403 data may insert into part head and be set loaded(I think we need overwrite it but not loaded).

### Another issue?
1. Multi retry may more large. default is 5(MultiRead) x 5(RequestPerfrom) = 25 count.
2. If offset and size is 0, may request all file without range.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -723,6 +723,12 @@ bool S3fsCurl::SetDnsCache(bool isCache)
     return old;
 }
 
+void S3fsCurl::ResetOffset(S3fsCurl* pCurl)
+{
+    pCurl->partdata.startpos = pCurl->b_partdata_startpos;
+    pCurl->partdata.size     = pCurl->b_partdata_size;
+}
+
 bool S3fsCurl::SetSslSessionCache(bool isCache)
 {
     bool old = S3fsCurl::is_ssl_session_cache;
@@ -1465,6 +1471,7 @@ S3fsCurl* S3fsCurl::ParallelGetObjectRetryCallback(S3fsCurl* s3fscurl)
 
     // duplicate request(setup new curl object)
     S3fsCurl* newcurl = new S3fsCurl(s3fscurl->IsUseAhbe());
+    
     if(0 != (result = newcurl->PreGetObjectRequest(s3fscurl->path.c_str(), s3fscurl->partdata.fd, s3fscurl->partdata.startpos, s3fscurl->partdata.size, s3fscurl->b_ssetype, s3fscurl->b_ssevalue))){
         S3FS_PRN_ERR("failed downloading part setup(%d)", result);
         delete newcurl;
@@ -2243,7 +2250,7 @@ int S3fsCurl::RequestPerform(bool dontAddAuthHeaders /*=false*/)
         curl_easy_setopt(hCurl, CURLOPT_HTTPHEADER, requestHeaders);
 
         // Requests
-        CURLcode curlCode = curl_easy_perform(hCurl);
+        curlCode = curl_easy_perform(hCurl);
 
         // Check result
         switch(curlCode){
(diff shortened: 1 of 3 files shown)


gaul (approved):

gaul:
Could you squash this down to a single commit?

We need a better way to automatically test these situations.  Maybe we can hook up https://github.com/bouncestorage/chaos-http-proxy to the integration tests?  Or add some kind of failpoints to s3fs itself.

gaul:
@ggtakec Please review.  Easier to understand if you add `?w=1` to the URL.

fly3366:
Sorry, may need some lint. please help me.

src/curl.h
```diff
@@ -398,6 +399,7 @@ class S3fsCurl
         headers_t* GetResponseHeaders() { return &responseHeaders; }
         BodyData* GetBodyData() { return &bodydata; }
         BodyData* GetHeadData() { return &headdata; }
+        CURLcode* getCurlCode() { return &curlCode; }
```
gaul, line 402: Why does this return a pointer?

ggtakec: This function should return the value of CURLcode.
Please add `const` like `CURLcode getCurlCode() const {...}`.

ggtakec:
@fly3366 Thank you for the fix.
I think the code is OK, but I don't think getCurlCode() needs to be a pointer(it was pointed by @gaul).
I added the comment in the review.

fly3366:
fixed, 
Q: GetResponseCode() May always return true. May need change?

fly3366:
And sometime, Retry block's startpos may after bad data.

fly3366:
Bad Data like:
<?xml version="1.0" encoding="UTF-8"?>.<Error><Code>500 ..... eg.

fly3366:
The root cause is:
Bad request write file,  some like error code's XML data.
We need reset offset when retry part for overwrite.

gaul (approved):
This makes sense to me.  I am working on some failure injection testing at https://github.com/gaul/s3fs-fuse/tree/chaos-http-proxy but this needs some more work.

fly3366:
Another cause:
1. May 403 when retry, S3 callback call 'Sign Method Error'.
2. Lock may use much cpu and hang in system status.
3. fuse3 use vfs for cache, any plan for support?

src/curl.cpp
```diff
@@ -1856,7 +1862,7 @@ S3fsCurl::S3fsCurl(bool ahbe) :
     LastResponseCode(S3FSCURL_RESPONSECODE_NOTSET), postdata(NULL), postdata_remaining(0), is_use_ahbe(ahbe),
     retry_count(0), b_infile(NULL), b_postdata(NULL), b_postdata_remaining(0), b_partdata_startpos(0), b_partdata_size(0),
     b_ssekey_pos(-1), b_ssetype(sse_type_t::SSE_DISABLE),
-    sem(NULL), completed_tids_lock(NULL), completed_tids(NULL), fpLazySetup(NULL)
+    sem(NULL), completed_tids_lock(NULL), completed_tids(NULL), fpLazySetup(NULL), totalwrite(0)
```
gaul, line 1865: Instead of tracking `totalwrite`, maybe it is better to store the original `startpos` and `size` as consts which `ResetOffset` can use?

gaul:
> * May 403 when retry, S3 callback call 'Sign Method Error'.
> * Lock may use much cpu and hang in system status.

Could you open separate issues for these?  It would also be good to experiment with Chaos HTTP Proxy in #1508 which should allow us to reproduce and verify error conditions.

> * fuse3 use vfs for cache, any plan for support?

#1159 tracks FUSE 3 so it would be good to mention caching as a possible benefit there.  However, I believe FUSE on macOS supports only FUSE 2 so this would be a significant concern.  It might make more sense to optimize network calls then user-kernel roundtrips to improve performance.

ggtakec (approved):

ggtakec:
@fly3366 Thanks for your help.

PR #1506, opened 2020-12-30
Quote | and && to allow passing it to retry

gaul:


test/integration-test-common.sh
```diff
@@ -279,11 +279,11 @@ function stop_s3fs {
     # Retry in case file system is in use
     if [ `uname` = "Darwin" ]; then
         if df | grep -q $TEST_BUCKET_MOUNT_POINT_1; then
-            retry 10 df | grep -q $TEST_BUCKET_MOUNT_POINT_1 && umount $TEST_BUCKET_MOUNT_POINT_1
+            retry 10 df | grep -q $TEST_BUCKET_MOUNT_POINT_1 "&&" umount $TEST_BUCKET_MOUNT_POINT_1
```
gaul, line 282: Huh, this fails on macOS but succeeds on Linux.  Some weird issue with operator precedence?

gaul: Quoted `|`.

test/integration-test-common.sh
```diff
@@ -279,11 +279,11 @@ function stop_s3fs {
     # Retry in case file system is in use
     if [ `uname` = "Darwin" ]; then
         if df | grep -q $TEST_BUCKET_MOUNT_POINT_1; then
-            retry 10 df | grep -q $TEST_BUCKET_MOUNT_POINT_1 && umount $TEST_BUCKET_MOUNT_POINT_1
+            retry 10 df "|" grep -q $TEST_BUCKET_MOUNT_POINT_1 "&&" umount $TEST_BUCKET_MOUNT_POINT_1
```
gaul, line 282: This is still a problem.  It appears that macOS's older bash parses differently?

gaul: Ignoring this for now and just changing the Linux path.

ggtakec (approved):

Issue #1511, opened 2021-01-04
1.88 release


gaul:
1.87 released in August so we should try to release 1.88 soon.  Master has a lot of fixes including some concurrency ones and I would like to merge #1503 as well.  I can run this release if needed.

ggtakec:
I agree launching new, and I create a PR for updating files such as ChangeLogs in preparation for new release.

gaul:
Fixed by a6563211af91996dd091d4e40cc2d8185a600bb8.

juliogonzalez:
RPM submissions:

# openSUSE 

To the devel project, then to factory, then to the relevant versions, most likely only 15.3 and Tumbleweed):
https://build.opensuse.org/request/show/862406

# Fedora 34/rawhide
Bodhi request (autoapproved): https://bodhi.fedoraproject.org/updates/FEDORA-2021-92dd943036

# Fedora 33
Bodhi request: https://bodhi.fedoraproject.org/updates/FEDORA-2021-204c29fde8

# Fedora 32
Bodhi request: https://bodhi.fedoraproject.org/updates/FEDORA-2021-26abcc82ea

# EPEL8
Bodhi request: https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2021-05940c8cba

# EPEL7
Bodhi request: https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2021-8873b594c6

gaul:
Thanks @juliogonzalez!  We really appreciate these packages which keep users up-to-date.

PR #1513, opened 2021-01-04
Updated ChangeLog for release 1.88

ggtakec:
### Relevant Issue (if applicable)
#1511 

### Details
I've listed only the major changes and added them to the `ChangeLog`.
For reference, here is a list of contributors who have posted these PRs.(Since I will not send mentions, `@` is omitted)
- `fly3366`, `bcmyers`, `adamchainz`, `kontrollanten`, `fujitayoshihiko`, `nmeyerhans`, `nkkashyap`, `liuyongqing`, `maresb`

PR #1514, opened 2021-01-05
Updated ChangeLog for release 1.88

gaul:
References #1511.


gaul:
@ggtakec Could we use this PR instead of #1513?  I rewrote, combined, and removed several entries to be more friendly for users.

ggtakec:
@gaul This is better than it, so I will withdraw #1513 and merge this.

PR #1540, opened 2021-01-30
Add timestamps to debug log and cache integrity check

### LINKED ISSUE #1528: Suggestions for improvement
Hello,

last weeks I've troubleshooted a eventual consistency issue with a 3rd party S3 system.

During this time, I noticed these 3 ways to improve s3fs-fuse.

**Add current time to to statistics headline enabled with "-o set_check_cache_sigusr1"**

Please add the current date and time to statistics output.

Current (fdcache.cpp:53):
``` 
------------------------------------------------------------
Check cache file and its stats file consistency
------------------------------------------------------------
```

Suggestion:
``` 
--------------------------------------------------------------------------
Check cache file and its stats file consistency at 2021-01-18 19:19:19 UTC
--------------------------------------------------------------------------
```

**Add error message why attribute update is failing in `put_header()` and log this reason to syslog or at least to debug output:**
```
    if(buf.st_size >= FIVE_GB){
        // multipart
        if(nocopyapi || nomultipart){
            return -EFBIG;    // File too large
        }
```

**What do you think about including the current time in the output of the debug log?**

This would simplify the correlation with external events for long-running jobs. In addition, you can calculate the time span between two different file system operations.

What do you think about this?

Many greetings
Carsten

CarstenGrohmann:
This PR adds timestamps to the debug output as well as to the cache integrity check output.

The new function `S3fsLog::GetCurrentTime()` isn't thread safe.

Closes #1528

### DIFF
--- a/src/s3fs_logger.h
+++ b/src/s3fs_logger.h
@@ -23,6 +23,7 @@
 
 #include <cstdio>
 #include <syslog.h>
+#include <sys/time.h>
 
 //-------------------------------------------------------------------
 // S3fsLog class
@@ -46,6 +47,7 @@ class S3fsLog
         static s3fs_log_level debug_level;
         static FILE*          logfp;
         static std::string*   plogfile;
+        static char           current_time[32];
 
     protected:
         bool LowLoadEnv();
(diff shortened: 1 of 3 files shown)


src/s3fs_logger.h
```diff
@@ -69,6 +71,18 @@ class S3fsLog
                      LEVEL_ERR  == (level & LEVEL_DBG) ? LOG_ERR     : LOG_CRIT );
         }
 
+    static const char* GetCurrentTime()
```
ggtakec, line 74: Please correct the indentation according to the front and back.

ggtakec (changes requested):
Thanks for your PR.
Please correct the indentation.
And A little worrisome is the performance issue.
When outputting as DEBUG MODE, I think that calling gettimeofday every time seems to be slow.
I will confirm it after as well, but it maybe that it is faster to use clock_gettime.
However, this can be fixed again after this PR.

ggtakec (approved):

ggtakec:
@CarstenGrohmann Thanks for your help. It will be merged soon.

Issue #1543, opened 2021-02-02
Run integration tests with larger files


gaul:
s3fs sets `BIG_FILE_LENGTH` to 25 MB which is large enough to trigger multipart upload but lower than other limits like the single file upload size.  This limit also makes sense when using S3Proxy's transient backend which store objects in-memory.  When apache/jclouds#96 merges, s3fs can instead use the filesystem backend which stores objects on-disk.  This will allow the integration tests to use larger objects, e.g., > 5 GB single-part objects.  Related to #1542.

gaul:
With larger files reading from `/dev/urandom` becomes a bottleneck.  We may want to create our own helper which makes pseudo-random data faster.

gaul:
To use the filesystem blobstore to test larger files we will need a new release of S3Proxy to include the jclouds fix.

gaul:
Testing with larger files is proving to be a rich source of bugs.  Some tests require files as large as 11 GB to trigger various conditions which takes up to 15 minutes to run on my desktop.  We could add some flags to s3fs and S3Proxy to use smaller thresholds to trigger conditions like copy part range more easily.

gaul:
Related to #1665, currently blocked on macOS Java xattr compatibility.

PR #1548, opened 2021-02-04
Allow integration tests to use larger files

gaul:
Previously these failed due to `dd` only copying 32 MB from `/dev/urandom`
and exhausting the limited space in `/tmp`.  References #1543.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -363,30 +363,30 @@ function test_rename_before_close {
 function test_multipart_upload {
     describe "Testing multi-part upload ..."
 
-    dd if=/dev/urandom of="/tmp/${BIG_FILE}" bs=$BIG_FILE_LENGTH count=1
-    dd if="/tmp/${BIG_FILE}" of="${BIG_FILE}" bs=$BIG_FILE_LENGTH count=1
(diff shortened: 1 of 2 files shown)


gaul:
It might make more sense to continue using `dd` but split the block size and count to reduce the s3fs debug logs when using `head` and its smaller buffers.  Let me respin this.

test/integration-test-main.sh
```diff
@@ -1047,7 +1047,7 @@ function test_truncate_cache() {
 function test_cache_file_stat() {
     describe "Test cache file stat ..."
 
-    dd if=/dev/urandom of="${BIG_FILE}" bs=${BIG_FILE_LENGTH} count=1
+    dd if=/dev/urandom of="${TEST_TEXT_FILE}" bs=$BIG_FILE_BLOCK_SIZE count=$BIG_FILE_COUNT
```
ggtakec, line 1050: This line is the reason why the test fails in Github Actions.
I think `BIG_FILE` is correct, not` TEST_TEXT_FILE`.

ggtakec (changes requested):
Please fix a line in integration-test-main.sh

test/integration-test-main.sh
gaul: Thanks!  I did not understand why this was failing.

ggtakec (approved):

PR #1565, opened 2021-02-12
set IsExpireTime to true by default

### LINKED ISSUE #1563: Default stat_cache_expire not taking effect
### Additional Information

_example: 1.00_
V1.86 (commit:e2fbcb4)

_example: 2.9.4_
2.9.2

_command result: uname -r_
3.10.0-1127.10.1.el7.x86_64

_command result: cat /etc/os-release_
```
NAME="Red Hat Enterprise Linux Server"
VERSION="7.8 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="7.8"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.8 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.8:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.8
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.8"
```

```
```
```
bucket-name /mountpoint fuse.s3fs _netdev,allow_other,uid=1000,gid=1000 0 0
```
```
```
### Details about issue
The man page specifies that the `-o stat_cache_expire (default is 900)`, but this does not appear to be the case, stat cache seems to last forever.
If the `stat_cache_expire` option is explicitly set to a value, then the caching expiry works.

I believe this may be due to `IsExpireTime` still being set to false by default:
https://github.com/s3fs-fuse/s3fs-fuse/blob/master/src/cache.cpp#L153

lincheney:
fixes #1563

#1563

### Details
Sets the stat cache `IsExpireTime` to `true` by default, so that the default `ExpireTime` actually takes effect.

Tested using:
1. Run s3fs: `s3fs -f bucket /mountpoint ...` *without* the `-o stat_cache_expire` option
2. Create a file in S3: `echo | aws cp - s3://bucket/test-file`
3. Run: `ls -l /mountpoint/test-file`
4. Remove the file from S3: `aws rm s3://bucket/test-file`
5. Wait for more than 15min. (During this time, s3fs will still report the file as existing)
6. Run: `ls -l /mountpoint/test-file` . This should print `No such file or directory`

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -150,7 +150,7 @@ pthread_mutex_t StatCache::stat_cache_lock;
 //-------------------------------------------------------------------
 // Constructor/Destructor
 //-------------------------------------------------------------------
-StatCache::StatCache() : IsExpireTime(false), IsExpireIntervalType(false), ExpireTime(15 * 60), CacheSize(100000), IsCacheNoObject(false)
+StatCache::StatCache() : IsExpireTime(true), IsExpireIntervalType(false), ExpireTime(15 * 60), CacheSize(100000), IsCacheNoObject(false)
 {
     if(this == StatCache::getStatCacheData()){
         stat_cache.clear();


gaul:
Thank you for your contribution @lincheney!

Issue #1541, opened 2021-01-31
readdir parallel GET and HEAD


gaul:
`s3fs_readdir` calls `list_bucket` which serially issues `ListBucket`, once per every 1,000 objects.  After completion it calls `readdir_multi_head` which calls `HeadObject` in parallel, one request per object.  Instead s3fs could overlap the calls to `ListBucketRequest` and `PreHeadRequest` to improve performance for large directories.

Also I wonder if there is an opportunity to call `filler` while requests are still completing to allow the application to make some progress.  This might require a different locking mechanism.

ggtakec:
@gaul I posted a PR for this issue.
However, there seems to be no major performance change.
Other than this change, there may be another improvement, so if the PR is merged, I will try to deal with it.

ggtakec:
@gaul I also added PR for #1569.
This PR sets EPERM instead of EIO for 400 HTTP errors, and also registers EPERM objects in the Stats cache.
This will prevent you from sending an error HEAD request every time, which can be expected to improve performance.

PR #1567, opened 2021-02-12
Enhanced parallel processing for list_bucket and multihead request

ggtakec:
#1541 

### Current
The process of readdir is to first create a list of files(objects) in the directory with the list_bucket function.(ListBucket Request)
The maximum value of ListBucket Request is 1000 files, and the list is created in units of 1000 files.
For directories with over 1000 files, list them all.
After making list, s3fs sends HEAD requests for all the files to get the Stats information.(readdir_multi_head)
Finally, after all the Stats information is complete, it returns the file list to FUSE, and it will also be registered in the Stats cache.

### Change
I changed the above process.
First, calling the list_bucket function is the same.
ListBucket Request lists files in units of 1000 as before.

Changed to start sending HEAD requests after listing 1000 instead of waiting for all listings.
This process is done in a separate thread, and also the HEAD request itself runs in a separate thread.
This has changed to list up files and executing HEAD requests in parallel.

It also registers the Stats  information in the Stats cache as soon as each HEAD request is completed.
This means updating the Stats cache while processing other HEAD requests and listing files.
If the Stats cache is updated while waiting for a HEAD request and its file information becomes hit, the HEAD request will not be sent.
(It's not clear if this situation exists, but it would be beneficial if the Stats cache was updated by another process.)

### About perfomance
After all, due to the large number of individual HEAD requests, the processing time of HEAD requests is longer than the processing time of ListBucket.
Therefore, I think the performance will not change much.
(There was not much difference in my results.)
However, I think that this PR should be merged for future changes.
(Because the multi-request logic can be used for other than HEAD request)

#### 400 HTTP response code
In the HEAD request, I noticed that accessing the HEAD request to an object such as SSE returns an error 400 from S3.
In the processing of S3fsMultiCurl, when this number 400 was received, it was retrying.
Changed this retry process so that it is not performed in the case of HEAD request.
I think this issue was previously pointed out in the Issue. (I couldn't find which issue it was.)

~~#### About libcrypt and multi-threading~~
~~When using openssl(libcrypt) from multithreading, it may get a `double free` error on OSX.~~
~~To avoid this, the insertV4(2)Headers function was changed exclusively controlled by mutex.~~
~~I wrote a little more information on gist at https://gist.github.com/ggtakec/a743affecf153e78f6b5d74e2bb1fcd5.~~

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -81,6 +81,7 @@ const long       S3fsCurl::S3FSCURL_RESPONSECODE_FATAL_ERROR;
 const int        S3fsCurl::S3FSCURL_PERFORM_RESULT_NOTSET;
 pthread_mutex_t  S3fsCurl::curl_warnings_lock;
 pthread_mutex_t  S3fsCurl::curl_handles_lock;
+pthread_mutex_t  S3fsCurl::cryptfunc_lock;
 S3fsCurl::callback_locks_t S3fsCurl::callback_locks;
 bool             S3fsCurl::is_initglobal_done  = false;
 CurlHandlerPool* S3fsCurl::sCurlPool           = NULL;
@@ -161,6 +162,9 @@ bool S3fsCurl::InitS3fsCurl()
     if(0 != pthread_mutex_init(&S3fsCurl::callback_locks.ssl_session, &attr)){
         return false;
     }
+    if(0 != pthread_mutex_init(&S3fsCurl::cryptfunc_lock, &attr)){
+        return false;
+    }
     if(!S3fsCurl::InitGlobalCurl()){
         return false;
     }
@@ -199,6 +203,9 @@ bool S3fsCurl::DestroyS3fsCurl()
         result = false;
     }
     if(!S3fsCurl::DestroyGlobalCurl()){
+        result = false;
+    }
+    if(0 != pthread_mutex_destroy(&S3fsCurl::cryptfunc_lock)){
         result = false;
     }
     if(0 != pthread_mutex_destroy(&S3fsCurl::callback_locks.dns)){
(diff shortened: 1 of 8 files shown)


ggtakec:
@gaul `About libcrypt and multi-threading` part may have been wrong, so double check.
Please wait until the confirmation is completed.

ggtakec:
The `About libcrypt and multi-threading` part was wrong, so I deleted that part.(Please squash when merging)

gaul:
This is a tricky pull request and it will take me some time to review.  I ran `ls` on a bucket in us-west-2 with 3,000 objects from Japan and performance is similar.  However s3fs uses a lot more CPU time.  Perhaps some accidental overhead slows the algorithm down?

Without this PR:

```
$ time s3fs gaultest mnt/ -f
...
user    0m5.792s
sys     0m1.919s
```

with this PR:

```
$ time src/s3fs gaultest mnt/ -f
...
user    0m23.325s
sys     0m3.044s
```

ggtakec:
I checked it in a directory of about 1000-2000, but your result was not as good as I expected.
Synchronous object contention may be slowing performance.
I will investigate more to see if it can be fixed.
So please leave this PR.

ggtakec:
@gaul 
I tried it in a directory with 2000 objects(all file size is 0).
The code for this PR has been modified with 1570 and 1572 to achieve the same conditions.
The only difference from master is the modification code for this PR.
It has been executed twice, but the first time is just after startup, the state without Stats cache, and the second time it is in the cached state.

### Current master( 4da56ac )
```
$ src/s3fs <mybucket> /mnt/s3fs -o allow_other,nonempty,enable_noobj_cache,enable_content_md5,use_cache=/tmp,multireq_max=50,endpoint=ap-northeast-1,dbglevel=err -f &
$ cd /mnt/s3fs/test
$ time ls -la >/dev/null
real    0m10.000s
user    0m0.005s
sys     0m0.020s

$ time ls -la >/dev/null
real    0m2.429s
user    0m0.017s
sys     0m0.005s
```

### This PR code( +1570, +1572)
```
$ src/s3fs <mybucket> /mnt/s3fs -o allow_other,nonempty,enable_noobj_cache,enable_content_md5,use_cache=/tmp,multireq_max=50,endpoint=ap-northeast-1,dbglevel=err -f &
$ cd /mnt/s3fs/test
$ time ls -la >/dev/null
real    0m8.513s
user    0m0.009s
sys     0m0.017s

$ time ls -la >/dev/null
real    0m2.313s
user    0m0.009s
sys     0m0.013s
```

After this, increase the number of files and check.
Can you tell me your startup options?
I would like to try with the same option.

ggtakec:
@gaul I tested with 2000/4000/8000 files.
https://gist.github.com/ggtakec/1b8725ca1021659f08c391d877ae7d67
As a result, no performance degradation has been detected.

Does your test include reading and writing files(in other words, other than ListBucket, Head requests)?
If these are included, this PR may be affecting other factors/logic, so I will investigate it.

gaul:
I tested this again with a 3,000 object bucket.

Without this PR:

```
$ time s3fs gaultest mnt/ -f
...
user    0m6.622s
sys     0m1.949s
```

with this PR:

```
$ time src/s3fs gaultest mnt/ -f
...
user    0m23.408s
sys     0m3.219s
```

The real time of `ls` is similar for both.

ggtakec:
@gaul Thanks for investigating the results of the poor performance.
This PR wasn't ready for merging yet, so I withdraw this.

PR #1577, opened 2021-02-16
Minor improvements for tests

CarstenGrohmann:
This PR contains a correction of a typo in an error message and allows changing the temporary directory used for tests.

### DIFF
--- a/test/test-utils.sh
+++ b/test/test-utils.sh
@@ -31,7 +31,7 @@ TEST_DIR=testdir
 ALT_TEST_TEXT_FILE=test-s3fs-ALT.txt
 TEST_TEXT_FILE_LENGTH=15
 BIG_FILE=big-file-s3fs.txt
-TEMP_DIR="/var/tmp"
+TEMP_DIR=${TMPDIR:-"/var/tmp"}
 # /dev/urandom can only return 32 MB per block maximum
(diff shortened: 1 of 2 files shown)


test/test-utils.sh
```diff
@@ -31,7 +31,7 @@ TEST_DIR=testdir
 ALT_TEST_TEXT_FILE=test-s3fs-ALT.txt
 TEST_TEXT_FILE_LENGTH=15
 BIG_FILE=big-file-s3fs.txt
-TEMP_DIR="/var/tmp"
+TEMP_DIR=${TMPDIR:-"/var/tmp"}
```
gaul, line 34: Would it be more clear to use the same variable:

```
TEMP_DIR="${TEMP_DIR:-/var/tmp}"
```

gaul (commented):
Thanks for looking at the tests!  If you are testing big files, the current S3Proxy and its configuration stores objects in-memory so you can run out of RAM.  Therefore you might want to test against AWS or configure S3Proxy to use the filesystem provider and compile the latest version which has some fixes.

test/test-utils.sh
CarstenGrohmann: Of course I can change this.

I decided to use TMPDIR because this variable is well-known and specified e.g. in the Single UNIX Specification. As admin I would try TMPDIR first.


Change it or leave it?

gaul:

Issue #1453, opened 2020-10-15
Corruption following rename()


init-js:
#### Version of s3fs being used (s3fs --version)

```
Amazon Simple Storage Service File System V1.87 (commit:194262c) with OpenSSL
Copyright (C) 2010 Randy Rizun <rrizun@gmail.com>
License GPL2: GNU GPL version 2 <https://gnu.org/licenses/gpl.html>
```

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)

`2.9.9`

#### Kernel information (uname -r)

`4.14.177-139.254.amzn2.x86_64`

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)

```
NAME="Ubuntu"
VERSION="20.04.1 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.1 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

#### s3fs command line used, if applicable
```
s3fs my_bucket /mnt -d -d -f -o use_cache=/tmp/s3fs-cache -o enable_noobj_cache -o endpoint=us-west-2 -o nomultipart -o use_path_request_style -o iam_role=auto -o del_cache
```
#### /etc/fstab entry, if applicable
```
```
#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
_if you execute s3fs with dbglevel, curldbg option, you can get detail debug messages_
```
```
### Details about issue

It feels like a caching bug is causing s3fs to upload a file filled with zeroes if you trigger a chmod/chown just after a rename.
At the end of the following program, "B" will have the correct length, but its contents will be nothing but nil bytes `\0`.
(This is assuming /mnt is the fuse mountpoint.)

```
#!/bin/bash
# s3fs.fuse mounted there
mntpoint=/mnt/

tmp="$mntpoint/hello_$RANDOM$RANDOM"
target="$mntpoint/my_target"

[...]

PR #1564, opened 2021-02-11
Test updating the metadata of an external object

gaul:
References #1453.

PR #1579, opened 2021-02-19
Fixed a bug that filling NULL bytes when changing the attribute after renaming

ggtakec:
#1453 

### Details
The rename logic internally copies the file and deletes the original file.
The cause of the #1453 was that s3fs was updating mtime immediately after this copy, at which time a new cache file was created and the contents of the file were null.

In this PR, updates such as mtime are added to the cache file before copying.
And leave the cache file open and copy (put_header).
After copying, rename the cache file and delete the original file.

Keeping the cache file open is necessary because fdcache cannot be renamed unless it is open and new cache files are not created.

And, although it was also the cause of the problem, I accidentally deleted the cache of the destination file.
This is also fixed. (Prevents unnecessary downloads and improves efficiency)

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1234,9 +1234,10 @@ static int s3fs_symlink(const char* _from, const char* _to)
 
 static int rename_object(const char* from, const char* to, bool update_ctime)
 {
-    int result;
+    int         result;
     std::string s3_realpath;
-    headers_t meta;
+    headers_t   meta;
+    struct stat buf;
 
     S3FS_PRN_INFO1("[from=%s][to=%s]", from , to);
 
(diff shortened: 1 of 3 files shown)


gaul:
Does this fix the failing test in #1564?  I still see this on my system:

```
test-s3fs.txt /dev/fd/63 differ: byte 1, line 1
test_update_metadata_external_object failed
```

ggtakec:
@gaul Please review new fixed code.
The reason for the failure of the test being used in #1564 was due to a code other than `rename`.
However, the root cause is related to `put_header` and a similar fix is needed.
Fixed the PR code and fixed a similar bug with `chmod`/`chown`/`utimens`.
Also, since the #1564 test had only one pattern of chmod, I added a test to test `chmod`/`chown`/`utimens` with a `small file` and a `large file`.
The new modified code confirms that these tests have passed.

PR #1586, opened 2021-02-23
Bypass an OSX test(no different from other OS) which takes very time consuming

ggtakec:
 #1579 #1582

### Details
test_update_metadata_external_large_object The test repeats uploading and downloading large objects several times.
It takes a long time to test, especially because it is not affected by the stats (noobj) cache.
Before adding this test, OSX took longer to run the test than other OSs, but after adding this test, it's up to 10 minutes slower.
The test itself is considered OS-independent and suggests bypassing in the case of OSX.
In addition, after improving the performance of s3fs itself, it is a condition to review this test and restore it.

I think it should be merged, but please give us your opinion.


test/integration-test-main.sh
```diff
@@ -1445,7 +1445,14 @@ function add_all_tests {
...
     add_tests test_read_external_object
     add_tests test_update_metadata_external_small_object
-    add_tests test_update_metadata_external_large_object
+    if ! uname | grep -q Darwin; then
+        # [NOTE]
+        # This test is very time consuming on OSX and will not run.
+        # And this test should be no different between OSX and
+        # other OSs. so skip this test on OSX.
+        #
+        add_tests test_update_metadata_external_large_object
```
gaul, line 1454: Why is this only slow on macOS?

ggtakec: Not only is this process heavy, but it probably runs slower on macOS than on other Linux.
(It seems that it is a difference in the specifications of Runner's HOST of Github Actions.)
I wanted to reduce the latency in the case of repeated builds, so I thought about bypassing this test that I added last.
But, even if we bypass this part, it will take longer than Linux.
If the test should pass equally on all operating systems, this PR will be withdrawn.

gaul: Strange but let's investigate this more later.

Issue #1589, opened 2021-02-24
Look at pjdfstest


gaul:
Might help find some POSIX incompatibilities:

https://github.com/pjd/pjdfstest

juicefs passes these tests so many of them should work on other network filesystems like s3fs.

gaul:
```
Test Summary Report
-------------------
/home/gaul/work/pjdfstest/tests/chmod/00.t          (Wstat: 0 Tests: 119 Failed: 38)
  Failed tests:  32-34, 36-38, 41-44, 46-48, 51, 74-81, 87
                91, 95, 98-105, 107, 112-114, 116
/home/gaul/work/pjdfstest/tests/chmod/01.t          (Wstat: 0 Tests: 17 Failed: 6)
  Failed tests:  8-13
/home/gaul/work/pjdfstest/tests/chmod/02.t          (Wstat: 0 Tests: 5 Failed: 1)
  Failed test:  5
/home/gaul/work/pjdfstest/tests/chmod/03.t          (Wstat: 0 Tests: 5 Failed: 4)
  Failed tests:  1-4
/home/gaul/work/pjdfstest/tests/chmod/05.t          (Wstat: 0 Tests: 14 Failed: 9)
  Failed tests:  4-12
/home/gaul/work/pjdfstest/tests/chmod/07.t          (Wstat: 0 Tests: 25 Failed: 20)
  Failed tests:  4-23
/home/gaul/work/pjdfstest/tests/chmod/11.t          (Wstat: 0 Tests: 109 Failed: 44)
  Failed tests:  26-28, 30-31, 33-36, 38-39, 41, 52-53, 55-56
                63-66, 73-76, 79-81, 83-86, 88-91, 93-96
                98, 103-106
/home/gaul/work/pjdfstest/tests/chmod/12.t          (Wstat: 0 Tests: 14 Failed: 6)
  Failed tests:  3-4, 7-8, 11-12
/home/gaul/work/pjdfstest/tests/chown/00.t          (Wstat: 0 Tests: 1323 Failed: 867)
  Failed tests:  5-6, 8, 21-22, 24, 37-38, 40, 50-54, 56-58
                61-70, 72-74, 77-81, 85-86, 88, 105-108
                110, 114-116, 118-120, 127-130, 135-138
                140, 144-146, 148-150, 157-160, 165-168
                170, 174-176, 178-180, 187-190, 192-198
                200-202, 204-206, 208-210, 213-228, 230-232
[...]

Issue #1551, opened 2021-02-07
Look at EACCES vs. EPERM


gaul:
`S3fsCurl::RequestPerform` maps HTTP 403 to `EPERM` but `EACCES` may be more correct:

https://stackoverflow.com/questions/35023667/difference-between-eacces-and-eperm

We should audit other uses of this errno.

PR #1617, opened 2021-04-11
Return EACESS when lacking required executable bit

gaul:
This makes the check consistent with read and write.  Found via
pjdfstests.  References #1551.  References #1589.


ggtakec (approved):

ggtakec:
The MacOS build will fail because the version of cppcheck has been upgraded, not because of this PR.
cppcheck errors have nothing to do with this PR. I will issue a modified PR later.

PR #1621, opened 2021-04-12
Fixed codes for cppcheck 2.4.1

ggtakec:
n/a

### Details
Fixed errors in cppchek 2.4.1(currently MacOS only).

### DIFF
--- a/src/addhead.cpp
+++ b/src/addhead.cpp
@@ -77,10 +77,10 @@ bool AdditionalHeader::Load(const char* file)
     std::string line;
     ADDHEAD *paddhead;
     while(getline(AH, line)){
-        if('#' == line[0]){
+        if(line.empty()){
             continue;
         }
-        if(line.empty()){
+        if('#' == line[0]){
             continue;
         }
         // load a line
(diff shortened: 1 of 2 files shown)


src/curl.cpp
```diff
@@ -3563,28 +3563,30 @@ int S3fsCurl::CompleteMultipartPostRequest(const char* tpath, const std::string&
     }
 
     // make contents
-    std::string postContent;
-    postContent += "<CompleteMultipartUpload>\n";
+    std::string* pPostContent = new std::string();
```
gaul, line 3566: I don't think this is right -- it just more complicated so cppcheck cannot see the dangling pointer.  See #1622 for an alternate approach.

src/addhead.cpp
```diff
@@ -77,10 +77,10 @@ bool AdditionalHeader::Load(const char* file)
     std::string line;
     ADDHEAD *paddhead;
     while(getline(AH, line)){
-        if('#' == line[0]){
+        if(line.empty()){
```
gaul, line 80: These changes look good.

src/curl.cpp
ggtakec: Thank you, the error handling of dnglingLifetime was incorrect.
I have corrected it, so please reconfirm it.

PR #1630, opened 2021-04-23
Fail CheckBucket when S3 returns PermanentRedirect

gaul:
Previously s3fs allowed mounting but all operations returned `EIO`.
References #693.


gaul:
@ggtakec I don't believe that is is possible to autodetect this the same way we do for `AuthorizationHeaderMalformed`.

ggtakec:
@gaul I agree with you, and I think it's difficult for the path request style.

PR #1640, opened 2021-04-29
Do not create zero-byte object when creating file

### LINKED ISSUE #1013: S3FS generating double S3 ObjectCreated event notifications
@ggtakec in relation to https://github.com/s3fs-fuse/s3fs-fuse/issues/427#issuecomment-478217372_ ...

we have application architecture that needs to support zero byte files and discovered this issue during testing. I am sure this has been considered but will ask anyway ... Could a configuration switch that one could set that would allow a choice to create the initial file with some sort of filename pattern? we could leverage in the S3 Event Notification that we could use to ignore the initial file creation, similar to an event filter proposed earlier to handle zero byte files?

IE.

1. Create initial file with [FILE_NAME] 'S3SF_TEMP' prefix, suffix, or file extension.
2. Implementors could create a S3 event filter to ignore the 'S3SF_TEMP' prefix, suffix, or file extension.
3. Open the file and write to the file.
4. Rename the file?

_Originally posted by @ggtakec in https://github.com/s3fs-fuse/s3fs-fuse/issues/427#issuecomment-478217372_

gaul:
Previously s3fs created this object to store metadata and overwrote it
when flushing.  This prevented use with object stores which do not
allow overwrites like HDS.  Instead only create an in-memory
representation which reduces the time to create small files.
Fixes #1013.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -970,20 +970,27 @@ static int s3fs_create(const char* _path, mode_t mode, struct fuse_file_info* fi
     }else if(0 != result){
         return result;
     }
-    result = create_file_object(path, mode, pcxt->uid, pcxt->gid);
-    StatCache::getStatCacheData()->DelStat(path);
-    if(result != 0){
-        return result;
+
+    time_t now = time(NULL);
+    headers_t meta;
+    meta["Content-Length"] = "0";
+    meta["x-amz-meta-uid"]   = str(pcxt->uid);
+    meta["x-amz-meta-gid"]   = str(pcxt->gid);
+    meta["x-amz-meta-mode"]  = str(mode);
+    meta["x-amz-meta-atime"] = str(now);
+    meta["x-amz-meta-mtime"] = str(now);
+    meta["x-amz-meta-ctime"] = str(now);
+    if(!StatCache::getStatCacheData()->AddStat(path, meta)){
+        return -EIO;
     }
 
     AutoFdEntity autoent;
     FdEntity*    ent;
-    headers_t    meta;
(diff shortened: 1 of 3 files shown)


src/s3fs.cpp
```diff
@@ -970,20 +970,25 @@ static int s3fs_create(const char* _path, mode_t mode, struct fuse_file_info* fi
...
     AutoFdEntity autoent;
     FdEntity*    ent;
-    headers_t    meta;
-    get_object_attribute(path, NULL, &meta, true, NULL, true);    // no truncate cache
     if(NULL == (ent = autoent.Open(path, &meta, 0, -1, false, true))){
         StatCache::getStatCacheData()->DelStat(path);
         return -EIO;
     }
+    ent->pagelist.SetPageLoadedStatus(0, 1, PageList::PAGE_LOAD_MODIFIED);
+    ent->is_meta_pending = true;
```
gaul, line 991: This PR is a work-in-progress but I wanted to get your feedback to see if this approach is valid.  I understand that these should modify the fields more cleanly but I did not see such a mechanism?

ggtakec: There is no method to directly operate PageList, so if you want to do such an operation, I thinnk it is better to prepare a method of FdEntity.

ggtakec:
@gaul It seems that this approach can eliminate the initially created 0 byte file.
I think that when s3fs opens(creates) a new file(not exist), write no bytes, and close it, we need to make sure that the file(0 byte) is created.

src/s3fs.cpp
gaul: Done.

ggtakec (approved):

PR #1641, opened 2021-04-29
Update CI to latest Fedora versions

gaul:


ggtakec:
Fedora33 seems to give the following error.
```
.github/workflows/linux-ci-helper.sh: line 199: pip3: command not found
.github/workflows/linux-ci-helper.sh: line 200: pip3: command not found
```
It seems that awscli was not installed because pip3 is not installed.

gaul:
The original PR lacked python-pip3 for Fedora 33 but CI passes now?

Also I notice that 31 and 32 still show up in the list as required -- is there some other configuration for this?

ggtakec:
It seems that pip3 exists in Fedora34(32/31), but for some reason only Fedora33 has an error.
This is an error on the line below.
https://github.com/s3fs-fuse/s3fs-fuse/blob/master/.github/workflows/linux-ci-helper.sh#L168
(We should have checked for errors here, I'll add it later)
I think pip3 doesn't exist in the Fedora 33 image(why?), but it's still unconfirmed.

Fedora 31/32 is still listed in the settings(`Status checks found in the last week for this repository`) in Github Setting->Branches page.
After this PR has been merged, remove Fedora 31/32 and check 33/34 items instead.

gaul:
I'm not sure what is wrong with Fedora 33 but I removed it since 34 is the latest version.

ggtakec:
Oddly enough, I just ran Fedora 33/34 only in Actions, but I didn't get any errors when installing pip3.
(There may have been something wrong with the image or Runner.)

I think that the code that has been modified only to Fedora 34 is fine.
Thanks, I merge this.

ggtakec (approved):

ggtakec:
I changed `Setting->Branches` for set fedora34 instead of fedora31/32.

gaul:
I think it's OK to test only the latest version of Fedora and maybe Ubuntu.  The important thing is testing the LTS distros which have weird dependencies.

I noticed that Ubuntu 16.04 goes end-of-life soon.  Maybe we should replace this with 21.04?

ggtakec:
I agree with adding 21.04 to CI.
(I think it's better to set it as `Required` after 16.04 is no longer supported)

PR #1650, opened 2021-05-06
Fixed make_random_string test code being slow on macos

ggtakec:
#1586

### Details
I noticed that the make_random_string function in my test code is very slow on macos.
When I executed `cat / dev / urandom | base64.....`, it used a lot of CPU and memory, and it took a lot of execution time.
This was also the reason why #1586 bypassed the test_update_metadata_external_large_object test.
With this fix, even if you re-add the bypassed test, the test time will be shorter.

### DIFF
--- a/test/test-utils.sh
+++ b/test/test-utils.sh
@@ -50,10 +50,12 @@ if [ `uname` = "Darwin" ]; then
     export STDBUF_BIN="gstdbuf"
     export TRUNCATE_BIN="gtruncate"
     export SED_BIN="gsed"
+    export BASE64_BIN="gbase64"
 else
     export STDBUF_BIN="stdbuf"
     export TRUNCATE_BIN="truncate"
     export SED_BIN="sed"
+    export BASE64_BIN="base64"
 fi
 export SED_BUFFER_FLAG="--unbuffered"
 
@@ -343,8 +345,10 @@ function make_random_string() {
     else
         END_POS=8
     fi
-    RANDOM_STR=`cat /dev/urandom | base64 | sed 's#[/|+]##g' | head -1 | cut -c 1-${END_POS}`
-    echo "${RANDOM_STR}"
+
+    ${BASE64_BIN} --wrap=0 < /dev/urandom | tr -d /+ | head -c ${END_POS}
+
+    return 0
 }
 
 #
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -1460,14 +1460,7 @@ function add_all_tests {
     add_tests test_external_modification
     add_tests test_read_external_object
     add_tests test_update_metadata_external_small_object
-    if ! uname | grep -q Darwin; then
-        # [NOTE]
-        # This test is very time consuming on OSX and will not run.
-        # And this test should be no different between OSX and
-        # other OSs. so skip this test on OSX.
-        #
-        add_tests test_update_metadata_external_large_object
-    fi
+    add_tests test_update_metadata_external_large_object
     add_tests test_rename_before_close
     add_tests test_multipart_upload
     add_tests test_multipart_copy


test/test-utils.sh
```diff
@@ -343,7 +343,16 @@ function make_random_string() {
...
+
+    # [NOTE]
+    # The number of bytes to be read is "((($ 1) / 2 + 1) / 4 + 1) * 4". 
+    # Since it is displayed in units of 4 bytes with the od command,
+    # it will be filled with 0 if it is not a 4-byte boundary.
+    # To receive the result in hexadecimal, you need at least half the
+    # number of bytes required.
+    #
+    TMP_LOAD_BYTES=$((((${END_POS}/2 + 1)/4 + 1)*4))
+    RANDOM_STR=$(od -An -N${TMP_LOAD_BYTES} -tx /dev/urandom | tr -d [:space:] | cut -c 1-${END_POS})
```
gaul, line 355: Linux supports `od --width` but unfortunately macOS does not.

gaul (approved):
I looked at a few options including `xxd` but it seems there are few good options between Linux and macOS.

test/test-utils.sh
```diff
@@ -343,7 +343,16 @@ function make_random_string() {
     else
         END_POS=8
     fi
-    RANDOM_STR=`cat /dev/urandom | base64 | sed 's#[/|+]##g' | head -1 | cut -c 1-${END_POS}`
```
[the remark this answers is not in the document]
ggtakec: It turns out that the original code doesn't use GNU base64.
The test code is using `base64` which is the initial installation of macOS, then that base64 can not wrap words and the process is clogged with sed.
I will modify this PR to use GNU base64(gbase64) that comes with coreutils.

test/test-utils.sh
ggtakec: This part is undone.

test/test-utils.sh
```diff
@@ -343,7 +343,16 @@ function make_random_string() {
...
+    # [NOTE]
+    # The number of bytes to be read is "((($ 1) / 2 + 1) / 4 + 1) * 4". 
+    # Since it is displayed in units of 4 bytes with the od command,
+    # it will be filled with 0 if it is not a 4-byte boundary.
+    # To receive the result in hexadecimal, you need at least half the
+    # number of bytes required.
+    #
+    TMP_LOAD_BYTES=$((((${END_POS}/2 + 1)/4 + 1)*4))
+    RANDOM_STR=$(od -An -N${TMP_LOAD_BYTES} -tx /dev/urandom | tr -d [:space:] | cut -c 1-${END_POS})
     echo "${RANDOM_STR}"
```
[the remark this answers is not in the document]
ggtakec: Thanks, I'll update code soon.

test/test-utils.sh
```diff
@@ -343,8 +345,15 @@ function make_random_string() {
...
     fi
-    RANDOM_STR=`cat /dev/urandom | base64 | sed 's#[/|+]##g' | head -1 | cut -c 1-${END_POS}`
-    echo "${RANDOM_STR}"
+
+    # [NOTE]
+    # If character substitution occurs, it will be less than the specified
+    # number of characters, but the value is only used for the file name
+    # now, so it is acceptable.
+    #
+    ${BASE64_BIN} --wrap=0 < /dev/urandom | head -c ${END_POS} | sed -e 's#[/|+]##g'
```
gaul, line 354: Does this return too few characters?  You should reverse `head` and `sed`:

```sh
${BASE64_BIN} --wrap=0 < /dev/urandom | sed -e 's#[/|+]##g' | head -c ${END_POS}
```

But it is probably better to replace `sed` with `tr` which likely handles long lines better:

```sh
${BASE64_BIN} --wrap=0 < /dev/urandom | tr -d /+ | head -c ${END_POS}
```

ggtakec: Thank you, it seems better to put tr in between as you point.
(Unlike sed, I comfirmed that tr is not blocked on macOS.)

Issue #1643, opened 2021-05-01
File creation issues 9 HTTP requests sequentially


gaul:
s3fs probes the for object existence sequentially which hurts performance:

```
HEAD /tmp/1 HTTP/1.1
HEAD /tmp/1/ HTTP/1.1
HEAD /tmp/1_%24folder%24 HTTP/1.1
GET /?delimiter=/&max-keys=2&prefix=tmp/1/ HTTP/1.1
HEAD /tmp/1 HTTP/1.1
HEAD /tmp/1/ HTTP/1.1
HEAD /tmp/1_%24folder%24 HTTP/1.1
GET /?delimiter=/&max-keys=2&prefix=tmp/1/ HTTP/1.1
PUT /tmp/1 HTTP/1.1
```

Setting `-o notsup_compat_dir` helps but s3fs can make this much faster by issuing requests in parallel.  The logic in `chk_dir_object_type` is a bit twisty though.

PR #1651, opened 2021-05-08
Issue parallel requests in get_object_attribute

gaul:
s3fs must issue HEAD requests for path, path/, and the legacy
path_$folder$ to return an object's attributes.  Previously it issued
these serially which minimized costs but caused poor performance.
Instead issue these in parallel since HeadObject is cheap.  Reduces
getattr time of non-existent files from 732 ms to 369 ms.
References #1643.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -122,7 +122,7 @@ static int check_parent_object_access(const char* path, int mask);
 static int get_local_fent(AutoFdEntity& autoent, FdEntity **entity, const char* path, bool is_load = false);
 static bool multi_head_callback(S3fsCurl* s3fscurl);
 static S3fsCurl* multi_head_retry_callback(S3fsCurl* s3fscurl);
-static int readdir_multi_head(const char* path, const S3ObjList& head, void* buf, fuse_fill_dir_t filler);
+static int readdir_multi_head(const char* path, const s3obj_list_t& head, void* buf, fuse_fill_dir_t filler);
 static int list_bucket(const char* path, S3ObjList& head, const char* delimiter, bool check_content_only = false);
 static int directory_empty(const char* path);
 static int rename_large_object(const char* from, const char* to);
@@ -339,6 +339,15 @@ static int remove_old_type_dir(const std::string& path, dirtype type)
     return 0;
 }
 
+int get_object_attribute_fill_dir(void *buf, const char *name, const struct stat *stbuf, off_t off)
+{
+    if(stbuf != NULL){
+        struct stat* pstat = reinterpret_cast<struct stat*>(buf);
+        *pstat = *stbuf;
+    }
+    return 0;
+}
+
 //
 // Get object attributes with stat cache.
 // This function is base for s3fs_getattr().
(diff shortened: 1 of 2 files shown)


gaul:
@ggtakec This may need some more work but I wanted to see what you think about the general approach?  It looks like there is some old unnecessary code so removing more may simplify this.

ggtakec:
At the moment, there was a reason for sending HEAD requests sequentially.
The number of requests is large because s3fs checks the three types of path, path/, and path_$folder$.
Ultimately, the cost of the number of requests will be borne by the user.
So s3fs sent requests sequentially and reduced the number.

However, comparing current cost with performance, I think it's better to resolve the performance dissatisfaction.
So I would like to actively support this PR.

Before merging, there are a few things I would like to check, so please take a look at the comments.

src/s3fs.cpp
```diff
@@ -393,78 +402,24 @@ static int get_object_attribute(const char* path, struct stat* pstbuf, headers_t
...
-        // the response without modifiying is registered in the
-        // cache, the file permission will be 0644(umask dependent)
-        // because the meta header does not exist.
-        // Thus, set the mode of 0000 here in the meta header so
-        // that s3fs can print a permission error when the file
-        // is actually accessed.
-        // It is better not to set meta header other than mode,
-        // so do not do it.
-        //
-        (*pheader)["x-amz-meta-mode"] = str(0);
```
ggtakec, line 419: Is it possible to confirm in the same situation as this?

If there is an object that causes an access permission error (returned from the S3 server side) and the cache is not registered, access to that object will send a HEAD request every time.
Maybe it should be handled by the get_object_attribute_fill_dir function.

src/s3fs.cpp
```diff
@@ -393,78 +402,24 @@ static int get_object_attribute(const char* path, struct stat* pstbuf, headers_t
...
-                // found "no dir object".
-                strpath  += "/";
-                *pisforce = true;
-                result    = 0;
-            }
-        }
+    s3obj_list_t headlist;
+    std::string parent;
+    std::string child;
+    size_t pos = strpath.rfind('/');
```
ggtakec, line 408: If strpath is `path /`, is there any effect in the subsequent processing?
(Will it be `path //` or `path / _ $ folder $`?)

src/s3fs.cpp
```diff
@@ -492,11 +456,7 @@ static int get_object_attribute(const char* path, struct stat* pstbuf, headers_t
     // (ex. getxattr() is called while writing to the opened file.)
     //
     if(add_no_truncate_cache || 0 != StatCache::getStatCacheData()->GetCacheSize()){
-        // add into stat cache
-        if(!StatCache::getStatCacheData()->AddStat(strpath, (*pheader), forcedir, add_no_truncate_cache)){
-            S3FS_PRN_ERR("failed adding stat cache [path=%s]", strpath.c_str());
-            return -ENOENT;
-        }
+        // readdir_multi_head calls AddStat but check it again
```
ggtakec, line 459: The original code called AddStat again because it took into account when `add_no_truncate_cache` was `true`.
There is a problem with the original logic, but in `readdir_multi_head`(`multi_head_callback`) the AddStat call is always `no_truncate=false`.

So, if `add_no_truncate_cache=true`, I think you need to call AddStat again or review the logic.

src/s3fs.cpp
```diff
@@ -2538,37 +2499,32 @@ static S3fsCurl* multi_head_retry_callback(S3fsCurl* s3fscurl)
...
-        std::string etag     = head.GetETag((*iter).c_str());
 
         std::string fillpath = disppath;
         if('/' == disppath[disppath.length() - 1]){
             fillpath.erase(fillpath.length() -1);
         }
         fillerlist.push_back(fillpath);
 
-        if(StatCache::getStatCacheData()->HasStat(disppath, etag.c_str())){
+        // TODO: ETag?
```
ggtakec, line 2526: I think it is necessary to compare ETags.

src/s3fs.cpp
```diff
@@ -393,78 +402,24 @@ static int get_object_attribute(const char* path, struct stat* pstbuf, headers_t
...
-                strpath  += "/";
-                *pisforce = true;
-                result    = 0;
-            }
-        }
+        parent = strpath.substr(0, pos + 1);
+        child = strpath.substr(pos + 1);
+    }
+    headlist.push_back(child);
+    headlist.push_back(child + "/");
```
ggtakec, line 417: I request path, path/, and path_$folder$ in parallel, but it seems that the priority is unclear if one or more of them exist.
(We need to consider whether that situation exists, because we are parallelizing from sequential.)
At present, it seems that the content that received the response (normal) at the very end is registered as stat.

ggtakec (approved):
Please check out some comments.

src/s3fs.cpp
gaul: I will add a test for this.  If there exists `path` and `path/` the current behavior is to return `path`.

gaul:
Closing this as stale.

PR #1654, opened 2021-05-08
Fix off_t to int narrowing in PageList::GetSparseFilePages

gaul:
Found via clang-tidy.


gaul:
@ggtakec can this cause data corruption?  If so I would like to run another release.

ggtakec (approved):
Thanks, this is my fault.

PR #1655, opened 2021-05-09
Additional bug fixing for not creating zero-byte object

ggtakec:
#1640

## Details
Current s3fs_create function no longer creates(does not upload) a 0-byte initial file, only adds StatCache for it.
However, there was a problem because the response to the situation that this file object on the s3 does not exist yet was insufficient.

First, StatCache has an expiration date, and if the expiration date has expired, StatCache will be deleted.
Even if it has not expired, it may be deleted by renewal or deleted directly.
When s3fs gets the information of the file, if StatCache does not exist, it tries to get the information from the real file(object on S3).
In other words, the current s3fs has a problem because the initial file(0 bytes) does not exist(is not uploaded).

Thus this PR changed the code to use the `no_truncate` flag to prevent automatic StatCache deletion.
(changed to specify this flag when calling AddStat of the s3fs_create function.)
And I made a fix related to this flag.

### About timeout for stat cache
FUSE on macOS seems to behave differently than that on other operating systems.
First, create a new file with s3fs_create, and with that file open, write to the file.
During this write process, s3fs_access may be called. (It is unknown whether the processing time is long or it depends on the file size)

Since s3fs_access was called, s3fs checks the Stat information.
If the Stats information has expired, it will immediately discard the Stats information and try to get the file information from the S3.
And since the actual file does not exist, an error(Device not configured) will occur.
This bug occurred when I ran the test_multipart_mix test on macOS with `-o dbglevel=dbg`.

In addition, we have not confirmed the occurrence of the problem due to this timeout on other OS.
But if s3fs gets a similar call, it is easy to imagine same error.

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -100,7 +100,9 @@ inline bool IsExpireStatCacheTime(const struct timespec& ts, const time_t& expir
 {
     struct timespec nowts;
     SetStatCacheTime(nowts);
-    return ((ts.tv_sec + expire) < nowts.tv_sec);
+    nowts.tv_sec -= expire;
+
+    return (0 < CompareStatCacheTime(nowts, ts));
 }
 
 //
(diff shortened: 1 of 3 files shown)


gaul:
This seems like a great fix -- is there any way we can test this, perhaps by opening a new file for writing then creating enough other files to evict the stat cache?

ggtakec:
I think this test case can be difficult.

Recreating the bug found on macOS seems to be quick.
To imitate it, you need to test the following steps.
(1) Create new file --Keep open(fd)
(2) Write to the file (it is allowed to be stopped in the middle, but it will continue until (3) is finished)
(3) Check the access to the file after the Cache expire time has elapsed (somehow call s3fs_access as a separate process, not via fd)
It maybe chmod etc is possible instead of access check.
(4) Continue writing, and flush / release(close fd)

The problem occurred after the process of (3) on macOS.

PR #1660, opened 2021-05-19
Increase test robustness by adding LC_ALL=C

### LINKED ISSUE #1658: Increase test robustness by adding LC_ALL=C
Some test cases check the output of commands for English expressions. This may cause the test suite to fail for users with non-English locale settings.

I suggest to add `LC_ALL=C` at the beginning of `test_utils.sh` to address this problem.

I would open a PR, if you agree with this solution.

CarstenGrohmann:
Use C locale, because some tests check for English expressions

Fixes #1658


ggtakec:

PR #1647, opened 2021-05-04
Don't do a multipart upload with first part smaller than 5MB

CarstenGrohmann:
When the temporary storage filled up, the old implementation starts a multipart upload with the current data even if the minimum multipart size is not reached. This can cause errors depending on the S3 implementation.

There is no real solution for a shortage of temporary storage. The change will implement 2 mitigations. They may help or may not. This depends on the speed of the incoming data vs. the speed of writing data to S3.

The new implementation handles two different scenarios:

1. minimum part not reached: emits a warning and returns -ENOSPC

2. temporary storage is between minimum part size and set multipart size: permanently reduce multipart size to the current size

   This scenario may cause the multipart size to be repeatedly, incrementally, and permanently reduced to the minimum size. 
   
   There is no guarantee that this frees enough memory fast enough. As you see in "Example Scenario 2", the multipart is reduced but copy fails nevertheless.
   
   Maybe this can be solved by an additional `ftruncate`.
   
   This condition can be merged with the first to emit a warning and return -ENOSPC for all requests smaller than the multipart size.

What do you think about this change?

## Starting s3fs
```
# ./s3fs mybucket /mnt -o url=http://mys3service,use_path_request_style,multipart_size=256,curldbg,dbglevel=debug -d -d -f
2021-05-04T09:12:18.519Z [INF] s3fs version 1.89(8c58ba8) : s3fs -o url=http://mys3service,use_path_request_style,multipart_size=256,curldbg,dbglevel=debug -d -d -f mybucket /mnt
2021-05-04T09:12:18.520Z [CRT] s3fs_logger.cpp:LowSetLogLevel(219): change debug level from [CRT] to [DBG]
2021-05-04T09:12:18.520Z [INF]     s3fs.cpp:set_mountpoint_attribute(4067): PROC(uid=0, gid=0) - MountPoint(uid=0, gid=0, mode=40700)
2021-05-04T09:12:19.522Z [DBG] curl.cpp:InitMimeType(408): Try to load mime types from /etc/mime.types file.
2021-05-04T09:12:19.522Z [DBG] curl.cpp:InitMimeType(413): The old mime types are cleared to load new mime types.
2021-05-04T09:12:19.523Z [INF] curl.cpp:InitMimeType(436): Loaded mime information from /etc/mime.types
2021-05-04T09:12:19.524Z [INF] fdcache_stat.cpp:CheckCacheFileStatTopDir(79): The path to cache top dir is empty, thus not need to check permission.
[...]
```

[...]

PR #1664, opened 2021-05-21
Fixed a bug when the disk capacity was insufficient in RowFlush

ggtakec:
n/a

### Details
I found a bug in the FdEntity::RowFlush method and fixed it.
It was forgot to start multipart upload in the process when it detected that the disk space was insufficient.
The current test was unable to detect this bug because it had to detect out of disk space just when flushing.
The test case for this will be difficult, though we need to think about it in the future.

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -1073,7 +1073,6 @@ int FdEntity::NoCacheLoadAndPost(off_t start, off_t size)
             }else{
                 // already loaded area
             }
-
             // single area upload by multipart post
             if(0 != (result = NoCacheMultipartPost(upload_fd, offset, oneread))){
               S3FS_PRN_ERR("failed to multipart post(start=%lld, size=%lld) for file(%d).", static_cast<long long int>(offset), static_cast<long long int>(oneread), upload_fd);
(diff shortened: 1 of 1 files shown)


src/fdcache_entity.cpp
```diff
@@ -1224,12 +1223,22 @@ int FdEntity::RowFlush(const char* tpath, bool force_sync)
...
-                  return static_cast<ssize_t>(result);
-              }
+                // no enough disk space
+                if(nomultipart){
+                    S3FS_PRN_WARN("Not enough local storage to flush: [path=%s][fd=%d]", path.c_str(), fd);
+                    return -ENOSPC;   // No space left on device
+                }
+                if(0 != (result = NoCachePreMultipartPost())){
+                    S3FS_PRN_ERR("failed to switch multipart uploading with no cache(errno=%d)", result);
+                    return static_cast<ssize_t>(result);
```
gaul, line 1233: I think these `ssize_t` casts are not necessary.  `Load`, `NoCacheLoadAndPost`, and `NoCachePreMultipartPost` all return `int`.  I am OK merging this PR but I am concerns about the free-space handling generally.

gaul (changes requested):
Could you change the typo from RawFlush to RowFlush?

@CarstenGrohmann Can you also look at this since it touches some `nomultipart` code?

src/fdcache_entity.cpp
ggtakec: It is changed not to do static_cast and changed the same part.

This change is simply a bug in processing omissions, so it's a different issue than you're concerned about.
I agree about regarding the check of free space. I also feel that the problem is that the area size that must be checked differs depending on the upload mode.

CarstenGrohmann: Reading your discussion and reflecting my suggested change in #1647, I'll rework the change and move the check to `NoCachePreMultipartPost()` - it's be better placed and would also enforce the minimum size for this case.

gaul: Agreed that this improves the correctness.  Let's think some more about free space handling in future PRs.

ggtakec: Thanks

PR #1672, opened 2021-06-04
Fix implicit narrowing conversions

gaul:
These do not appear to be problematic but rather just clean up warnings.
Found via clang `-Wshorten-64-to-32`.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -607,10 +607,10 @@ size_t S3fsCurl::ReadCallback(void* ptr, size_t size, size_t nmemb, void* userp)
     if(0 >= pCurl->postdata_remaining){
         return 0;
     }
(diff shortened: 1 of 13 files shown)


ggtakec:
The following error has occurred in CentOS7.
```
terminate called after throwing an instance of'std :: bad_alloc'
   what (): std :: bad_alloc
```
This seems to be happening because the processes such as s3fs, test programs, s3proxy, etc. have fragmented memory and cannot be allocated for processes.
We will need to investigate the cause of this, but it is necessary to investigate whether the cause is s3fs at first.
(It may be trying to secure a large memory area.)

I don't think it's the effect of this PR, so I'll merge this PR.

ggtakec (approved):

PR #1673, opened 2021-06-05
Prefer std::string::rbegin over operator[]

gaul:
This is more concise and safer due to not repeating the variable name.
We cannot use `std::string::back` since it is not available in C++03.

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -249,7 +249,7 @@ bool StatCache::GetStat(const std::string& key, struct stat* pst, headers_t* met
     AutoLock lock(&StatCache::stat_cache_lock);
 
     stat_cache_t::iterator iter = stat_cache.end();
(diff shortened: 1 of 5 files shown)


gaul:
Something interesting is happening with the CentOS 7 builds.  Perhaps a regression from #1666?

ggtakec (approved):

ggtakec:
The following error has occurred in CentOS7.
```
terminate called after throwing an instance of'std :: bad_alloc'
   what (): std :: bad_alloc
```
This seems to be happening because the processes such as s3fs, test programs, s3proxy, etc. have fragmented memory and cannot be allocated for processes.
We will need to investigate the cause of this, but it is necessary to investigate whether the cause is s3fs at first.
(It may be trying to secure a large memory area.)

I don't think it's the effect of this PR, so I'll merge this PR.

PR #1678, opened 2021-06-13
Protect orgmeta with fdent_lock

### LINKED ISSUE #1677: CentOS7 tests may give std::bad_alloc
### Related Issue (PR)
#1672 , #1673

### Detail
In the `test_concurrent_directory_updates` test of CentOS7, it seems that the following error is occasionally displayed.
```
terminate called after throwing an instance of 'std::bad_alloc'
   what():std::bad_alloc
```
There is a possibility that memory fragmentation has occurred, or that a large memory area is reserved.
However, it is possible that the process that caused this is not just s3fs, but other programs running at the same time.

gaul:
Fixes a crash seen with the CentOS 7 builder and when running with
AddressSanitizer.  Regression introduced by
ac578d188e811ed6bdc7a29fa3b0c810fc29d739.  Fixes #1677.

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -1323,9 +1323,9 @@ int FdEntity::RowFlush(int fd, const char* tpath, bool force_sync)
     }
     int         result     = 0;
     std::string tmppath    = path;
-    headers_t   tmporgmeta = orgmeta;
 
     AutoLock auto_lock(&fdent_lock);
+    headers_t tmporgmeta = orgmeta;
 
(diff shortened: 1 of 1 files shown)


gaul:
It would be nice if we could integrate `test/run_tests_using_sanitizers.sh` into CI but these can take an hour or more to complete.

ggtakec:
Thanks!
It was a bug embedded in #1666.
(opensuse/leap:15 also fails, but this has nothing to do with the code, so merge this.)

ggtakec (approved):

PR #1681, opened 2021-06-14
Set CURLOPT_UNRESTRICTED_AUTH when authenticating

jstastny:
This is necessary for authentication to work with AWS when it responds with 307 to the check bucket request when mounting. This happens to newly created buckets according to https://aws.amazon.com/premiumsupport/knowledge-center/s3-http-307-response/.

N/A

### Details
Prior to this, curl would follow the redirect, but would not include the `Authorization` header which would end up with `404`.

This is related to the change in cURL `7.58.0`, where cURL stopped propagating the `Authorization` header to redirects by default as mitigation of this CVE https://curl.se/docs/CVE-2018-1000007.html


gaul:
This makes sense but is there a safer way to address this, e.g., resign the request after 307?

jstastny:
I guess we could, but at the same time, trusting the `Location` of the `307` we get from AWS (or alternative provider), is something we could trust.

In case we wanted to go with resigning, we would have to stop following the redirects.

ggtakec:
@jstastny Thanks for your help.
About this issue, I think it seems that the redirect destination can be detected from the response body in the case of 307 response code, then it may be possible to use it for retry processing.

However, I think it is difficult for s3fs to determine whether the response is from a trusted server when the 307 is received, whether the redirect destination is safe, and so on.
So it's better to add a option(ex. `redirect....`) and let the s3fs user decide, rather than having code that automatically determines this issue.
For example, we can reduce the work of the user by providing that option to redirect(set CURLOPT_UNRESTRICTED_AUTH for curl) as default behavior.
We don't want to have more options, but I feel this problem needs it.

@gaul How about this solution?

jstastny:
@ggtakec Thanks for the comment.

I don't have a deep understanding of the code, but this https://github.com/s3fs-fuse/s3fs-fuse/blob/8b77821f4cad342f221320842c981b1471cb3d8b/src/curl.cpp#L1926 seems to suggest that curl is configured to follow the redirects automatically by default. In case you wanted to make this configurable, I can see two options -- it would either have to be configurable globally (for all curl calls -- and therefore removing the `CURLOPT_FOLLOWLOCATION` mentioned above) or for example just for selected curl calls (like the `checkBucket` affected by this PR). But again -- not being well versed in the architecture of the code, I don't feel like being able to make the right call here.
I personally don't see an issue in trusting the redirect's `Location` and passing the authentication headers to it.

ggtakec:
@jstastny Thank you for taking the time, I will merge this PR as it is.

Your opinion would be correct, and I missed this setting.
It can be assumed that the first endpoint accessed is a user-specified server, and that the trust relationship of the server itself is the responsibility and trust of the user.

ggtakec (approved):

jstastny:
Thanks a lot for merging this to master.

Issue #1676, opened 2021-06-09
Deadlock in current development version


CarstenGrohmann:
Dear S3FS Developer,

Running the current master to copy a file and do a parallel `ls` causes a deadlock.

After I started a copy job, I repeatedly displayed the contents of the mount. After a few repetitions, s3fs hung at `fdcache.cpp:Open(468)`.

strace shows a `futex(0x7ffe383480b0, FUTEX_WAIT_PRIVATE, 0, NULL` in this state. After waiting a few minutes I terminated `s3fs` with `kill -9` (SIGKILL). I simple `kill` doesn't terminate the process.

I can reproduce this issue only with the current master f5701fa. I'm not able to reproduce this issue with the older commit 9bf525e.

**More details:**
```
# ps -ef | grep s3
root     28283 28213 22 12:44 pts/4    00:00:33 ./s3fs mybucket /s3 -o url=http://mys3service:8088,use_path_request_style,notsup_compat_dir,enable_noobj_cache,curldbg,dbglevel=debug -d -d -f

# strace -p 28283
strace: Process 28283 attached
futex(0x7ffe383480b0, FUTEX_WAIT_PRIVATE, 0, NULL^Cstrace: Process 28283 detached
 <detached ...>

# gdb -p 28283
GNU gdb (GDB) Red Hat Enterprise Linux 7.6.1-114.el7
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
For bug reporting instructions, please see:
[...]
(gdb) bt full
#0  0x00007f95b9d7cadb in futex_abstimed_wait (cancel=true, private=<optimized out>, abstime=0x0, expected=0, futex=0x7ffe383480b0) at ../nptl/sysdeps/unix/sysv/linux/sem_waitcommon.c:43
        __to = 0x0
        __status = <optimized out>
        _val = 0
        err = <optimized out>
        oldtype = 0
#1  do_futex_wait (sem=sem@entry=0x7ffe383480b0, abstime=0x0) at ../nptl/sysdeps/unix/sysv/linux/sem_waitcommon.c:223
        err = <optimized out>
[...]

PR #1682, opened 2021-06-16
Set decimal base for converting subseconds to int

CarstenGrohmann:
The leading 0 triggers an automatic conversion as an octal value. This fails because it is a decimal value.

Setting the base to 10 prevents this automatism and treats the value as a decimal value.

Fixes: 
```
[WAN] string_util.cpp:cvt_strtoofft(96): something error is occurred in convert std::string(017080564) to off_t, thus return 0 as default.
```

Related to #1676


gaul:
Thank you for your contribution @CarstenGrohmann!  I will tidy up the rest of the code base in #1683.

PR #1684, opened 2021-06-17
Improve warning for terminated threads with non-zero return code

CarstenGrohmann:
I struggled a little with the meaning of the previous warning. So I have improved the wording a little to be more precise.


gaul:

PR #1688, opened 2021-06-18
Set Content-MD5 header even for empty objects

hdima:
If Content-MD5 header is enabled set it even for empty objects. The Content-MD5 header is required for any request to upload an object with a retention period configured using Amazon S3 Object Lock.


gaul:
Thank you for your contribution @hdima!

Issue #1685, opened 2021-06-18
Parallel Multipart Upload failed because of Content Length 0


CarstenGrohmann:
I've a simple copy job. The `cp` command reports an IO error for setting timestamps. The file content is correct.

When I follow the log entries, I see that setting the timestamps triggers a multipart upload. This upload fails. In doing so `cp` reports `preserving times for "/s3mnt/3750M": Input/output error`.

But at the end, the flush command also triggers a multipart upload and this second upload is successful.

The difference between the two multipart uploads is: the first upload uses the `x-amz-copy-source-range` header but does not provide any content (`Content-Length: 0`) and the second upload does not use `x-amz-copy-source-range` but the correct `Content-Length:`.

Additionally, I wonder why the final multipart upload does not set the correct times in the initial post request as shown in `step 7`?

**Steps to reproduce:**

 1. Start s3fs
 
     ``` 
    ./s3fs mybucket /s3mnt -o url=http://mys3service:8088,use_path_request_style,notsup_compat_dir,enable_noobj_cache,curldbg,dbglevel=debug -d -d -f -o tmpdir=/extradisk/tmp/
    ``` 
	
 2. Copy a larger file
 
    ``` 
	# cp -p /extradisk/3750M /s3mnt/
	cp: overwrite "/s3mnt/3750M"? y
	cp: preserving times for "/s3mnt/3750M": Input/output error
    ``` 

 3. Content is correct but the timestamp is wrong
 
    ``` 
	# ll /extradisk/3750M /s3mnt/3750M
	-rw-r--r-- 1 root root 3929478700 Jun 16 14:57 /s3mnt/3750M
	-rw-r--r-- 1 root root 3929478700 Mar  4 14:08 /extradisk/3750M

	# md5sum /s3mnt/3750M /extradisk/3750M
	23cff1bf6eabb1a7743a35d8d91699ed  /s3mnt/3750M
	23cff1bf6eabb1a7743a35d8d91699ed  /extradisk/3750M
    ``` 
	
**Detailed analysis:**

 1. file has fully written to the temporary file

    ``` 
	419966 unique: 59966, opcode: WRITE (16), nodeid: 2, insize: 65616, pid: 6790
	419967 write[2] 65536 bytes to 3929341952 flags: 0x8001
[...]

PR #1689, opened 2021-06-20
Fixed a bug utimens is calling before flush

ggtakec:
#1685

### Details
In the case where utimens was called before flush after updating the file was called, it was wrong to determine that it was being updated.
(It seems to have existed for a long time, but it may have become apparent with the modification of #1666.)
There was a file fix that required the meta data to be updated, and I was getting an error because I was doing an upload with utimens before uploading that meta information.

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -1745,7 +1745,7 @@ bool FdEntity::MergeOrgMeta(headers_t& updatemeta)
     if(0 <= atime.tv_sec){
         SetAtime(atime, true);
     }
-    is_meta_pending |= IsUploading(true);
+    is_meta_pending |= (IsUploading(true) || pagelist.IsModified());
 
     return is_meta_pending;
 }
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -624,6 +624,20 @@ function test_multipart_mix {
     rm_test_file "${BIG_FILE}-mix"
 }
 
+function test_utimens_during_multipart {
+    describe "Testing utimens calling during multipart copy ..."
+
+    dd if=/dev/urandom of="${TEMP_DIR}/${BIG_FILE}" bs=$BIG_FILE_BLOCK_SIZE count=$BIG_FILE_COUNT
+
+    cp ${TEMP_DIR}/${BIG_FILE} ${BIG_FILE}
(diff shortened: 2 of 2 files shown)


src/fdcache_entity.cpp
```diff
@@ -1745,7 +1745,7 @@ bool FdEntity::MergeOrgMeta(headers_t& updatemeta)
     if(0 <= atime.tv_sec){
         SetAtime(atime, true);
     }
-    is_meta_pending |= IsUploading(true);
+    is_meta_pending |= (IsUploading(true) || pagelist.IsModified());
```
gaul, line 1748: Is there any test we can add for this condition?  @CarstenGrohmann

ggtakec: @gual This test case can be achieved by adding a simple test.
I will try to add it.

gaul (approved):

CarstenGrohmann:
The change solves my issue #1685:

**Test results w/o fix:**
```
# cp -p /extradisk/3750M /extradisk/

# cp -p /extradisk/3750M /extradisk/
cp: overwrite "/extradisk/3750M"? y
cp: preserving times for "/extradisk/3750M": Input/output error
```
```
test_utimens_during_multipart: "Testing utimens calling during multipart copy ..."
1+0 records in
1+0 records out
26214400 bytes (26 MB) copied, 0.171156 s, 153 MB/s
cp: preserving times for 'big-file-s3fs.txt': Input/output error
test_utimens_during_multipart failed
FAIL: test_utimens_during_multipart
```


**Test results w/ fix:**
```
# cp -p /extradisk/3750M /extradisk/

# cp -p /extradisk/3750M /extradisk/
cp: overwrite "/extradisk/3750M"? y
```
```
test_utimens_during_multipart: "Testing utimens calling during multipart copy ..."
1+0 records in
1+0 records out
26214400 bytes (26 MB) copied, 0.173588 s, 151 MB/s
test_utimens_during_multipart passed
PASS: test_utimens_during_multipart
```

@ggtakec Thank you for solving this issue.

ggtakec:
@CarstenGrohmann Thank you for your kindness.
By providing the logs and giving us the details of the phenomenon, I was able to find the cause of the problem in a short amount of time.

PR #1691, opened 2021-06-20
Set LC_ALL=en_US.UTF-8 in tests

gaul:
This allows tests to pass using the S3Proxy filesystem provider.
References #1665.


gaul:
Sorry, this is only needed for the filesytem jclouds backend.  I will update the description.

ggtakec (approved):

PR #1694, opened 2021-06-20
Ensure NUL-terminated result after strncpy

CarstenGrohmann:
Long symlinks may cause that the result buffer is filled and not proper terminated with a null byte.


ggtakec (approved):

ggtakec:

PR #1696, opened 2021-06-22
Set explicit Content-Length: 0 when initiating MPU

### LINKED ISSUE #1661: Investigate GCS S3 multipart upload support
This is in public preview:

https://cloud.google.com/storage/docs/xml-api/post-object-multipart

This would vastly improve metadata mutations, e.g., `chmod`.

gaul:
This works around a GCS quirk that incorrectly requires this header.
Fixes #1661.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -3528,7 +3528,6 @@ int S3fsCurl::PreMultipartPostRequest(const char* tpath, headers_t& meta, std::s
     }
 
     requestHeaders = curl_slist_sort_insert(requestHeaders, "Accept", NULL);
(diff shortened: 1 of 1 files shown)


gaul:
I tracked the origin of omitting `Content-Length` to 4f953f9bd78c81afd825dab49a81ad2a68db0be4 and there does not seem to be a strong reason to omit the header.

gaul:
I also reported this S3 deviation to the GCS team.  But I would expect that other clients are sending the zero header.

ggtakec:
@gaul Thanks for your help.
Maybe it wasn't something we intended, and I think we forgot to reset it with CURLOPT_INFILESIZE after removing Content-Length.

ggtakec (approved):

PR #1702, opened 2021-06-27
Remove unneeded volatile qualifiers

gaul:
We should use proper locking instead.  Specifically for
`is_meta_pending` this does not do what was intended due to the
read-modify-write of the member.

### DIFF
--- a/src/curl.h
+++ b/src/curl.h
@@ -178,7 +178,7 @@ class S3fsCurl
         headers_t            responseHeaders;      // header data by HeaderCallback
         BodyData             bodydata;             // body data by WriteMemoryCallback
         BodyData             headdata;             // header data by WriteMemoryCallback
-        volatile long        LastResponseCode;
+        long                 LastResponseCode;
         const unsigned char* postdata;             // use by post method and read callback function.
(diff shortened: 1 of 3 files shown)


ggtakec:
@gaul Thanks, is_meta_pending is a boolean write, but it didn't seem to be atomic.
There is something wrong with the call to fdent_lock, so can I fix it with this PR if it's okay?


The code below: (fdcache_entity.cpp)
```
int FdEntity::UploadPendingMeta()
{
    if(!is_meta_pending) {
       return 0;
    }

    AutoLock auto_lock(&fdent_lock);
    ...
```

It should have been written as:
```
int FdEntity::UploadPendingMeta()
{
    AutoLock auto_lock(&fdent_lock);

    if(!is_meta_pending) {
       return 0;
    }
    ...
```

gaul:
> @gaul Thanks, is_meta_pending is a boolean write, but it didn't seem to be atomic.
> There is something wrong with the call to fdent_lock, so can I fix it with this PR if it's okay?

Done.

ggtakec (approved):

ggtakec:
@gaul Thanks

PR #1710, opened 2021-07-01
Add flag to run all tests

gaul:
This allows CI to run the full set of tests while making local
development faster.


gaul:
The tests are completing unusually quickly -- it seems like the `ALL_TESTS` flag doesn't work in CI?

ggtakec (approved):

PR #1713, opened 2021-07-03
Prevents the etag string buffer from being destroyed during use

ggtakec:
n/a

### Details
There was a potential bug in the etag string saving process in parallel  multipart uploads.
Currently, it has not been realized, but it may be a problem in future function addition(planned now), so it has been fixed.

This bug is the part that saves etag string when CURL processing is completed in parallel multipart upload.
The object has an etag as a pointer, but it is possible that the entity does not exist in the future.
Therefore, I added an `etag_entities` member to the `PseudoFdInfo` class and modified it so that the etag entity always exists for the period when the `PseudoFdInfo` object exists.

And the `MPPART_INFO` structure has almost the same purpose and member variables as the `filepart` structure, so it is unified as `filepart`.

The `MPPART_INFO` changes are useless after this integration, so the above fix and this integration are combined into a single(this) PR.

### DIFF
--- a/src/fdcache_fdinfo.cpp
+++ b/src/fdcache_fdinfo.cpp
@@ -159,8 +159,13 @@ bool PseudoFdInfo::GetEtaglist(etaglist_t& list)
     AutoLock auto_lock(&upload_list_lock);
 
     list.clear();
-    for(mppart_list_t::const_iterator iter = upload_list.begin(); iter != upload_list.end(); ++iter){
-        list.push_back(iter->etag);
+    for(filepart_list_t::const_iterator iter = upload_list.begin(); iter != upload_list.end(); ++iter){
+        if(iter->petag){
+            list.push_back(*(iter->petag));
+        }else{
+            S3FS_PRN_ERR("The pointer to the etag string is null(internal error).");
+            return false;
+        }
     }
     return !list.empty();
 }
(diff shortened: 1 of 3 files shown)


src/fdcache_fdinfo.cpp
```diff
@@ -159,8 +159,12 @@ bool PseudoFdInfo::GetEtaglist(etaglist_t& list)
     AutoLock auto_lock(&upload_list_lock);
 
     list.clear();
-    for(mppart_list_t::const_iterator iter = upload_list.begin(); iter != upload_list.end(); ++iter){
-        list.push_back(iter->etag);
+    for(filepart_list_t::const_iterator iter = upload_list.begin(); iter != upload_list.end(); ++iter){
+        if(iter->petag){
+            list.push_back(*(iter->petag));
+        }else{
+            S3FS_PRN_ERR("The pointer to the etag string is null(internal error). But will continue and the upload will fail.");
```
gaul, line 166: Should s3fs eagerly fail this instead of waiting for the server to fail it after the upload?

ggtakec: By returning an error with this method, I confirmed that EIO can immediately interrupt the process.
I have uploaded the code, so please check it.

PR #1714, opened 2021-07-04
Splited some methods of FdEntity class by uploading mode

ggtakec:
n/a 

### Details
This is not a bug or a fix for a logic change.

The `FdEntity::RowFlush` and `FdEntity::Write` methods are common method for all uploading options(`nomultipart`, `multipart`, `mixmaultipart`).

Depending on the upload mode, the processing of these methods is roughly similar logic, but slightly different.
For this reason, we should be aware of the impact on other modes when modifying or changing methods.

I've split these methods in consideration of adding more modes in the future.
The processing of `multipart` and `mixmultipart` is very similar and seems verbose, but I focused on mode specific independence.

@gaul Please give me your opinion.
I thought it would be better to split it when adding a new mode(ex.  adding new mode which is uploading file as a stream).

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -1312,6 +1312,20 @@ off_t FdEntity::BytesModified()
     return pagelist.BytesModified();
 }
 
+// [NOTE]
+// There are conditions that allow you to perform multipart uploads.
+// 
+// According to the AWS spec:
+//  - 1 to 10,000 parts are allowed
+//  - minimum size of parts is 5MB (except for the last part)
+// 
+// For example, if you set the minimum part size to 5MB, you can upload
+// a maximum (5 * 10,000)MB file.
+// The part size can be changed in MB units, then the maximum file size
+// that can be handled can be further increased.
+// Files smaller than the minimum part size will not be multipart uploaded,
+// but will be uploaded as single part(normally).
+//
 int FdEntity::RowFlush(int fd, const char* tpath, bool force_sync)
 {
     S3FS_PRN_INFO3("[tpath=%s][path=%s][pseudo_fd=%d][physical_fd=%d]", SAFESTRPTR(tpath), path.c_str(), fd, physical_fd);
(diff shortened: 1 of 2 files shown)


gaul (commented):
This seems reasonable but maybe the comment is stale copy and paste?

src/fdcache_entity.cpp
```diff
@@ -1312,18 +1312,27 @@ off_t FdEntity::BytesModified()
...
 
+// [NOTE]
+// There are conditions that allow you to perform multipart uploads.
+// 
+// According to the AWS spec:
+//  - 1 to 10,000 parts are allowed
+//  - minimum size of parts is 5MB (expect for the last part)
+// 
+// For our application, we will define minimum part size to be 10MB (10 * 2^20 Bytes)
+// minimum file size will be 64 GB - 2 ** 36
```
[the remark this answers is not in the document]
ggtakec: That's right, I was copying the old comment as it was.
The content is incorrect, so I modified the comments to match the current situation.
Please let me know if you find that it is incrrect yet.

And, I think `Middleware-large-object-mocking` may be useful for testing s3fs. (Although this is just an idea, we maybe get a good performance for s3fs large object test.)

src/fdcache_entity.cpp
```diff
@@ -1312,18 +1312,29 @@ off_t FdEntity::BytesModified()
     return pagelist.BytesModified();
 }
 
+// [NOTE]
+// There are conditions that allow you to perform multipart uploads.
+// 
+// According to the AWS spec:
+//  - 1 to 10,000 parts are allowed
+//  - minimum size of parts is 5MB (expect for the last part)
```
gaul, line 1320: s/expect/except/

gaul (approved):

src/fdcache_entity.cpp
ggtakec: Thanks, it has be fixed.

PR #1716, opened 2021-07-06
Add initiator pid, uid and gid to a debug log

dolfinus:
### Details
I've discovered a lot of requests to my Minio installation from one of servers with s3fs installed.
After changing the log level to debug I saw a lot of log items that some process is recursively fetching file attributes. But I was not able to find any information about the process which made these requests - `lsof` could return only the information about opened files, but it does not handle `get_attr` kernel calls.

So I've added a debug info in the s3fs process itself. This is a really useful feature.


ggtakec:
@dolfinus Thanks for your contributing.
Started CI build of PR(only the first time permission was required).
When the build is complete, it will be merged into master.

PR #1722, opened 2021-07-17
Fixed a bug in disk free space calculation

ggtakec:
n/a

## Details
There were two problems with the calculation of free disk space.

### (1) Call PageList::GetTotalUnloadedPageSize method with size=0
The `PageList::GetTotalUnloadedPageSize` method should check the entire area when `size=0`, but It wasn't able to do this.
Fixed this bug by doing the same process as the `PageList::GetUnloadedPages` method.

### (2) Calculation of free space in Mix MultipartUpload
Before #1714, I didn't notice it because I used the same method in all upload modes.
The calculation of free space in the case of Mix MultipartUpload is different from normal Multipart and No Multipart, and it has the feature that the object is not needed to download.
Therefore, the calculation of free space is different, but it was not able to distinguish in same function.
Thus, added an argument to the `GetTotalUnloadedPageSize` method and modified it so that it does not accumulate if it is larger than the minimum size(for multipart upload).

### NOTES
Free space testing needs to be enhanced, but I think it can only be achieved by giving s3fs a special debug test mode.
(We should think about it in the future)

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -1557,7 +1557,7 @@ int FdEntity::RowFlushMixMultipart(PseudoFdInfo* pseudo_obj, const char* tpath)
         // Start uploading
 
         // If there is no loading all of the area, loading all area.
-        off_t restsize = pagelist.GetTotalUnloadedPageSize();
+        off_t restsize = pagelist.GetTotalUnloadedPageSize(/* start */ 0, /* size = all */ 0, MIN_MULTIPART_SIZE);
 
         // Check rest size and free disk space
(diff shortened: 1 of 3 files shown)


gaul:
I think we need to make s3fs easier to unit test some of the behaviors since testing things like free space is difficult in integration tests.

ggtakec:
@gaul I'm thinking of adding test-specific code to `FdManager::GetFreeDiskSpace`.
From outside s3fs, give this method a value(dummy free disk size) and return the given size instead of the actual free size.
This is able to set free disk space for testing without being affected by actual size.
I think environment variables(which affect the whole test) or a file(which I recommend) are good ways to give them from external.

PR #1723, opened 2021-07-20
Add jitter to avoid thundering herd

gaul:


src/curl.cpp
```diff
@@ -2382,12 +2382,14 @@ int S3fsCurl::RequestPerform(bool dontAddAuthHeaders /*=false*/)
...
 
                     case 500:
-                    case 503:
+                    case 503: {
                         S3FS_PRN_INFO3("HTTP response code %ld was returned, slowing down", responseCode);
                         S3FS_PRN_DBG("Body Text: %s", bodydata.str());
-                        sleep(4 << retry_count);
+                        // Add jitter to avoid thundering herd.
+                        unsigned int sleep_time = 4 << retry_count;
+                        sleep(sleep_time / 2 + random() % sleep_time);
```
ggtakec, line 2390: Isn't this the same as below?
```
unsigned int sleep_time = 2 << retry_count;
sleep(sleep_time + random() % sleep_time);
```

ggtakec (commented):
Please check comment.

src/curl.cpp
gaul: It is similar but not the same since the PR has more dynamic range, for example retry #2 will have 12 ± 4 compared to 12 ± 8:

```c
unsigned int sleep_time = 2 << retry_count;
sleep(sleep_time + random() % sleep_time);
```

retry_count | sleep_time | sleep
----------- | ---------- | -------------
0           | 2          | 2 + (0 .. 1)
1           | 4          | 4 + (0 .. 3)
2           | 8          | 8 + (0 .. 7)

```c
unsigned int sleep_time = 4 << retry_count;
sleep(sleep_time / 2 + random() % sleep_time);
```

retry_count | sleep_time | sleep
----------- | ---------- | -------------
0           | 4          | 2 + (0 .. 3)
1           | 8          | 4 + (0 .. 7)
2           | 16         | 8 + (0 .. 15)

Your suggestion seems easier to understand so let's go with that.

ggtakec: Thanks, I agree.

ggtakec (approved):

PR #1727, opened 2021-07-22
Added fake_diskfree option to deceive free disk space for test

ggtakec:
#1722 

### Details
The `ensure_diskfree` option is an option to protect the free disk space.
The reason for not checking the disk usage of s3fs is that it causes performance degradation. (like the `du` command)
So that s3fs is using disk stat information will improve performance.
That's why we provide the `ensure_diskfree` option.

However, there are some issues that are very difficult to test, such as Github Actions.
This is because the Runner(VM, HOST, etc) have different capacities, so the same value cannot be used for the `ensure_diskfree` option.

To avoid this situation, we have added the option `fake_diskfree` to deceive free disk space.
The value of this option is stored as free disk space when s3fs is started, and after that, this value(difference from the actual free disk space) will be used.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -82,6 +82,7 @@ bool            FdManager::is_lock_init(false);
 std::string     FdManager::cache_dir;
 bool            FdManager::check_cache_dir_exist(false);
 off_t           FdManager::free_disk_space = 0;
+off_t           FdManager::fake_used_disk_space = 0;
 std::string     FdManager::check_cache_output;
 bool            FdManager::checked_lseek(false);
 bool            FdManager::have_lseek_hole(false);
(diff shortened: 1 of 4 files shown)


doc/man/s3fs.1
```diff
@@ -215,6 +215,12 @@ sets MB to ensure disk free space. This option means the threshold of free space
 s3fs makes file for downloading, uploading and caching files.
 If the disk free space is smaller than this value, s3fs do not use diskspace as possible in exchange for the performance.
 .TP
+\fB\-o\fR fake_diskfree (default 0)
```
gaul, line 218: We shouldn't add these kinds of test-only options to the help or man page.  It will just confuse users.

gaul (changes requested):
I would rather we moved to a unit test approach than expose these as integration tests.  However this is fine as an interim step.

doc/man/s3fs.1
ggtakec: OK, It will be removed soon.

src/s3fs_help.cpp
```diff
@@ -263,6 +263,17 @@ static const char help_string[] =
     "        space is smaller than this value, s3fs do not use diskspace\n"
     "        as possible in exchange for the performance.\n"
     "\n"
+    "   fake_diskfree (default 0)\n"
```
gaul, line 266: Do we need this help string?

ggtakec: The man page was confusing and should have been removed.
I'm a little worried about the help string.
The help string just remembers "what is this option for?", So it's unnecessary if this PR remains.(same reason as man page)

Since there is a history of this PR, do you want to delete it?

gaul: We could leave the help string but enclose it in an `#if 0 .. #endif` pair so it is not included?  I think a more robust solution would be a `--help-hidden` flag but think seems like too much work.

ggtakec: Since the PR history remains on github and we can see the code, I will delete this part as well.
(Because putting #ifdef in the code too much is confusing during maintenance)
Thanks.

PR #1729, opened 2021-07-25
Add support for AWS-style environment variables

### LINKED ISSUE #1708: AWSACCESSKEYID vs AWS_ACCESS_KEY_ID
The environment variables for an access key, secret key, and session token differ between s3fs and aws cli. 
In s3fs, these variables are in uppercase but without underscore, and aws cli uses the same naming but with underscore. Is this intentional?

https://github.com/s3fs-fuse/s3fs-fuse/blob/487df2700842c0a74e6314b898d726ff57a55589/src/s3fs.cpp#L3992-L3994
AWS reference: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html#envvars-set

If this is not intended, then I can contribute a PR.

CarstenGrohmann:
Support AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY and AWS_SESSION_TOKEN
in addition to the AWSACCESSKEYID, AWSSECRETACCESSKEY and
AWSSESSIONTOKEN.

The old environment variables are still supported, but they are deprecated and no longer documented.

I would add a deprecation warning for the original environment variables if suggested. Users would 

Close #1708

### DIFF
--- a/doc/man/s3fs.1
+++ b/doc/man/s3fs.1
@@ -40,7 +40,7 @@ Password files can be stored in two locations:
  \fB$HOME/.passwd-s3fs\fP   [0600]
 .RE
 .PP
-s3fs also recognizes the \fBAWSACCESSKEYID\fP and \fBAWSSECRETACCESSKEY\fP environment variables.
+s3fs also recognizes the \fBAWS_ACCESS_KEY_ID\fP and \fBAWS_SECRET_ACCESS_KEY\fP environment variables.
 .SH OPTIONS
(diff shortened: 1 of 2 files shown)


gaul (approved):
I don't think we need any kind of deprecation.  We will probably support both variables forever.

ggtakec:
@CarstenGrohmann Thanks for your PR.
I agree with @gaul.

ggtakec (approved):

PR #1735, opened 2021-08-01
Add a partial page_list unit test

gaul:


src/fdcache_page.h
```diff
@@ -115,6 +117,7 @@ class PageList
         bool IsModified() const;
         bool ClearAllModified();
 
+        bool Compress();
```
gaul, line 120: Visible for testing.

gaul:
I think we can test some things more easily from unit tests instead of only integration tests.

ggtakec (approved):
Would the test programs under the src directory be moved to the test directory(or other directories, etc.) in the future?
As test code grows, we maybe consider separating it from main code directory.

PR #1747, opened 2021-08-24
fix misuse of IsUploading

VVoidV:
When disk cache space is not enough, write file will failed with EIO.

### Details
IsUploading() return false when upload_id is assigned, I think here we misused it.


ggtakec (approved):

ggtakec:
@VVoidV Thank you for discovering this stupid mistake.
This will be merge soon.

Issue #1746, opened 2021-08-24
locking problem


VVoidV:
When I test copy file into s3fs while disk cache is low, I got s3fs hang at 
```
#0  0x00007f5054f8e4ed in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007f5054f89dcb in _L_lock_883 () from /lib64/libpthread.so.0
#2  0x00007f5054f89c98 in pthread_mutex_lock () from /lib64/libpthread.so.0
#3  0x00000000004a1757 in AutoLock::AutoLock (this=0x7f504bffe500, pmutex=0x7f50440052d0, type=AutoLock::NONE) at autolock.cpp:47
#4  0x000000000048d57c in FdEntity::UploadPendingMeta (this=0x7f50440052d0) at fdcache_entity.cpp:2083
#5  0x000000000048976a in FdEntity::RowFlushMixMultipart (this=0x7f50440052d0, pseudo_obj=0x7f5044007160, tpath=0x0) at fdcache_entity.cpp:1658
#6  0x0000000000486a71 in FdEntity::RowFlush (this=0x7f50440052d0, fd=2, tpath=0x0, force_sync=false) at fdcache_entity.cpp:1362
#7  0x0000000000428e10 in FdEntity::Flush (this=0x7f50440052d0, fd=2, force_sync=false) at fdcache_entity.h:121
#8  0x0000000000416c46 in s3fs_flush (_path=0x7f504400efc0 "/ffff", fi=0x7f504bffebb0) at s3fs.cpp:2418
#9  0x00007f50563fe447 in fuse_flush_common () from /lib64/libfuse.so.2
#10 0x00007f50563fe6d0 in fuse_lib_flush () from /lib64/libfuse.so.2
#11 0x00007f5056404d06 in do_flush () from /lib64/libfuse.so.2
#12 0x00007f5056405b6b in fuse_ll_process_buf () from /lib64/libfuse.so.2
#13 0x00007f5056402401 in fuse_do_work () from /lib64/libfuse.so.2
#14 0x00007f5054f87dd5 in start_thread () from /lib64/libpthread.so.0
#15 0x00007f5054cb0ead in clone () from /lib64/libc.so.6
```
I think there is no need to lock  fdent_lock in UploadPendingMeta, because it already locked in RowFlush ?

![image](https://user-images.githubusercontent.com/17823956/130623087-e08410c1-3f45-4e9a-9071-c5a1e91383a8.png)

[...]

PR #1753, opened 2021-08-30
1. fix RowFlush can not upload last part smaller than 5MB using NoCac…

VVoidV:
1. fix RowFlush can not upload last part smaller than 5MB using NoCacheMultipartPost;
2. fix deadlock in UploadPendingMeta

https://github.com/s3fs-fuse/s3fs-fuse/issues/1746#issue-978094828

### Details
1. see issue 1746 above. And *x-amz-metadata-directive* Header is missing in this case.
2. pseudo_obj->GetLastUntreated gets part lager than 5MB with default para, we should get all data in flush case.


ggtakec:
@VVoidV Thanks for your great help.
The content of your PR seems to fix the problem correctly.
This PR will be merged immediately.

ggtakec (approved):

PR #1758, opened 2021-08-31
fclose(FILE*) instead of close(fileno(FILE*))

gaul:
This is the same thing but confuses Coverity.


ggtakec (approved):

ggtakec:
@gaul Thanks.
As you pointed out, there was a problem with this code.

PR #1760, opened 2021-09-03
fix IAM role retrieval from IMDSv2

nmeyerhans:
AWS IMDSv2 support was added in #1462, but the implementation did not
cover the addional IMDS access that occurs with the iam_role=auto
configuration.  This change implements IMDSv2 support for the IMDS
call to determine the instance's role name.

See also
https://stackoverflow.com/questions/69031023/how-to-make-s3fs-use-imds-v2-when-mounting-s3-buckets-from-ec2-instance

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -3188,10 +3188,40 @@ bool S3fsCurl::LoadIAMRoleFromMetaData()
     }
 
     // url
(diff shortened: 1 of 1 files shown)


gaul:
Thank you for your contribution @nmeyerhans!

PR #1774, opened 2021-10-09
Fix wrong function name in log message

CarstenGrohmann:
.


gaul:
CI fails because botocore uses Python 3.6 f-strings but Ubuntu 16.04 has Python 3.5.

PR #1780, opened 2021-10-16
Fixed wrong stat of cache after new creation file

ggtakec:
#1640

### Details
When creating an empty file(when opening a non-existent file with the create flag), the `FdEntity::MarkDirtyNewFile()` method adds a page information to the pagelist.
The page information to be added was start=0 and size=1.

This works now, but the stat information(`.<bucket>.stat/<file>`) for the cache file immediately after the file is created that it shows a 1byte file.
However, the size of the real cache file is zero and it seems to work correctly.

If s3fs does not have the modified page information, it can not create an empty file, so it is as above for avoiding this.

But it is not correct, since the stat information for cache file is creating an incorrect state, this page addition process has been fixed.
After that, enabled to add page information of size 0 to pagelist.

For safety, if any other page information exists, the size 0 page information will be compressed and deleted.
This empty page information can exist only when there is no other page information(ex: when creating an empty file).

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -2173,6 +2173,16 @@ bool FdEntity::PunchHole(off_t start, size_t size)
     return true;
 }
 
+// [NOTE]
+// Indicate that a new file's is dirty.
+// This ensures that both metadata and data are synced during flush.
+//
+void FdEntity::MarkDirtyNewFile()
+{
+    pagelist.Init(0, false, true);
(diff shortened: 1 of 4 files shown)


ggtakec:
@gaul 
Added a test to check the cache file stat when creating an empty file.
The contents of the stat file have already been inspected with `test_cache_file_stat`, but since this case is special, I prepared it with another function(`test_zero_cache_file_stat`).

PR #1781, opened 2021-10-17
More parallelized GithubActions(CI) execution and reduced time

ggtakec:
n/a

### Details
Execution of current our GithubActions is matrixed by only OS.
For each OS, we are testing cases of 10 different s3fs option combinations.
Thus, due to the long execution time of the entire test, it takes more than 15m-20m.

To save time, I've changed the test to be isolated in a combination case of 2-4 options and run matrix in this isolated test case with the OS.
This reduced the test time to 3m-5m.

This fix increases the number of parallel runs of Runner from 10 to 30.
I can't tell if this increase is violent for Github Actions, but since the execution time for each Runner is less than 1/3, I think that the occupancy time for Github Actions is the same or less.

Assuming that the number of test cases will increase in the future, I have shortened it as much as possible.
If you have any problems, please point them out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
       fail-fast: false
 
       #
-      # matrix for containers
+      # matrix for containers and test cases
       #
       matrix:
         container:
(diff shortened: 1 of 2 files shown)


gaul:
I am suspicious of this PR since for Ubuntu 21.10 install packages takes 1m 27s, build takes 50s, but tests only take 2s.  Is this testing all the flags?

ggtakec:
There was a ridiculous mistake in this PR, then I fixed it and rebase it to master.

I think this PR is working, but I'll change it to Draft once.

PR #1797, opened 2021-10-28
Use AWS CLI to create explicit times in the past

gaul:
s3fs can also do this via `utimensat` but tests should not trust this.
Also break tests into individual functions.  This further reduces test
run-time 8 seconds per flag.


ggtakec:
> Don't worry about AWS CLI overhead today. We can either use mc or s3cmd if AWS doesn't fix their client in the future.

All right.

ggtakec (approved):

PR #1801, opened 2021-11-01
Add option to allow unsigned payloads

gaul:
This reduces CPU usage of sigv4.  This reduces test
run-time by 7 seconds per flag.


test/small-integration-test.sh
```diff
@@ -49,13 +49,13 @@ if [ -n "${ALL_TESTS}" ]; then
...
         sigv2
-        sigv4
+        "sigv4 -o unsigned-payload"
         singlepart_copy_limit=10  # limit size to exercise multipart code paths
         #use_sse  # TODO: S3Proxy does not support SSE
     )
 else
     FLAGS=(
-        sigv4
+        "sigv4 -o unsigned-payload"
```
gaul, line 58: Actually, I guess we should enable this for all flags.

gaul: Done.

ggtakec (approved):

PR #1806, opened 2021-11-09
Bump CI to Fedora 35

gaul:


gaul:
Some of the CI runners fail with `CERTIFICATE_VERIFY_FAILED` although not all of them?  This is weird.  I have been thinking about using HTTP for tests to reduce overhead although this is only a small improvement.

ggtakec:
Maybe this is a python related error?
The error string is similar to the string output by python.

ggtakec (approved):

ggtakec:
This PR is for Fedora 35 and it's working fine.
The `CERTIFICATE_VERIFY_FAILED` error is ongoing and has nothing to do with this PR.(It reported to #1812)
So that I will merge this PR.

PR #1809, opened 2021-11-21
fix  mixupload return EntityTooSmall while a copypart is less than 5MB after split

VVoidV:
None

### Details
When I test append file, I found append 1 byte after a 513MB file while always reulsting in EIO.
The reason is mixupload will use copy API for previous 513MB.This will be splited into 512MB and 1MB. The data appeded will be uploaded as part 3. So we have 512M, 1M and 1 Byte part, this will cause EntityTooSmall when completing the MultipartUpload.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -1458,11 +1458,15 @@ int S3fsCurl::ParallelMixMultipartUploadRequest(const char* tpath, headers_t& me
                 S3fsCurl* s3fscurl_para              = new S3fsCurl(true);
 
                 bytes = std::min(static_cast<off_t>(GetMultipartCopySize()), iter->bytes - i);
-                /* every part should be larger than MIN_MULTIPART_SIZE */
+                /* every part should be larger than MIN_MULTIPART_SIZE and smaller than FIVE_GB */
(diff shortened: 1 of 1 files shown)


src/curl.cpp
```diff
@@ -1453,10 +1453,22 @@ int S3fsCurl::ParallelMixMultipartUploadRequest(const char* tpath, headers_t& me
             }
         }else{
             // Multipart copy
-            for(off_t i = 0; i < iter->bytes; i += GetMultipartCopySize()){
+            off_t bytes = 0;
+            for(off_t i = 0; i < iter->bytes; i += bytes){
```
ggtakec, line 1457: Can this for statement be changed as follows?
```
for(off_t i = 0, bytes = 0; i < iter->bytes; i += bytes){
```

ggtakec (changes requested):
Please check one line.

ggtakec:
@VVoidV Thank you for finding and fixing the bug. I think this fix is correct.
Please correct only one line.

ggtakec (approved):

ggtakec:
@VVoidV Thanks for your fix, it was merged.

Issue #1787, opened 2021-10-26
Reduce AWS CLI run-time


gaul:
The s3fs integration tests invoke the CLI 20 times * 10 flags = 200 invocations total.  Since the CLI takes 300 ms to start up this contributes 60 seconds to the 15 minute run-time.  We could reduce this by moving to minio CLI which only takes 25 ms.  References aws/aws-cli#6500.

gaul:
[aws-cli-repl](https://github.com/janakaud/aws-cli-repl) might address this.  But I opened several issues against it that which we should resolve first.

PR #1810, opened 2021-11-25
Use aws-cli-repl to avoid AWS CLI startup overhead

gaul:
This reduces test run-time by 6 seconds per flag.  References #1787.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,6 +76,7 @@ src/s3fs
 src/test_curl_util
 src/test_page_list
 src/test_string_util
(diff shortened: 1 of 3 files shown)


gaul:
This is broken in two different ways:

* integration-test-common.sh: line 180: kill: (7849) - No such process
* moved file content-type is not as expected expected:text/plain got:

I can reproduce the first symptom on my Debian system (but not my Fedora system).  I cannot reproduce the second symptom.

gaul:
#2689 will obsolete this.

PR #1823, opened 2021-12-29
Support CEPH multi tenancy bucket naming convention

tervo:
None

### Details
I added flag to handle CEPH multi tenancy naming convention (https://docs.ceph.com/en/latest/radosgw/multitenancy/), and would be great to the official version.

### DIFF
--- a/src/curl_util.cpp
+++ b/src/curl_util.cpp
@@ -288,6 +288,7 @@ std::string prepare_url(const char* url)
     }
 
     url_str = uri + hostname + path;
+
     S3FS_PRN_INFO3("URL changed is %s", url_str.c_str());
 
     return url_str;
--- a/src/curl_util.cpp
+++ b/src/curl_util.cpp
@@ -256,8 +256,13 @@ std::string prepare_url(const char* url)
     std::string uri;
     std::string hostname;
     std::string path;
+    std::string token;
     std::string url_str = std::string(url);
-    std::string token = std::string("/") + bucket;
+    if(is_multi_tenancy) {
+      token = std::string("/") + urlEncode(bucket);
+    } else {
+      token = std::string("/") + bucket;
+    }
     size_t bucket_pos;
     size_t bucket_length = token.size();
     size_t uri_length = 0;
@@ -283,7 +288,6 @@ std::string prepare_url(const char* url)
     }
 
     url_str = uri + hostname + path;
-
     S3FS_PRN_INFO3("URL changed is %s", url_str.c_str());
 
     return url_str;
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -73,7 +73,7 @@ static bool is_mp_umask           = false;// default does not set.
 static std::string mountpoint;
 static std::string passwd_file;
 static std::string mimetype_file;
-//static std::string tenancy_id;
+
 static bool nocopyapi             = false;
 static bool norenameapi           = false;
 static bool nonempty              = false;
@@ -94,7 +94,6 @@ static bool create_bucket         = false;
 static off_t multipart_threshold  = 25 * 1024 * 1024;
 static int64_t singlepart_copy_limit = 512 * 1024 * 1024;
 static bool is_specified_endpoint = false;
-//static bool is_multi_tenancy      = true;
 static int s3fs_init_deferred_exit_status = 0;
 static bool support_compat_dir    = true;// default supports compatibility directory type
 static int max_keys_list_object   = 1000;// default is 1000
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -73,6 +73,7 @@ static bool is_mp_umask           = false;// default does not set.
 static std::string mountpoint;
 static std::string passwd_file;
 static std::string mimetype_file;
+//static std::string tenancy_id;
 static bool nocopyapi             = false;
 static bool norenameapi           = false;
 static bool nonempty              = false;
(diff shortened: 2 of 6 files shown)


src/s3fs.cpp
```diff
@@ -73,6 +73,7 @@ static bool is_mp_umask           = false;// default does not set.
 static std::string mountpoint;
 static std::string passwd_file;
 static std::string mimetype_file;
+//static std::string tenancy_id;
```
gaul, line 76: Why is this commented out?

src/curl_util.cpp
```diff
@@ -283,7 +288,6 @@ std::string prepare_url(const char* url)
     }
 
     url_str = uri + hostname + path;
-
```
gaul, line 286: Please remove noisy changes.

src/curl_util.cpp
```diff
@@ -256,8 +256,13 @@ std::string prepare_url(const char* url)
     std::string uri;
     std::string hostname;
     std::string path;
+    std::string token;
     std::string url_str = std::string(url);
-    std::string token = std::string("/") + bucket;
+    if(is_multi_tenancy) {
+      token = std::string("/") + urlEncode(bucket);
```
gaul, line 262: I don't understand this change -- can we always urlEncode the bucket name?  What are the valid characters?  How do other S3 clients handle this?

gaul (changes requested):

src/s3fs.cpp
tervo: Moved to globals. Removed in the latest commit.

src/curl_util.cpp
tervo: Removed this.

src/curl_util.cpp
tervo: Url (containing bucket) is passed to the function as urlEncoded, so the token used to separate hostname and path need to be urlencoded as well. I guess bucket could be urlEncoded always, but I wanted to keep changes minimum. 

For example [AWSJavaSDK doc](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3URI.html
) says: 
> URL encoding is recommended if you work with bucket names or object keys with special characters. This can be disabled if you pre-encode URI strings before passing them to this class.

ggtakec: The argument `url` of this prepare_url function is URL encoded by MakeUrlResource function, so if we don't encode `token` as well, we may not be able to match.
In other words, it seems that the bucket name required by CEPH contains `:`, which is encoded.

It should be noted that except for path style, the domain name will contain `:`, so I think that it will not be processed correctly as FQDN. (Even if encoded, it will contain `%`)

Wouldn't it be necessary to enforce the path style to get this CEPH support?

tervo: [Ceph documentation](https://docs.ceph.com/en/latest/radosgw/multitenancy/) says:
>Note that it’s not possible to supply an explicit tenant using a hostname. Hostnames cannot contain colons, or any other separators that are not already valid in bucket names. Using a period creates an ambiguous syntax. Therefore, the bucket-in-URL-path format has to be used.

So, forcing path style when CEPH multi tenancy naming convention is used makes sense.

Committed corresponding changes to the pull request.

ggtakec: Thank you, I've confirmed your code fix that checks for a combination of is_multi_tenancy and pathrequeststyle.

@gaul
I think this fix is fine, please give us your opinion.

src/s3fs_help.cpp
```diff
@@ -147,6 +147,10 @@ static const char help_string[] =
     "        s3fs sets nocopyapi option automatically when public_bucket=1\n"
     "        option is specified.\n"
     "\n"
+    "   tenancy_id (default=\"\" which means disabled)\n"
+    "      - Tenancy id for CEPH multi tenancy naming convention\n"
+    "        (tenancy_id:bucket_name)\n"
+    "\n"
```
gaul, line 153: Please add a similar change to `doc/man/s3fs.1`.

src/s3fs_help.cpp
```diff
@@ -31,7 +31,7 @@
 //-------------------------------------------------------------------
 // Contents
 //-------------------------------------------------------------------
-static const char help_string[] = 
+static const char help_string[] =
```
gaul, line 34: Please remove _all_ noisy formatting changes from all files.  This ensures that the git log remains easy to understand.  You can submit these kinds of cleanup in a separate PR, preferably machine-generated by `sed` or `clang-tidy`.

src/curl_util.cpp
gaul: What are other clients doing with this multi-tenant option?  Do they have an explicit option, always URL-encode the bucket name, or require the user to manually URL-encode the `:`?  I am concerned about adding even more flags unless they are necessary.

gaul (changes requested):

src/curl_util.cpp
tervo: > What are other clients doing with this multi-tenant option? [...] I am concerned about adding even more flags unless they are necessary.
Well, this is a CEPH functionality so I assume the need to support this is general.

> Do they have an explicit option, always URL-encode the bucket name, or require the user to manually URL-encode the :?
Not sure if I understood this question, but user do no tneed to url-encode the :

gaul: When I look at the [Ceph example](https://docs.ceph.com/en/latest/radosgw/multitenancy/#s3) it doesn't have a special tenancy id option -- it includes the tenant in the bucket.  Thus I believe that s3fs users should also specify the tenant in the bucket and if there is an encoding issue it should be part of the bucket and not some weird option that only s3fs has.

gaul:
Closing due to unclear motivation and inactivity.

PR #1827, opened 2022-01-02
Remove unnecessary calls to dup

gaul:
These functions already use `pread` and keep track of an explicit offset
for the SHA-256 calculations.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -1279,29 +1279,18 @@ int S3fsCurl::ParallelMultipartUploadRequest(const char* tpath, headers_t& meta,
     int            result;
     std::string    upload_id;
     struct stat    st;
-    int            fd2;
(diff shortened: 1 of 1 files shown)


gaul:
This simplification requires the caller to keep the original fd open until all the requests complete but I think it is safe.  Let's merge the larger stream upload change first and I need to fix the single-part upload TODO as well.

ggtakec (approved):

ggtakec:
Since the TODO part is also a synchronous process (PUT request), I think that dup can be eliminated.

ggtakec:
I checked the TODO code again, but this part still needs to use dup().
This is because FILE* by fdopen linked to fd is closed by fclose, and fd is closed at that time.
I think that there is no problem with the other parts, so I will merge.
Sorry for my late response.

Issue #1828, opened 2022-01-02
Ubuntu 18.04, Buster, and Leap 15 CI fail


gaul:
These fail with:

```
moved file content-type is not as expected expected:text/plain got:
```

gaul:
Buster mime-support should have `/etc/mime.types`:

https://packages.debian.org/buster/mime-support

But this is not installed by default and `apt-get` does not install it?!

gaul:
The MIME issue was a red herring; the real error is above:

```
SSL validation failed for https://127.0.0.1:8080/s3fs-integration-test/testrun-23332/test-s3fs-ALT.txt [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate (_ssl.c:1056)
moved file content-type is not as expected expected:text/plain got:
```

I am not sure how this succeeded before but using HTTP resolve this.

petrus-v:
> I am not sure how this succeeded before but using HTTP resolve this.

Just an idea in reading this PR (I haven't tools to open the keystore.jks on my local machine) I'm wondering if the  keystore.jks could contains self signed  ssl certificat for localhost that could be used for testing secure things which coud expire at some point ?

gaul:
The tests use a self-signed certificate which succeeds for newer distributions but fails for older ones.  I suspect that the certificate that Java generates is somehow incompatible with the older Python implementation.  There could be other solutions to this but using HTTP provides a slight performance advantage so maybe not worth investigating.

petrus-v:
ok thanks for reply !

gaul:
Duplicate of #1812?

ggtakec:
@gaul
Yes, this issue is the same as #1812.
I've been investigating (and still investigating) a lot, but I still don't know the cause.

Could it be the cause of keystore.jks?
(I'm not very familiar with it)
I also tested renewal of CA certificate(package) and everything I can think, but the result was the same.

gaul:
Duplicate of #1812.

PR #1829, opened 2022-01-03
Add missing mailcap for openSUSE CI

gaul:
References #1828.


gaul:
This is incorrect -- I believe `/etc/mime.types` should automatically come from aaa_base.

Also I added fedora:35 and ubuntu:21.10 to the required actions and removed fedora:34 which should clean up some CI issues.

PR #1830, opened 2022-01-03
Add media-types for Debian Buster and Ubuntu 18.04

gaul:
This resolves the missing MIME types issue.  References #1828.


gaul:
Looks like this only exists in Bullseye and newer.

PR #1832, opened 2022-01-03
Use HTTP to avoid unneeded overhead in tests

gaul:
This also may resolve the MIME type issues seen on older Linux CI.
References #1828.


ggtakec:
Is it possible to separate http and https by CI runner(OS)?
I am worried that all OS will not check the connection with https (TLS).
I think that we should test using https.

gaul:
Let's defer the HTTP optimization for now since #1845 can address the CI failure.

PR #1834, opened 2022-01-03
Install default-jdk-headless instead of default-jdk

gaul:
This reduces the dependencies installed.


gaul:
For ubuntu:21.10 this reduces run-time from 1m 20s to 1m 8s and the logs change from:

```
1 upgraded, 283 newly installed, 0 to remove and 0 not upgraded.
Need to get 580 MB of archives.
After this operation, 1690 MB of additional disk space will be used.
```

to:

```
1 upgraded, 205 newly installed, 0 to remove and 0 not upgraded.
Need to get 534 MB of archives.
After this operation, 1502 MB of additional disk space will be used.
```

ggtakec:
The reply was late, thank you for merging.

PR #1835, opened 2022-01-04
Install python2-minimal instead of python2

gaul:
This reduces the dependencies installed.


gaul:
This doesn't seem to avoid the dependency on python2 (not minimal).

gaul:
Superseded by #1837 and #1838.

PR #1838, opened 2022-01-04
Make ut_test compatible with Python 3

gaul:
This may allow removing Python 2 on newer distros.


gaul:
Not sure if this will work on all CI runners.  We might need to merge #1832 first.  We also may need to specify Python 3 explicitly although I think there may be some issue with our older macOS runner (newer ones are incompatible with FUSE).

gaul:
After merging #1837 and this PR we can remove the dependency on Python 2.

ggtakec (approved):

ggtakec:
Thank you, this will be merged soon.

PR #1839, opened 2022-01-04
Use JRE instead of JDK for Debian-based distros

gaul:
This reduces the dependencies installed.  It does not appear that
Fedora-based distros support this.


gaul:
This reduces run-time from ubuntu:21.10 package install time from 1m17s to 57s and the logs change from:

```
1 upgraded, 205 newly installed, 0 to remove and 0 not upgraded.
Need to get 534 MB of archives.
After this operation, 1502 MB of additional disk space will be used.
```

to:

```
1 upgraded, 203 newly installed, 0 to remove and 0 not upgraded.
Need to get 317 MB of archives.
After this operation, 1275 MB of additional disk space will be used.
```

ggtakec (approved):

PR #1845, opened 2022-01-08
Generate S3Proxy SSL certificate during tests

### LINKED ISSUE #1812: Not work --no-verify-ssl option for aws cli on some Runner
### Additional Information
In some GithubActions Runners(`ubuntu 18.04`, `debian buster`, `opensuse/leap 15`), the aws command call is in error.

The error is:
```
SSL validation failed for https://127.0.0.1:8080/s3fs-integration-test/testrun-30317/test-s3fs-ALT.txt [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:852)
```

The aws command is called with `--no-verify-ssl` but receives `CERTIFICATE_VERIFY_FAILED`.

I've tried some workarounds, but I still can't solve them, so I'll report them as Issue.
- Specify `--ca-bundle` option
- Specify `AWS_CA_BUNDLE` environment
- Upgrade `certifi` pip pacakge
- Reinstall `ca-certificates` and execute `update-ca-certificates`
- etc

I've tried the above, but it still doesn't work.
I hope someone knew something ...

gaul:
Also provide CA bundle to AWS CLI to work around CI failures instead
of ignoring errors.  Fixes #1812.


gaul:
This should work now.  I replaced the AWS CLI `--no-verify-ssl` with `--ca-bundle`.  I believe the underlying error is some Python certifi package aws/aws-cli#1545.  But it seems better to trust the specific CA cert and maybe we should add a similar option to s3fs instead of `-o no_check_certificate`?

gaul:
Needed to rebase to get bullseye config.

ggtakec:
We may be able to use the `CURL_CA_BUNDLE` environment variable when specifying a CA certificate.
`--no-verify-ssl` (`--insecure` with curl command) is not provided as an environment variable and should be implemented.
But we need to be careful because those option(env) will not be safe.

ggtakec (approved):
Thanks for this greate fix!

PR #1851, opened 2022-01-09
Use setup-java instead of distribution packages

gaul:
This may be more efficient to install.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,13 @@ jobs:
       - name: Checkout source code
         uses: actions/checkout@v2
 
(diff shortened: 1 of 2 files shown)


ggtakec:
`actions/setup-java` looks good.

gaul:
This always fails on opensuse/leap:15 with:

```
OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "sh": executable file not found in $PATH: unknown
Error: Process completed with exit code 126.
```

ggtakec:
I also checked your PR code, but I couldn't figure out why the `sh` wasn't found.
It's hard to understand why docker exec can't even find sh.
After running actions/setup-java, I got the same error no matter what command run.

PR #1855, opened 2022-01-12
Windows compilation support

### LINKED ISSUE #728: Build s3fs with WinFSP
## Details about issue
About trials to be able to build and execute with WinFSP/Cygwin.
The beginning was PR #712 from @chrilith, and wrote comments by @billziss-gh who is it's author and @feoff3.
I thought it would be better to record it as this Issue of s3fs rather than continuing to follow Issue with PR comments.

- **Google group**  
https://groups.google.com/forum/#!topic/winfsp/xXe6M_MoW8w
- **WinFSP**  
https://github.com/billziss-gh/winfsp
- **First Trigger**  
#712 (please see this PR)
- **Current Issue** on WinFSP  
https://github.com/billziss-gh/winfsp/issues/143

We will actively participate if we can help, and we are happy to help s3fs work on windows.

siketyan:
Closes #728 

### Details
Added support to compile s3fs-fuse on/for Windows.
It requires MSYS2 environment.
To complete the compilation, it was needed to put polyfills for functions that do not exist in MSYS2.
This pull request also adds the polyfills tho.

Confirmed that works:
  - Mounting to a drive (`.\s3fs.exe [bucket] "S:" -f -o uid=-1 -o gid=-1 -o umask=0000 -o mp_umask=0000`)
  - Listing, stating, reading, and writing files
  - Naming files/directories in CJK languages (Unicode support confirmed)

Known issues:
  - MSYS2 shell is required to execute because s3fs-fuse try to find AWS credentials only from `/home/$USER/.aws/credentials`, which does not exist on Windows
 
This pull request does not propose full support for Windows.
This is just a initial experiment to port this project to Windows, a not *nix operating system.
Note that the changes will not break any backward compatibility on *nix platforms.


configure.ac
```diff
@@ -19,11 +19,11 @@
 ######################################################################
 dnl Process this file with autoconf to produce a configure script.
 
-AC_PREREQ(2.59)
-AC_INIT(s3fs, 1.90)
-AC_CONFIG_HEADER([config.h])
+AC_PREREQ([2.71])
```
gaul, line 22: This change causes CI to fail:

```
configure.ac:22: error: Autoconf version 2.71 or higher is required
```

Do you need the latest version?  Not even Fedora 35 packages 2.71.  I think we could change to 2.69 if necessary which both CentOS 7 and Ubuntu 18.04 offer.

src/strptime.h
```diff
@@ -0,0 +1,50 @@
+
+/*
+ * Polyfill for strptime function
```
gaul, line 3: Are there any other alternatives to this polyfill, e.g., https://docs.microsoft.com/en-us/answers/questions/572611/strptime-alternative-in-mingw-vs2019.html ?

src/strcasestr.h
```diff
@@ -0,0 +1,45 @@
+
+/*
+ * Polyfill for strcasestr function
```
gaul, line 3: It appears that the state of case-insensitive string comparisons is not good, even in C++20.  :disappointed:

gaul (changes requested):
Thanks for your PR!  This is a long-desired feature by several users.  I don't have any experience with Windows -- how does it actually interact with FUSE?

src/strcasestr.h
gaul: Actually, this should be part of C89 and C99?

gaul: We only use `strcasestr` once in `MapPutErrorResponse` -- maybe it is better to just lowercase the entire string or use the correct XML parsing functions to avoid this polyfill?

gaul:
> Closes #728

Is it correct to say that this closes the issue?  Can this work end-to-end today or are only the compilation issues addressed?

src/strptime.h
gaul: Perhaps we can call `std::get_time` instead?  This only exists on C++11 and s3fs only requires C++03 but it seems better for the win32 port to use this mode instead of importing this code.

siketyan:
> how does it actually interact with FUSE?

TL;DR: s3fs-fuse -> [WinFsp](https://github.com/billziss-gh/winfsp) -> [winfuse](https://github.com/billziss-gh/winfuse) -> Windows

As you probably know, Windows does not have FUSE API.
However it can have a virtual filesystem using "User Mode Driver" such as [Dokan](https://github.com/dokan-dev/dokany) .
Of course s3fs-fuse is developed for fuse (libfuse), we needed to create a compat layer between FUSE and the driver.
WinFsp, which is also a user-mode driver, is the solution to have the compatibility (https://github.com/billziss-gh/winfsp/blob/master/inc/fuse/fuse.h) .
Therefore we can port s3fs-fuse to Windows by injecting WinFsp libarary as a FUSE implementation such as libfuse.

> Is it correct to say that this closes the issue? Can this work end-to-end today or are only the compilation issues addressed?

Compiling and running s3fs-fuse on/for Windows can be done at HEAD of this branch.
I think resolving the known issues on Windows or creating an installer for the OS is not in the scope of the issue.

src/strcasestr.h
siketyan: This PR is scoped to porting s3fs-fuse to Windows; I think improving existing code should be another issue 🤔

gaul: This isn't an improvement; it is working around win32 lack of support.  I opened #1858 to address the `strcasestr` comment but expect you to address the `strptime` comment.  Importing a bunch of random code imposes a maintenance cost and it is your responsibility to minimize this.

gaul:
> > Is it correct to say that this closes the issue? Can this work end-to-end today or are only the compilation issues addressed?
> 
> Compiling and running s3fs-fuse on/for Windows can be done at HEAD of this branch. I think resolving the known issues on Windows or creating an installer for the OS is not in the scope of the issue.

I am happy to merge the win32 compatibility fixes after you address the broken CI.  But I still don't understand if this commit alone allows mounting s3fs on Windows, e.g., from a user's perspective, can they use s3fs when following the compilation instructions?  As for an installer, this is probably something best hosted in an external repository, at least initially.  But we would be happy to link to the installation instructions.

configure.ac
siketyan: Reverted to 2.59 at 478266f .
The warning was:
```
$ ./autogen.sh
--- Make commit hash file -------
--- Finished commit hash file ---
--- Start autotools -------------
configure.ac:24: warning: The macro `AC_CONFIG_HEADER' is obsolete.
configure.ac:24: You should run autoupdate.
../autoconf-2.71/lib/autoconf/status.m4:719: AC_CONFIG_HEADER is expanded from...
configure.ac:24: the top level
configure.ac:26: warning: The macro `AC_CANONICAL_SYSTEM' is obsolete.
configure.ac:26: You should run autoupdate.
../autoconf-2.71/lib/autoconf/general.m4:2081: AC_CANONICAL_SYSTEM is expanded from...
configure.ac:26: the top level
--- Finished autotools ----------
```

src/strptime.h
siketyan: Replaced the polyfill with the code at d5f1101 .

src/strcasestr.h
siketyan: > Importing a bunch of random code imposes a maintenance cost and it is your responsibility to minimize this.

I agree with this.
So should I merge #1858 and delete this polyfill after you merged the PR?

siketyan: Finally I merged #1858 into HEAD of this PR. Merge this PR after #1858 was merged.

siketyan:
✅ I pushed some commits and got a green in [this workflow run](https://github.com/siketyan/s3fs-fuse/actions/runs/1693583801)!

ggtakec:
It's nice to be able to build for WinFSP :-) 

But, there seems to be one thing that needs to be confirmed.
s3fs has an option(use_cache) for using the cache.
If you use this option, s3fs may not work with the Windows file system.(Because s3fs with this option uses a hard link)

Could you try to run s3fs with this option?

siketyan:
@ggtakec 
It worked with use_cache option enabled as same as disabled.
Command line:
```
./s3fs.exe s6n-jp-docs "S:" -f -o uid=-1 -o gid=-1 -o umask=0000 -o mp_umask=0000 -o aws_credentials_file="C:\\Users\\Siketyan\\.aws\\credentials" -o tmpdir="C:\\Users\\Siketyan\\AppData\\Local\\Temp" -o use_sse -o use_cache
```

Note:
- `aws_credentials_file` is a option I created on another branch [commit diff](https://github.com/siketyan/s3fs-fuse/commit/5bb4d85c7a4bc9aa89ee35bc617164b03a8c8215) because s3fs-fuse tries to load .aws/credentials only from $HOME, which will be "/home/$USER" on Windows where does not exist as a path.
- Reason of specifying `tmpdir` is the same with above.
- It is required to prepare `ca-bundle.crt` then set the path to a environment variable `CURL_CA_BUNDLE`.

siketyan:
Maybe this logs helps you to clear that works :)
```
2022-01-14T10:00:00.727Z [INF]       cache.cpp:AddSymlink(687): add symbolic link cache entry[path=/, value=]
2022-01-14T10:00:00.738Z [INF]       cache.cpp:AddStat(372): add stat cache entry[path=/.]
2022-01-14T10:00:00.738Z [DBG] cache.cpp:GetStat(294): stat cache hit [path=/.][time=165417.165917900][hit count=0]
```

ggtakec:
@siketyan 
_`use_cache` option requires an argument, but it doesn't look like it's on your command line above.(Is it a copy & paste mistake?)_

`use_cache` caches the file content, not the file's stats information.
In the source code, it is the following line.
https://github.com/s3fs-fuse/s3fs-fuse/blob/master/src/fdcache_entity.cpp#L335
A hard link is created to this file cache when you upload a file.

NTFS supports the hard link, so it seemed like my misunderstanding.(Maybe it's working fine.)

I don't think the `aws_credentials_file` option is needed. Instead of that your fix, you should be able to use the `passwd_file` option.
(Maybe you don't even need `tmpdir`, you can do the same with ` use_cache`.)

siketyan:
@ggtakec 
> `use_cache` option requires an argument,

I am sorry, it was my misunderstanding.
And confirmed s3fs-fuse also works with `use_cache=(dir)`, and the cached (mirror) file looks a hard link to the original (see the screenshot below)!

image

> I don't think the aws_credentials_file option is needed. Instead of that your fix, you should be able to use the passwd_file option.
I want to use .aws/credentials, not passwd_file. These files have the different format.
Reading the file from C:\Users\$USER\.aws\credentials will be another solution to address this.

ggtakec:
> And confirmed s3fs-fuse also works with use_cache=(dir), and the cached (mirror) file looks a hard link to the original (see the screenshot below)!

There seems to be no problem.

@gaul
Could you tell me your opinion about supporting the configure file (.aws/credentials) used by the AWS CLI?
(It's an INI file format, not difficult)
If we are going to support it, we can also read region from .aws/config.
I will post another Issue about this, and we talk about this.( #1860 )

src/strptime.h
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * This source code is from musl library, located at https://git.musl-libc.org/cgit/musl/tree/src/time/strptime.c .
+ */
+
+/*
+ * musl as a whole is licensed under the following standard MIT license:
+ * 
+ * ----------------------------------------------------------------------
+ * Copyright © 2005-2020 Rich Felker, et al.
```
gaul, line 12: Is this copyright header still valid?

src/strptime.h
```diff
@@ -0,0 +1,50 @@
...
+#ifndef S3FS_STRPTIME_H_
+#define S3FS_STRPTIME_H_
+
+#include <stdlib.h>
+#include <langinfo.h>
+#include <time.h>
+#include <ctype.h>
+#include <stddef.h>
+#include <string.h>
+#include <strings.h>
```
gaul, line 44: Do these `#include`s still make sense?

src/Makefile.am
```diff
@@ -90,8 +96,14 @@ test_page_list_SOURCES = \
     s3fs_logger.cpp \
     string_util.cpp \
     test_page_list.cpp
+if MSYS
+    test_page_list_SOURCES += strptime.cpp
+endif
 
 test_string_util_SOURCES = string_util.cpp test_string_util.cpp s3fs_logger.cpp
+if MSYS
+    test_string_util_SOURCES += strptime.cpp
```
gaul, line 105: Does it make more sense to add the `strptime.cpp` polyfill to `string_util.cpp`?  This will reduce the amount of Windows conditionals.

gaul (changes requested):
Please rebase to the latest master.

src/strptime.h
siketyan: Nope, I forgot to delete them.
The header file is deleted at d8092b4, resolving this.

src/strptime.h
siketyan: These also thing I forgot to delete and deleted at d8092b4.

src/Makefile.am
siketyan: Moved the polyfill into `string_util.cpp` at d8092b4.

siketyan:
@gaul 
I rebased the branch onto `master`, resolving the requested changes.

src/s3fs.cpp
```diff
@@ -2336,6 +2336,12 @@ static int s3fs_statfs(const char* _path, struct statvfs* stbuf)
     stbuf->f_bfree  = stbuf->f_blocks;
     stbuf->f_bavail = stbuf->f_blocks;
     stbuf->f_namemax = NAME_MAX;
+#ifdef __MSYS__
+    // WinFsp resolves the free space from f_bfree * f_frsize, and the total space from f_blocks * f_frsize (in bytes).
+    stbuf->f_frsize = stbuf->f_bsize;
+    stbuf->f_blocks = INT32_MAX;
+    stbuf->f_bfree = INT32_MAX;
+#endif
```
gaul, line 2344: Should this have `#else` to avoid setting `f_blocks` and `f_bfree` more than once?  You need to shuffle the existing code to do this.

gaul (changes requested):
Almost there!

src/s3fs.cpp
siketyan: Fixed at 1e12274.

gaul:
Thank you for your contribution @siketyan!  I intended to squash history but forgot.  :disappointed:

PR #1857, opened 2022-01-12
Prefer = over == for older shell compatibility

gaul:


skepticalwaves:
If you're going to do this, should probably just add shellcheck in POSIX mode to check everything.

gaul:
I looked at shellcheck and https://github.com/gaul/s3fs-fuse/pull/new/shellcheck is my work-in-progress from a few years ago.  It did not seem to find any really useful problems but might help you if you want to take this further.  I suspect it is best to add ignores for all the existing issues and fix them individually.

ggtakec (approved):

PR #1858, opened 2022-01-13
Use XML parsing with PUT HTTP 200 responses

gaul:
This works around the missing `strcasestr` on win32.  References #728.


ggtakec (approved):

ggtakec:
@gaul I commented on #728 how to use WSL2 instead of WinFSP, so please refer to that as well.

PR #1862, opened 2022-01-16
Added test by a shell script static analysis tool(ShellCheck)

ggtakec:
### Details
s3fs uses Bourne shell(`sh`) and `bash` for build and test.
However, these codes have not been tested except visually and directly at run time.

In the future, we will add new test and scripts for it, thus these test script codes should be checked/linted.
So I added it to my code to use [ShellChek](https://github.com/koalaman/shellcheck).
I think `ShellCheck` is popular and has a proven track record.

We can run `ShellCheck` with `make shellcheck`, just like `cppcheck`.
In this PR, the all script codes were corrected the errors and warnings detected by `ShellCheck`.
These code changes did not change the original logic.

Some errors and warnings cannot(should not) be fixed, then they have not done.
For that purpose, I use inline comment(`# shellcheck disable=SCxxx`) to suppress the output of `ShellCheck`.
The scripts used in s3fs are build and test, so I think that we won't need to take strict fix(ex. security issues).

The test on `CentOS Stream 8` does not perform `ShellCheck` test because `ShellCheck` cannot be installed from the package.
Also, CentOS 7 ShellCheck is excluded because the version is low and the error(`hGetContents: invalid argument (invalid byte sequence)`) can not be resolved.
(Since we are checking with other OS, I think that it is not a problem not to implement it with `CentOS 8/7`.)

### DIFF
--- a/.github/workflows/linux-ci-helper.sh
+++ b/.github/workflows/linux-ci-helper.sh
@@ -24,7 +24,7 @@ echo "${PRGNAME} [INFO] Start Linux helper for installing packages."
 #-----------------------------------------------------------
 # Common variables
 #-----------------------------------------------------------
-PRGNAME=`basename $0`
+PRGNAME=$(basename "$0")
 
 #-----------------------------------------------------------
 # Parameter check
@@ -40,8 +40,10 @@ fi
 # Container OS variables
 #-----------------------------------------------------------
 CONTAINER_FULLNAME=$1
-CONTAINER_OSNAME=`echo ${CONTAINER_FULLNAME} | sed 's/:/ /g' | awk '{print $1}'`
-CONTAINER_OSVERSION=`echo ${CONTAINER_FULLNAME} | sed 's/:/ /g' | awk '{print $2}'`
+# shellcheck disable=SC2034
+CONTAINER_OSNAME=$(echo "${CONTAINER_FULLNAME}" | sed 's/:/ /g' | awk '{print $1}')
+# shellcheck disable=SC2034
+CONTAINER_OSVERSION=$(echo "${CONTAINER_FULLNAME}" | sed 's/:/ /g' | awk '{print $2}')
 
 #-----------------------------------------------------------
 # Common variables for pip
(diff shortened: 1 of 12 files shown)


.github/workflows/linux-ci-helper.sh
```diff
@@ -148,16 +171,16 @@ fi
 # Update packages (ex. apt-get update -y -qq)
 #
 echo "${PRGNAME} [INFO] Updates."
-${PACKAGE_MANAGER_BIN} ${PACKAGE_UPDATE_OPTIONS}
+/bin/sh -c "${PACKAGE_MANAGER_BIN} ${PACKAGE_UPDATE_OPTIONS}"
```
gaul, line 174: This is a weird change -- why should we invoke a shell to invoke a program?

gaul (approved):
This seems fine in principle but there were too many changes to red everything.  Not too sure about invoking `/bin/sh $PROG`?

.github/workflows/linux-ci-helper.sh
ggtakec: There was no error(by shellcheck) in this line, but I fixed it in the same way as the error( [SC2086](https://github.com/koalaman/shellcheck/wiki/Sc2086) ) for the `${INSTALL_PACKAGES}` variable a few lines down(L:180).
Most of shellcheck's suggestions are variable quotes, so I've fixed this line as well.

If it is not needed to run another shell, I can undo it.
Please tell me your opinion.

ggtakec:
> This seems fine in principle but there were too many changes to read everything. 

Most of the warnings from ShellCheck were one of the following:
There are many corrections, but most of the corrections are the same.
_(In particular, `SC2086` was very popular)_
[SC2004](https://github.com/koalaman/shellcheck/wiki/SC2004)
[SC2006](https://github.com/koalaman/shellcheck/wiki/SC2006)
[SC2046](https://github.com/koalaman/shellcheck/wiki/SC2046)
[SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086)
[SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155)

PR #1867, opened 2022-01-22
Experiment with FUSE 3

gaul:
This increases the maximum write size from 128 KB to 1 MB and supports
some interesting `FUSE_READDIR_PLUS` flags.  Remove stale
`FUSE_CAP_BIG_WRITES` option.


gaul:
This is not a serious proposal but I want to see how much breaks in CI.

gaul:
#1868 will address some of the CI failures but Ubuntu 18.04 and Debian Stretch lack FUSE 3.  macOS has its own issues.

ggtakec:
This fix is noteworthy.
Ultimately, I think it's better to wrap or separate the I/F functions, taking into account the differences between fuse2 and fuse3.
(I'm sure that support for fuse2 will continue)

gaul:
I don't think we need to merge this today -- perhaps after Ubuntu 18.04 and Debian Stretch are EOL (2023).  I am more concerned about any other blockers, e.g., macOS.  Having too much conditional logic to support old FUSE will become complicated.

ggtakec:
When I thought about the support for FUSE3 before, I noticed the difference between FUSE2 and FUSE3 and the difference in the prototype of the I/F function.
However, when the difference due to the parameter was dealt with by the function in s3fs(whether to prepare multiple or avoid it with macro etc.), I thought that it could be dealt with only by changing the following initialization part.
https://github.com/s3fs-fuse/s3fs-fuse/blob/master/src/s3fs.cpp#L5199-L5237

_Eventually, I think we can do that. (but we should still fix other parts now unfortunately...)_

gaul:
@macos-fuse-t says that this will break the userspace macOS FUSE library and presumably the older kernel-based one.  I don't think we should merge this yet.

mesaglio:
fuse2 is not having more updates, we dont need to update fuse version to 3?

ggtakec:
@mesaglio 
We have the intention to support fuse3.
However, this does not mean that fuse2 will no longer be supported, so you will need to clean up your code.
Please give me some time. (I think some fixes are needed before that sorting out)

Bhagyashreek8:
@ggtakec Is there any ETA for fuse3 support ?
cc @nkkashyap

gaul:
@Bhagyashreek8 what advantage does requiring FUSE 3 provide to you?

gaul:
@macos-fuse-t any updates on FUSE-T FUSE 3 support?

Bhagyashreek8:
@gaul To give some background, we build and package s3fs-fuse as custom binary based on different os versions. And this binary would be installed on kubernetes clusters.  
The challenge is ubuntu24 by default comes with fuse3 and s3fs-fuse expects fuse2 and its dependant libs such as libfuse2t. (The error pops up when trying to mount a volume to a workload)

```
Events:
  Type     Reason       Age               From               Message
  ----     ------       ----              ----               -------
  Normal   Scheduled    42s               default-scheduler  Successfully assigned default/test-pod to 10.208.108.143
  Warning  FailedMount  8s (x7 over 41s)  kubelet            MountVolume.SetUp failed for volume "pvc-33474d80-8f50-416c-bcb7-446b274dc7c2" : mount command failed, status: Failure, reason: Error mounting volume: s3fs mount failed: s3fs: error while loading shared libraries: libfuse.so.2: cannot open shared object file: No such file or directory
  ```

Catch is if tried to install fuse2, fuse3 gets cleaned up first and then fuse2 gets installed. This might break some defaults on the OS.

**Comment by mesaglio** [comment]
Adding fuse2 is no longer mantained so features and bug fixes will be on [fuse3](https://github.com/libfuse/libfuse/releases)

**Comment by macos-fuse-t** [comment]
@gaul I need to assess the demand for FUSE-3. It's not difficult to port it to mac but then it won't be compatible with osxfuse and most of the folks want fuse-t as a drop-in replacement.

**Comment by gaul** [comment]
Requires #2626.

**Comment by gaul** [comment]
I'm not sure why this fails with:
```
2025-05-15T17:01:29.1545067Z first line(cache file stat) is different: "4536795:0" != "4536795:26214400"
```

**Comment by ggtakec** [comment]
> I'm not sure why this fails with:
> 
> ```
> 2025-05-15T17:01:29.1545067Z first line(cache file stat) is different: "4536795:0" != "4536795:26214400"
> ```

I checked the log of the failure, but I was unable to identify the cause from this log alone.
However, if I imagine the meaning of this(`4536795:0`), it seems to be the cache file information immediately after the file was created (file size=0).
If things are normal, it should be `4536795:26214400`, as this is immediately after the file was uploaded.
I think we need to increase the debug level and check more detailed logs to find the cause of this.

**Comment by Bhagyashreek8** [comment]
Hi @ggtakec @gaul Any plans to have fuse3 support in near time?

**Comment by gaul** [comment]
@macos-fuse-t Any updates on FUSE3 support?  Several Linux distributions want to drop FUSE2 and I prefer not to support both in s3fs.

**Comment by alexfs** [comment]
@gaul At this moment I don't have concrete plans to implement FUSE3. Truth to be told I'm not asked about it too often. But if there would be more interest, I'd love to do it at some point.

**Comment by gaul** [comment]
I added some hacks to support both FUSE 2 and 3.  I dislike these but this seems like the best we can do given the limited state of macOS.  This is ready to review but the Valgrind build seems to fail due to #2709.

**Comment by ggtakec** [comment]
I will try to find reason and fix about #2709 as soon as possible.

@gaul
Please consider the following points regarding FUSE3 compatibility.
(This PR code is fine(no problem is occured), but be aware that we will need to make changes later.)

In FUSE2, functions that don't accept `fuse_file_info*` have a process that checks permissions based on the path, as shown below.
```
static int s3fs_chmod(const char* _path, mode_t mode)
{
    ...
    ...
    if(0 != (result = check_parent_object_access(path, X_OK))){
        return result;
    }
    if(0 != (result = check_object_owner(path, &stbuf))){
        return result;
    }
```

In contrast, when receiving fuse_file_info*, there is actually no need to perform these steps.
The reason is that s3fs was receiving an already open fd, so the permissions have already been checked when it was opened.
Instead, s3fs needs to check whether the fd in this `fuse_file_info*` is an open file in s3fs.
For example, the following function does this:
```
static int s3fs_read(const char* _path, char* buf, size_t size, off_t offset, struct fuse_file_info* fi)
{
    ...
    ...
    AutoFdEntity autoent;
    FdEntity*    ent;
    if(nullptr == (ent = autoent.GetExistFdEntity(path, static_cast<int>(fi->fh)))){
        S3FS_PRN_ERR("could not find opened pseudo_fd(=%llu) for path(%s)", (unsigned long long)(fi->fh), path);
        return -EIO;
    }
```
I hope that with FUSE3 support, this permission check (on the s3fs side) will no longer be necessary.

Note: We can still make code for this suggestion this after this PR is merged.

If you would like to add this suggestion to this PR, I can send you a PR with the fixes, wrapping the above common processing into a utility function that is easier to call.
Which way is the best order for you to marge the changes?

**Comment by gaul** [comment]
Let me add the changes in this PR.

**Comment by gaul** [comment]
I'm not sure that this is correct based on `/usr/include/fuse3/fuse.h`:

```
    /** Change the permission bits of a file
     *
     * `fi` will always be NULL if the file is not currently open, but
     * may also be NULL if the file is open.
     */
    int (*chmod) (const char *, mode_t, struct fuse_file_info *fi);
```

ggtakec:
@gaul 
It seems that `fi` (`fuse_file_info*`) can sometimes be null even if the target file is open before calling chmod(or other).
(Worst of all, in my environment, it never became anything other than null.)
As explained by fuse, `it's sometimes null when the file isn't open, and sometimes null even when the file is open.`, this seems to be correct.

It appears that this isn't dependent on the fuse implementation, but on the implementation of VFS, etc. (This opinion may not be correct.)
In my environment, `fchmod(fd)` references the `fd` of the opened file to identify the inode, but the `VFS`->`filesystem (FUSE)` call path doesn't seem to pass file(open handle).

I couldn't find an environment(or calling method) that would reliably set `fi` to a non-null value.

In any case, we'll be including both null and non-null implementations of fi in FUSE3 support. (for values ​​other than null can be implemented later.)

_Unfortunately, I don't think we'll be able to achieve the expected results (performance) in my current environment(and comparable environments)._

ggtakec:
@gaul I will add these explicit handlings(using "fi") to FUSE3 after release.

gaul:
> @gaul At this moment I don't have concrete plans to implement FUSE3. Truth to be told I'm not asked about it too often. But if there would be more interest, I'd love to do it at some point.

@macos-fuse-t s3fs 1.96 moved to FUSE 3 but maintained FUSE 2 compatibility but I don't know how long we can do this.  Ubuntu has an initiative to upgrade all FUSE clients from 2 to 3: https://ubuntu-archive-team.ubuntu.com/transitions/html/fuse-to-fuse3.html#good,bad,partial,unknown,!notintesting and it seems like half of the packages depend on FUSE 3 now.

macos-fuse-t:
Added libfuse3 support in the latest fuse-t release

PR #1868, opened 2022-01-23
Handle UTIME_NOW and UTIME_OMIT special values

gaul:
FUSE 3 will require this behavior.  References #1159.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -45,6 +45,7 @@
 #include "string_util.h"
 #include "s3fs_auth.h"
 #include "s3fs_help.h"
+#include "s3fs_util.h"
 #include "mpu_util.h"
 
 //-------------------------------------------------------------------
@@ -1927,6 +1928,17 @@ static int s3fs_chown_nocopy(const char* _path, uid_t uid, gid_t gid)
     return result;
 }
 
+static timespec handle_utimens_special_values(timespec ts, timespec now, timespec orig)
+{
+    if(ts.tv_nsec == UTIME_NOW){
+        return now;
+    }else if(ts.tv_nsec == UTIME_OMIT){
+        return orig;
+    }else{
+        return ts;
+    }
+}
+
 static int s3fs_utimens(const char* _path, const struct timespec ts[2])
 {
     WTF8_ENCODE(path)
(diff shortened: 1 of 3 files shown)


src/s3fs.cpp
```diff
@@ -1940,6 +1951,11 @@ static int s3fs_utimens(const char* _path, const struct timespec ts[2])
 
     S3FS_PRN_INFO("[path=%s][mtime=%lld][ctime/atime=%lld]", path, static_cast<long long>(ts[1].tv_sec), static_cast<long long>(ts[0].tv_sec));
 
+    struct timeval now;
+    gettimeofday(&now, /*tz=*/ NULL);
```
ggtakec, line 1955: I think this should use `s3fs_clock_gettime` instead of `gettimeofday`.

src/s3fs.cpp
```diff
@@ -2050,6 +2066,11 @@ static int s3fs_utimens_nocopy(const char* _path, const struct timespec ts[2])
 
     S3FS_PRN_INFO1("[path=%s][mtime=%lld][atime/ctime=%lld]", path, static_cast<long long>(ts[1].tv_sec), static_cast<long long>(ts[0].tv_sec));
 
+    struct timeval now;
+    gettimeofday(&now, /*tz=*/ NULL);
```
ggtakec, line 2070: Same as the comment above.

ggtakec (commented):
Please check my comment.

src/s3fs.cpp
gaul: `s3fs_clock_gettime` is local to `cache.cpp` and `S3fsLog::GetCurrentTime` does not call it.  I don't understand the advantage of `s3fs_clock_gettime` in this situation?

ggtakec: It was originally caused by POSIX.1-2008, I remember that clock_gettime is recommended over gettimeofday.
If you just want to get the current time like this process, you can use gettimeofday.
But I think it's better to use clock_gettime.
(Although it is different from this case, clock_gettime is faster (option dependent) and more accurate in time measurement etc.)
Also, if necessary, we can remove static from the declaration of s3fs_clock_gettime.

gaul:
macOS has a different field name for the `timespec` values.

ggtakec (approved):

PR #1869, opened 2022-01-23
Convert S3FS_LOW_LOGPRN from a macro to a function

gaul:
This shrinks the binary size from 770 to 540 KB and reduces compile
times.


gaul:
I don't think that this PR causes the cppcheck failures, rather, it makes the code simpler so that cppcheck can analyze it.  Older versions appear to have false positives.

ggtakec (approved):

PR #1871, opened 2022-01-24
Always call clock_gettime(2)

gaul:
e01ded9e27f41af45e8614860b29164d13c0101e introduced this compatibility
shim but macOS 10.12 (2016) added this:
https://stackoverflow.com/a/39801564 .  Also remove fallback to
`time(3)` which loses precision.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Many systems provide pre-built packages:
   sudo zypper install s3fs
   ```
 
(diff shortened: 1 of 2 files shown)


gaul:
Even my MacBook Air 2010 can run macOS 10.13 so I don't think we should retain compatibility with 10.11 and earlier.

ggtakec (approved):
I understand that macos before 10.12 is not supported.

Issue #1872, opened 2022-01-25
S3fs Mounted folder show invalidate timestamp as "1969-12-31 23:59:59"


tomato0401:
s3fs version:  v1.90
s3 backend:   minio-RELEASE.2019-10-12T01-39-57Z

I use a minio instance for s3 storage, when i mount it to local use s3fs command, some folder shows invalid timestamp. That cause my python code could not import libs from those folders. 
```
drwxrwx--- 1 user user    0 1969-12-31 23:59:59.000000000 +0000 configs
```
I found that in this file "https://github.com/s3fs-fuse/s3fs-fuse/blob/v1.90/src/metaheader.cpp" get_time get_ctime get_mtime return -1 while "x-amz-meta-xxxx" is not provided.

Why should these method return -1 but not 0, some where -1 is consided as wrong time and cause many unnormal action.

gaul:
This seems like an issue where -1 should indicate the absence of a timestamp on the object but should reflect as 0 to the application.  This should be an easy fix to `s3fs_getattr`; would you like to try fixing this?  Also add a test to `test_read_external_object` for the metadata.

ggtakec:
There was a problem with the processing(this issue), and I posted new PR(#1877) to fix this.

Note:
These functions return `-1`, which reason is used in the internal processing to determine whether or not the object does not have a headers(x-meta-***).
In that PR fix, it set `0` when getting `-1` in function which does not require judgment.

If the PR is merged without any problem, I would be grateful if you could confirm it.

gaul:
Closing as fixed.  Please reopen if you can reproduce these symptoms with 1.93.

PR #1877, opened 2022-01-29
Fixed the fault tolerance when time stamp getting fails

ggtakec:
#1872

### Details
Fixed a problem in processing when get_atime(ctime/mtime) failed to get the time.
In convert_header_to_stat, reset 0 if the above function returns -1.

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -786,6 +786,10 @@ bool convert_header_to_stat(const char* path, const headers_t& meta, struct stat
     if(pst->st_mtime < 0){
         pst->st_mtime = 0L;
     }else{
+        if(mtime.tv_sec < 0){
+            mtime.tv_sec  = 0;
+            mtime.tv_nsec = 0;
+        }
 #if defined(__APPLE__)
         pst->st_mtime = mtime.tv_sec;
 #else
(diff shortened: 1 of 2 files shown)


src/cache.cpp
```diff
@@ -786,6 +786,10 @@ bool convert_header_to_stat(const char* path, const headers_t& meta, struct stat
     if(pst->st_mtime < 0){
         pst->st_mtime = 0L;
     }else{
+        if(mtime.tv_sec < 0){
+            mtime.tv_sec  = 0;
+            mtime.tv_nsec = 0;
```
gaul, line 791: Can you add a test to `test_read_external_object`?

ggtakec: Yes, I added

ggtakec:
Added a test case called `test_read_external_dir_object`.

This bug occurs when `a/c/ mtime` is acquired by `stat`(or `ls`, etc) of a directory that s3fs cannot recognize.
You can check the occurrence of the problem by following the steps below:
1. Upload a new file(ex. `subdir/testfile`) containing subdirectories with `aws cli`.
2. At this point, there is no object(ex. `subdir/` for s3fs) indicating the subdirectory.
3. Run the `stat` command to the subdirectory via s3fs.
4. As a result, there was a problem that got `-1` of unixtime.

The time notation(`stat` command) for `unixtime=-1` is as follows:
```
Access: 1969-12-31 23:59:59.000000000 +0000
Modify: 1969-12-31 23:59:59.000000000 +0000
Change: 1969-12-31 23:59:59.000000000 +0000
```
With this PR fix, `a/c/mtime` for non-existent directory objects will be `unixtime=0`.

Thus, I added a test case to do this inspection.
Please be careful for this test case only checks for one `unitxtime=-1` value. (Inspects the string `1969-12-31 23:59:59.000000000`.)

PR #1881, opened 2022-01-30
Use CLOCK_REALTIME for UTIME_NOW

gaul:
Previously `s3fs_utimens` used `CLOCK_MONOTONIC_COARSE` which was not
1970-based.  Found via pjdfstest.  References #1589.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1966,7 +1966,7 @@ static int s3fs_utimens(const char* _path, const struct timespec ts[2])
     }
 
     struct timespec now;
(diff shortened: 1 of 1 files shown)


ggtakec (approved):
Thanks.
For utimens functions, I think there is no problem with this fix.

Please keep `SetStatCacheTime(...)` and `S3fsLog::GetCurrentTime()` as `CLOCK_MONOTONIC(_COARSE)` as much as possible.
(These are the cases where speed is required for time stamping and the case where the difference between times where accuracy is not required is calculated, thus we can use `MONOTONIC`)

PR #1882, opened 2022-01-30
Run passing tests from pjdfstest

gaul:
This downloads a tarball by hash instead of using a submodule.
References #1589.

### DIFF
--- a/.github/workflows/linux-ci-helper.sh
+++ b/.github/workflows/linux-ci-helper.sh
@@ -136,28 +136,28 @@ elif [ "${CONTAINER_FULLNAME}" = "rockylinux:9" ]; then
     #
     PACKAGE_INSTALL_ADDITIONAL_OPTIONS="--allowerasing"
 
-    INSTALL_PACKAGES="clang-tools-extra curl-devel fuse fuse-devel gcc libstdc++-devel gcc-c++ glibc-langpack-en java-17-openjdk-headless jq libxml2-devel mailcap git automake make openssl openssl-devel attr diffutils curl python3 procps unzip xz https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm"
+    INSTALL_PACKAGES="clang-tools-extra curl-devel fuse fuse-devel gcc libstdc++-devel gcc-c++ glibc-langpack-en java-17-openjdk-headless jq libxml2-devel mailcap git automake make openssl openssl-devel perl-Test-Harness attr diffutils curl python3 procps unzip xz https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm"
 
 elif [ "${CONTAINER_FULLNAME}" = "rockylinux:8" ]; then
     PACKAGE_MANAGER_BIN="dnf"
     PACKAGE_UPDATE_OPTIONS="update -y -qq"
     PACKAGE_INSTALL_OPTIONS="install -y"
 
-    INSTALL_PACKAGES="clang-tools-extra curl-devel fuse fuse-devel gcc libstdc++-devel gcc-c++ glibc-langpack-en java-17-openjdk-headless jq libxml2-devel mailcap git automake make openssl openssl-devel attr diffutils curl python3 unzip"
+    INSTALL_PACKAGES="clang-tools-extra curl-devel fuse fuse-devel gcc libstdc++-devel gcc-c++ glibc-langpack-en java-17-openjdk-headless jq libxml2-devel mailcap git automake make openssl openssl-devel perl-Test-Harness attr diffutils curl python3 unzip"
 
 elif [ "${CONTAINER_FULLNAME}" = "fedora:41" ]; then
     PACKAGE_MANAGER_BIN="dnf"
     PACKAGE_UPDATE_OPTIONS="update -y -qq"
     PACKAGE_INSTALL_OPTIONS="install -y"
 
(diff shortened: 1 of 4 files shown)


autogen.sh
```diff
@@ -38,6 +38,8 @@ aclocal \
 && automake --add-missing \
 && autoconf
 
+(cd test/pjdfstest && autoreconf -ifs && ./configure)
```
gaul, line 41: Any ideas on what is the correct way to do this?

ggtakec: Setting pjdfstest as a submodule is a bit thoughtful.

I want the s3fs repository to be able to work independently.
I don't think I agree with mixing the pjdfstest build with the s3fs build.
The pjdfstest build is the `environment` required for the s3fs build(and test), and I think it's better to separate this part.

If you want to use pjdfstest in s3fs test, what about the following method?
- In github actions, git clone pjdfstest under the test directory(this can be done with ci.yml. You can prepare a helper script if necessary)
- And in the pjdfstest directory, do that build

How about performing this step only during CI?

Anyone can deploy it on localhost other than CI by preparing a dedicated helper script.
(Alternatively, you may prepare another repository(ex. s3fs-fuse/s3fs-pjdfstest, etc.) and dedicate it. But this is still an idea I just come up with, so it may be wrong.)

ggtakec: I came up with another one.
You may want to call the script as a PHONY target in the `Makefile (.am)` of the test directory.
(Clone and build pjdfstest in the script)

ggtakec: I made a example. (see proto_pjdfstest branch in my personal repo)
https://github.com/ggtakec/s3fs-fuse/compare/master...proto_pjdfstest

I haven't implemented a "test" using pjdfstest yet...(because I don't know how you want to do it)
This way, pjdfstest can be made independent for now, which is in line with my wishes.

gaul: I reworked this to download the tarball which is cleaner.

gaul:
This seems to show some real failure with `rmdir` and `use_cache`

gaul:
This doubles the CI run-time so we should be careful if we want to merge this PR.

gaul:
CI recently started failing with these kinds of errors:

```
2023-09-14T19:14:38.4260487Z cp(-p) expected times: mtime( 1694718792.836254751 == 1694718792.852005368 ), ctime( 1694718792.836254751 != 1694718792.848255678 ), atime( 1694718792.834613324 == 1694718792.834613324 )
2023-09-14T19:14:38.4260909Z test_mtime_file failed

2023-09-14T19:14:38.4362592Z cp with -p option expected updated ctime: 1000000000.000000000 != 1694718798.132663109 and same mtime: 1000000000.000000000 == 1000000000.000000000, atime: 1000000000.000000000 == 1694718798.133826699
2023-09-14T19:14:38.4363207Z test_update_time_cp_p failed
```

gaul:
macOS runner fails to compile due to a dependency on macOS 13:

```
 pjdfstest.c:848:10: error: 'mkfifoat' is only available on macOS 13.0 or newer [-Werror,-Wunguarded-availability-new]
                rval = mkfifoat(NUM(0), STR(1), (mode_t)NUM(2));
                       ^~~~~~~~
/Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/stat.h:394:9: note: 'mkfifoat' has been marked as being introduced in macOS 13.0 here, but the deployment target is macOS 12.0.0
int     mkfifoat(int, const char *, mode_t) __API_AVAILABLE(macos(13.0), ios(16.0), tvos(16.0), watchos(9.0));
        ^
pjdfstest.c:848:10: note: enclose 'mkfifoat' in a __builtin_available check to silence this warning
                rval = mkfifoat(NUM(0), STR(1), (mode_t)NUM(2));
                       ^~~~~~~~
pjdfstest.c:894:11: error: 'mknodat' is only available on macOS 13.0 or newer [-Werror,-Wunguarded-availability-new]
                        rval = mknodat(NUM(0), STR(1), ntype | NUM(3), dev);
                               ^~~~~~~
/Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/stat.h:395:9: note: 'mknodat' has been marked as being introduced in macOS 13.0 here, but the deployment target is macOS 12.0.0
int     mknodat(int, const char *, mode_t, dev_t) __API_AVAILABLE(macos(13.0), ios(16.0), tvos(16.0), watchos(9.0));
        ^
pjdfstest.c:894:11: note: enclose 'mknodat' in a __builtin_available check to silence this warning
                        rval = mknodat(NUM(0), STR(1), ntype | NUM(3), dev);
                               ^~~~~~~
```

References #2552.

ggtakec:
> macOS runner fails to compile due to a dependency on macOS 13:

Looking at the configure.ac file, it looks like this problem has been addressed.
https://github.com/pjd/pjdfstest/blob/master/configure.ac#L52-L55

And it looks like code that excludes these issues has been committed on MacOS.
https://github.com/pjd/pjdfstest/commit/2dd780aaeeaa422565e5df99f445b9bd4eab226c

Will the result be the same if you re-run configure?

gaul:
This increases test run-time substantially.  Perhaps better as a separate CI task?

ggtakec (approved):
@gaul Could you rebase onto the latest master?
I'll try to merge once it's rebased.

PR #1884, opened 2022-02-05
Propagate return code properly

gaul:
Previously this did not propagate test failures.  A bad rebase
introduced this logic in 495d51113cbe4da63493a51d47dc94145481455a.

PR #1885, opened 2022-02-06
Fixed test_external_directory_creation test when cache enabled

ggtakec:
#1884 

### Details
It is a correction of the defect detected by the correction of #1884.

The `test_external_modification` test checks the reflection of the updated file from the external program (awscli).

However, if s3fs currently has the stat cache and cache directory enabled, this test will fail.
This is because the time value of the file timestamp does not change before and after the change.
When using the stat cache and the file does not update after modifying, the cache hit and reading contents from local cache file.
Then the test is failed.
_To fix this test, use sleep in `test_external_modification`._

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -357,6 +357,17 @@ function test_external_directory_creation {
 function test_external_modification {
     describe "Test external modification to an object ..."
     echo "old" > "${TEST_TEXT_FILE}"
+
+    # [NOTE]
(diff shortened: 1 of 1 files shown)


ggtakec:
@gaul 
I want to POST some PRs after this, so I'll merge this fix.
I'm getting some errors on Github Actions and the impact on other PRs is getting bigger and I'd like to fix them first.

If there is a problem, we will fix it immediately, so please point it out.

PR #1886, opened 2022-02-13
Fixed some Github Actions errors.

ggtakec:
n/a

### Details
Fixed some Github Actions errors.
- Fix knownConditionTrueFalse cppcheck(2.7) error on MacOS
- Fixed package installing failure of appstream download on centos8


ggtakec:
@gaul 
Merge first so that other PR etc. can proceed.
If you have any problems, please point them out. I'll try to fix it right away.

This fix eliminates the error currently occurring in Github Actions.
For CentOS 8, it seems better to proceed with the migration to CentOS 9, but this is a temporary fix.
(see. https://stackoverflow.com/questions/70930615/no-urls-in-mirrorlist-with-yum-on-centos-due-to-appstream )

Issue #1875, opened 2022-01-27
Do we have to flush on s3fs_truncate?


orozery:
We have an application which writes large files (~4GB) in 32MB chunks,
where before each 32MB write, the application is calling `ftruncate` to increase the file size by 32MB.
This maps to `s3fs_truncate`, which flushes the file.
The writing to the S3 backend becomes very inefficient, as the entire file is overwritten every 32MBs written.

I'm wondering if the flush on `s3fs_truncate` can be avoided.
In the man page of ftruncate, I did not see they mention that a flush is guaranteed.

@gaul will be happy to get your thoughts.
Thanks!

gaul:
I believe that s3fs has always `fsync` on `ftruncate`, as far back as 4a30df1ff24aeac92d3ab0f8e84b34df8e70ec3d.  I do not believe that POSIX actually requires this and the slower implementation was only a reflection of s3fs' limited dirty data tracking at that time.  Since `ftruncate` is an uncommon operation we have not optimized it yet.  Could you submit a PR for this?

gaul:
One thing to investigate is whether s3fs calls `ftruncate` on its temporary file so that it can return `ENOSPC` to the application if local storage is too small.

ggtakec:
I think we need to re-check the code that uses the file size after truncate operation.
The file size is used by other operations that occur after truncate. (In various places)

For example, when using a cache file(in the case where there is a directory capacity), it tries to substitute the file size acquisition from the local cache file.
However, if the cache is not used, the current s3fs will have a problem if the size is not reflected in the S3 server side object after calling the truncate operation.

However, we may be able to turn off fsync while the file is open.
(But that doesn't seem to be a simple fix.)

gaul:
If we added performance counters, suggested in #1571, we could modify the integration tests to check the expected number of RPCs.  This would ensure that we are not regressing performance on operations like truncate.

ggtakec:
As you say, performance evaluation for cache is necessary.
(Let's define what should be measured in #1571)

In the first comment of @orozery, there is the following sentence.
> where before each 32MB write, the application is calling ftruncate to increase the file size by 32MB.
> This maps to s3fs_truncate, which flushes the file.

In other words, the application repeates call truncate->write pair.
s3fs(and FUSE) does not know that write will be called after truncate.
That's why in the current implementation, it's flushing at every trauncate...

If you want to avoid this, we will be ables to change s3fs as:
[...]

PR #1631, opened 2021-04-24
Allow truncation of open and modified files

gaul:
Regression introduced in f5bf41cf117e6dfc0d510348cdc1cc11817662ed.
Fixes #1575.

PR #1887, opened 2022-02-13
Fixed not to call Flush even if the file size is increased

ggtakec:
#1875 

### Details
Changed s3fs_truncate function.
This change reduces the number of file uploads if the file size is changed.

On macos, I have found that the truncate call when "size=0" cannot reflect the file size.(This reason is not understood...)
To avoid this, only when "size=0", the flush method is called as before.

Other than that, I found a bug in FdEntity::Open() and fixed it.

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -426,7 +426,7 @@ int FdEntity::Open(const headers_t* pmeta, off_t size, time_t time, int flags, A
         // check only file size(do not need to save cfs and time.
         if(0 <= size && pagelist.Size() != size){
             // truncate temporary file size
-            if(-1 == ftruncate(physical_fd, size)){
+            if(-1 == ftruncate(physical_fd, size) || -1 == fsync(physical_fd)){
                 S3FS_PRN_ERR("failed to truncate temporary file(physical_fd=%d) by errno(%d).", physical_fd, errno);
                 return -errno;
             }
@@ -440,7 +440,7 @@ int FdEntity::Open(const headers_t* pmeta, off_t size, time_t time, int flags, A
         off_t new_size = (0 <= size ? size : size_orgmeta);
         if(pmeta){
             orgmeta  = *pmeta;
-            new_size = get_size(orgmeta);
+            size_orgmeta = get_size(orgmeta);
         }
         if(new_size < size_orgmeta){
             size_orgmeta = new_size;
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -2168,20 +2168,35 @@ static int s3fs_truncate(const char* _path, off_t size)
 
     // Get file information
     if(0 == (result = get_object_attribute(path, NULL, &meta))){
-        // Exists -> Get file(with size)
+        // File exists
+
+        // [NOTE]
+        // If the file exists, the file has already been opened by FUSE before
+        // truncate is called. Then the call below will change the file size.
+        // (When an already open file is changed the file size, FUSE will not
+        // reopen it.)
+        // The Flush is called before this file is closed, so there is no need
+        // to do it here.
+        //
         if(NULL == (ent = autoent.Open(path, &meta, size, -1, O_RDWR, false, true, AutoLock::NONE))){
             S3FS_PRN_ERR("could not open file(%s): errno=%d", path, errno);
             return -EIO;
         }
-        if(0 != (result = ent->Load(0, size, AutoLock::NONE))){
-            S3FS_PRN_ERR("could not download file(%s): result=%d", path, result);
-            return result;
-        }
-
         ent->UpdateCtime();
+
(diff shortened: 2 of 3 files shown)


src/s3fs.cpp
```diff
@@ -2183,20 +2183,35 @@ static int s3fs_truncate(const char* _path, off_t size)
...
-            S3FS_PRN_ERR("could not download file(%s): result=%d", path, result);
-            return result;
-        }
-
         ent->UpdateCtime();
+
+        // [NOTE]
+        // Only for macos, this truncate calls to "size=0" do not reflect size.
+        // The cause is unknown now, but it can be avoided by flushing the file.
+        //
```
gaul, line 2205: Should this be guarded by an `#ifdef __APPLE__`?

src/fdcache_entity.cpp
```diff
@@ -426,7 +426,7 @@ int FdEntity::Open(const headers_t* pmeta, off_t size, time_t time, int flags, A
         // check only file size(do not need to save cfs and time.
         if(0 <= size && pagelist.Size() != size){
             // truncate temporary file size
-            if(-1 == ftruncate(physical_fd, size)){
+            if(-1 == ftruncate(physical_fd, size) || -1 == fsync(physical_fd)){
```
gaul, line 429: Is it necessary to `fsync` locally?

gaul (commented):
This makes sense to me but I have some small questions.

@orozery Can you test your application with this PR?

src/s3fs.cpp
ggtakec: That's a good idea.
I'll fix it soon.

src/fdcache_entity.cpp
ggtakec: Same code has attached it to the code behind this, so I attached it just in case.
It may not be necessary in practice, but since it is only this case and it is a sync to a local file, I think that there is almost no impact on performance.

orozery:
@gaul My team tested this PR and indeed the run time is much better now.

However, a new problem arose now, specifically with this piece of code that you added in #1631 :

```
            // If the file is being modified and it's size is larger than size parameter, it will not be resized.
            off_t cur_size = 0;
            if(ent->GetSize(cur_size) && size <= cur_size){
                size = -1;
            }
```

Basically, our application is doing a final `ftruncate` on the file which actually reduces the file size.
Because of the above code, this `ftruncate` turns to a no-op.
I'm not sure why we need the above logic.

ggtakec:
@orozery Thank you for the survey.

Is the process you are doing as follows?
- File open
- File size increment (truncate or write to tail)
- File size reduction (truncate)
- File close

Is your process working on these operation with just one file open(fd)?
(Or do you have multiple opened fd?)

I think this problem code (probably) assumed that there were multiple opens for the same file and there was an fd that wasn't closed yet.
If you are doing the above, I think this response is inappropriate.

I said in #1875 that there isn't much difference in fuse 2/3, but that was premature. To successfully work around this issue, fuse3 which has a well-defined fd to be truncated may be more convenient.

I will continue to confirm this matter, including investigations.

orozery:
> @orozery Thank you for the survey.
> 
> Is the process you are doing as follows?
> 
> * File open
> * File size increment (truncate or write to tail)
> * File size reduction (truncate)
> * File close
> 

Yes, or more precisely:
1. open
2. REPEAT: increment file size + write to tail
3. reduce file size to offset of last write
4. close

> Is your process working on these operation with just one file open(fd)? (Or do you have multiple opened fd?)
> 

A single fd.

> I think this problem code (probably) assumed that there were multiple opens for the same file and there was an fd that wasn't closed yet. If you are doing the above, I think this response is inappropriate.
> 

But even if you have multiple opens for the same file, the previous logic (before PR #1887) was to always flush, so why not truncate for all open fd's? 

> I said in #1875 that there isn't much difference in fuse 2/3, but that was premature. To successfully work around this issue, fuse3 which has a well-defined fd to be truncated may be more convenient.
> 
> I will continue to confirm this matter, including investigations.
> 
> * Additional comments
>   You can see why this process was added in the test below.
>   https://github.com/s3fs-fuse/s3fs-fuse/blob/master/test/integration-test-main.sh#L1291-L1303

ggtakec:
@orozery Thank you for your comments.
It was late, but I tested it and was able to reproduce the problem.
And I uploaded a new code in #1919.
Please refer to the comment in #1919 for the details of the bug and how to fix it.
If you can test it, test it again.

ggtakec:
@orozery 
#1919 has been merged.
I think this problem is solved, so please try to test it.
If you still have problems, please contact us.

orozery:
@ggtakec initial testing seems good.
I will update if any issues come up.
Thank you!!

ggtakec:
@orozery Thank you for checking.

PR #1888, opened 2022-02-13
Include climits to support musl libc

arkamar:
`PATH_MAX` constant is not visible from any of currently included header files in system with musl libc, where compilation fails with an error below. The constant is defined in `limits.h` which is directly include via `climits` header file.

```
fdcache.cpp: In static member function 'static FILE* FdManager::MakeTempFile()':
fdcache.cpp:381:14: error: 'PATH_MAX' was not declared in this scope
  381 |     char cfn[PATH_MAX];
      |              ^~~~~~~~
```

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -21,6 +21,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cerrno>
+#include <climits>
 #include <unistd.h>
 #include <sys/types.h>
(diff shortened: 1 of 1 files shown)


gaul:
Thank you for your contribution @arkamar!  What is the use case for musl and s3fs -- are you building a minimized Docker image?

arkamar:
Well, I use musl in regular desktop.

Issue #927, opened 2019-01-24
Consider changing `-o notsup_compat_dir` default


gaul:
#571 added an optimization to disable the legacy `dirname_$folder$` checking.  Should s3fs default to the optimization on since most installations should use the `dirname/` format and have an option to check for the legacy folders instead?

ggtakec:
@gaul 
Currently, I agree that I no longer need to search for the `_$folder$` object.

In addition to this, s3fs supports the `no directory object` case.
This is when there is no intermediate path `dir` or `dir/` and there is only a `dir/path` object.
I think we need to continue to support this pattern.

With the fix of #571 , `support_compat_dir flag` is defined in s3fs.cpp. However, this flag is also used to check the `no directory object` above.
So we need to think about the correspondence of the `no directory object` case with this issue as well.

I think we let s3fs not check the `_$folder$` object by default, but a case of `no directory object` should continue checking.

Please let me know your opinion about it.

gaul:
We should also add a script to detect and convert buckets which use the older compat dirs.

PR #1898, opened 2022-02-19
Add "Performance considerations" to manpage

### LINKED ISSUE #1892: Add "Performance considerations" to manpage
Dear S3FS Developer,

what do you think about adding the section below to the manpage?

## PERFORMANCE CONSIDERATIONS
This section discusses settings to improve s3fs performance.

In most cases, backend performance cannot be controlled and is therefore not part of this discussion.

Details of the local storage usage is discussed in "LOCAL STORAGE CONSUMPTION".

### CPU and Memory Consumption
s3fs is a multi-threaded application. Depending on the workload it may use multiple CPUs and a certain amount of memory. You can monitor the CPU and memory consumption with the "top" utility.

### Performance of S3 requests
s3fs provides several options (e.g. "-o multipart_size", "-o parallel_count") to control behaviour and thus indirectly the performance. The possible combinations of these options in conjunction with the various S3 backends are so varied that there is no individual recommendation other than the default values. Improved individual settings can be found by testing and measuring.

The two options "Enable no object cache" ("-o enable_noobj_cache") and "Disable support of alternative directory names" ("-o notsup_compat_dir") can be used to control shared access to the same bucket by different applications:

 * Enable no object cache ("-o enable_noobj_cache")
 
   If a bucket is used exclusively by an s3fs instance, you can enable the cache for non-existent files and directories with "-o enable_noobj_cache". This eliminates repeated requests to check the existence of an object, saving time and possibly money.
 
 * Disable support of alternative directory names ("-o notsup_compat_dir")

   s3fs supports "dir/", "dir" and "dir_$folder$" to map directory names to S3 objects and vice versa.
 
   Some applications use a different naming schema for associating directory names to S3 objects. For example, Apache Hadoop uses the  "dir_$folder$" schema to create S3 objects for directories.

   The option "-o notsup_compat_dir" can be set if all accessing tools use the "dir/" naming schema for directory objects and the bucket does not contain any objects with a different naming scheme. In this case, accessing directory objects saves time and possibly money because alternative schemas are not checked.

CarstenGrohmann:
Add a new section about performance considerations to the manpage and align the description of notsup_compat_dir.

Closes #1892

### DIFF
--- a/doc/man/s3fs.1
+++ b/doc/man/s3fs.1
@@ -332,16 +332,16 @@ This name will be added to logging messages and user agent headers sent by s3fs.
 s3fs complements lack of information about file/directory mode if a file or a directory object does not have x-amz-meta-mode header.
 As default, s3fs does not complements stat information for a object, then the object will not be able to be allowed to list/modify.
 .TP
-\fB\-o\fR notsup_compat_dir (not support compatibility directory naming)
-As default, s3fs supports objects of the directory type as much as possible and recognizes them as directories.
-Objects that can be recognized as directory objects are "dir/", "dir", "dir_$folder$", and there is a file object that does not have a directory object but contains that directory path.
-s3fs needs redundant communication to support all these directory names.
-The object as the directory created by s3fs is "dir/".
-By restricting s3fs to recognize only "dir/" as a directory, communication traffic can be reduced.
-This option is used to give this restriction to s3fs.
-However, if there is a directory object other than "dir/" in the bucket, specifying this option is not recommended.
-s3fs may not be able to recognize the object correctly if an object created by s3fs exists in the bucket.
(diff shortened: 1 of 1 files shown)


ggtakec:
LGTM

@gaul Please give us your opinion.

ggtakec (approved):

gaul (commented):
Thanks for documenting these!  I think there is some question of whether this kind of information should exist in man pages, wikis, or blog posts, but this PR is definitely progress.

We could use more help understanding the performance of s3fs in real user workloads.  You might want to look at #927 and #1901 where we might invert the default behavior of these flags so that users can opt-into of the more compatible but slower paths instead of opting-out into more performant behavior.

PR #1900, opened 2022-02-20
Add test for external object creation

gaul:
This test demonstrates the behavior before and after the stat cache
timeout when using `noobj_cache`.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -374,6 +374,19 @@ function test_external_modification {
     rm -f "${TEST_TEXT_FILE}"
 }
 
+function test_external_creation {
+    describe "Test external creation of an object ..."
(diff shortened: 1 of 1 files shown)


gaul:
Not sure why these are failing.

ggtakec:
The reason maybe `if ps u -p "${S3FS_PID}"`, I think `ps u`(no `-p`) is correct.

gaul:
Unfortunately the integration tests fail locally for me even without this PR:

```
File times do not match:  1645587457 != 1645587458
test_mtime_file failed
...
cp with -p option expected updated ctime: 1000000000 != 1645587460 and same mtime: 1000000000 == 1000000000, atime: 1000
000000 == 1645587460
test_update_time_cp_p failed
...
FAIL: test_mtime_file
FAIL: test_update_time_cp_p
```

ggtakec (approved):
The tests have been successful on Github Actions, so merge these.
Please let me know if you have any information about local failures.

PR #1902, opened 2022-02-20
Use more new file names for every test

gaul:
This makes the tests more robust.  Also fix filename to end in .txt.

### DIFF
--- a/test/test-utils.sh
+++ b/test/test-utils.sh
@@ -248,8 +248,12 @@ function run_suite {
    cd_run_dir "${key_prefix}"
    for t in "${TEST_LIST[@]}"; do
        # Ensure test input name differs every iteration
-       TEST_TEXT_FILE="test-s3fs.txt-${RANDOM}"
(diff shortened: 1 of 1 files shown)


ggtakec (commented):
I confirmed that there was no problem I fixed.

However,  I'm worried about the part of each test function that doesn't delete the test file at the beginning.
Certainly, since the test file used at the end of the test function has been deleted, I know that there is no problem...
(I'm worried that if a test function that I forgot to delete is added, it will bypass the error in the unexpected part)

gaul:
I reverted the logic which removes files.  We should think about how to make the test methods more independent of each other.  For example, we could use a new s3fs instance for every test case (need to check performance) or use AWS CLI to clear the bucket.

ggtakec (approved):

ggtakec:
The number of test files will increase, but what about adding the test function name to the test file name?
By making good use of shell variables, I think that test function can use "file name with function name" in it without being aware of it.
(I try to make it after...)
What I want to avoid is that there is a file that shouldn't exist and the test fails / avoids it.

PR #1899, opened 2022-02-20
Changed LockyLinux 8 instead of CentOS 8

ggtakec:
### Relevant Issue (if applicable)
#1886

### Details
We merged the changes about GithubActions for `CentOS 8` in #1886 to avoid the error, but we are getting the error again. (it seems in relating java packages)
`CentOS 8` is no longer supported and we have the option of migrating to `CentOS Stream` or migrating to `LockyLinux`.
I think that migrating to `LockyLinux 8` is easy and I post this PR for it.

PR #1906, opened 2022-02-20
Changed Rocky Linux 8 instead of CentOS 8

ggtakec:
#1899 

### Details
We merged the changes about GithubActions for `CentOS 8` in #1886 to avoid the error, but we are getting the error again. (it seems in relating java packages)
`CentOS 8` is no longer supported and we have the option of migrating to `CentOS Stream` or migrating to `LockyLinux`.
I think that migrating to `LockyLinux 8` is easy and I post this PR for it.

It is a continuation of PR of #1899. (#1899 was accidentally closed)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
           - debian:bullseye
           - debian:buster
           - debian:stretch
-          - centos:centos8
+          - rockylinux:8
           - centos:centos7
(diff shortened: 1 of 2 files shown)


gaul:
I updated the branch protection rule to rockylinux8.

ggtakec:
> I updated the branch protection rule to rockylinux8.

Thanks!

Issue #1225, opened 2020-01-05
Unify help and man page


gaul:
`src/s3fs_util.cpp` and `doc/man/s3fs.1` become desynchronized when contributors edit one but not the other.  Can we unify these somehow?  [ronn](https://github.com/rtomayko/ronn) generates man pages from Markdown; perhaps we can generate plain text from the Markdown and include it in `src/s3fs_util.cpp`?

gaul:
[help2man](https://linux.die.net/man/1/help2man) can do this although the output does not look good due to the way s3fs formats its `--help`.

PR #1908, opened 2022-02-21
Update notsup_compat_dir in --help

CarstenGrohmann:
Align `s3fs --help` with man page changes introduced with c2a49b7


gaul:
Thank you for your contribution!  I am afraid that these will keep creeping in until we find some single source of truth #1225.

PR #1922, opened 2022-03-11
Enable noobj_cache by default

### LINKED ISSUE #1901: Consider changing `-o noobj_cache` default
Currently the `noobj_cache` is disabled but we should consider enabling it by default, similar to #927.  I am not sure if we need a mechanism to disable it?

gaul:
This should improve performance in many situations.  s3fs already
enables the stat cache by default so memorizing noobj makes this more
consistent.  Fixes #1901.


test/small-integration-test.sh
```diff
@@ -44,7 +44,7 @@ if [ -n "${ALL_TESTS}" ]; then
     FLAGS=(
         "use_cache=${CACHE_DIR} -o ensure_diskfree=${ENSURE_DISKFREE_SIZE} -o fake_diskfree=${FAKE_FREE_DISK_SIZE}"
         enable_content_md5
-        enable_noobj_cache
+        disable_noobj_cache
```
ggtakec, line 47: Isn't it better to test this part as enabled noobj cache?
Until now, it was inspected as enabled, so I would like to recommend this inspection.
I think that the test of disabled noobj cache is better to run in another test or not to run it.

ggtakec (commented):
I agree with enabling the noobj cache.

test/small-integration-test.sh
gaul: I changed the default so that no option specified = `-o enable_noobj_cache`.  Thus we should test the opposite or we can remove this option and speed up CI.

ggtakec: I'm sorry, it was my misunderstanding.

Issue #1068, opened 2019-07-02
Consider using the AWS C++ SDK


gaul:
AWS provides an SDK which could eliminate 1/4 of all s3fs lines of code:

https://aws.amazon.com/sdk-for-cpp/

Using this has drawbacks including forcing a C++11 dependency and likely less SSL library flexibility.

gaul:
s3fs currently has an implicit dependency on RHEL 7.  Here is a chart of enterprise GCC versions:

| Distro | GCC |
| --- | --- |
| RHEL 6 | 4.4.7 |
| RHEL 7 | 4.8.5 |
| RHEL 8 | 8.3.1 |
| Ubuntu 14.04 | 4.8.2 |
| Ubuntu 16.04 | 5.3.1 |
| Ubuntu 18.04 | 7.3.0 |
| Ubuntu 20.04 | 9.3.0 |

It appears that all of C++11 was implemented in GCC in 4.8:

https://gcc.gnu.org/projects/cxx-status.html#cxx11

However, the AWS SDK has a minimum requirement of GCC 4.9: aws/aws-sdk-cpp#1417

gaul:
Using the official SDK might help work around the many IAM issues that users report.

mariadb-JeffBachtel:
Additional motivation: s3fs currently does not implement the `AWS_WEB_IDENTITY_TOKEN_FILE` environment variable auth chain (used with service-linked roles in EKS), which is available in recent AWS SDKs.

mariadb-JeffBachtel:
I don't know if this is helpful, but RHEL7 has `devtoolset-8` which provides GCC 8 for use. I'm sure it'll be annoying for people compiling from scratch to install that as a prerequisite, but it would allow usable RHEL7 builds for folks, if s3fs were using the aws cpp sdk.

gaul:
Ubuntu 16.04 LTS reaches end of standard support this month which removes a potential objection to requiring C++ 11.

gaul:
Using the AWS SDK would allow more options to configure credentials.  References #1122, #1290, and #1724.

gaul:
I would like to prototype using the SDK to see how much work it might be to port s3fs.  One of the biggest problems is how to handle the dependency.  Could we use a git submodule to bring in the external repo?  This would allow us to compile it as if it were our own code and statically link the library.  After doing this we can experiment with a single RPC like `check_bucket`.

ggtakec:
[...]

PR #1927, opened 2022-03-18
Enabled to load shared library for Credential and Token

ggtakec:
#1068 

### Overview
Allows an external shared library to be used instead of s3fs' built-in credentials(and tokens) processing.
Shared libraries are expected to implement credential(and token) processing.

This allows user to make implementations other than the built-in credential processing.
For example, you can replace the built-in processing by providing the credential processing provided by `s3fs-sdk-cpp` as a shared library.
And it is also possible to support servers that perform credential processing that is not compatible with AWS S3.
(For example, IBM's credential processing for storage could be a shared library.)

### Details
Add two options to s3fs, `credlib` and `credlib_ops`. Each definition is shown below:
- `credlib`  
Specify the shared library name(or path) for this option.  
The specified shared library will be loaded at s3fs startup and unloaded at exit.  
- `credlib_ops`  
This option specifies the option string to be passed when loading(initializing) the shared library specified by `credlib`.  
This option parameter string is decided by the shared library.  
Specify an appropriate character string according to the contents.  

The built-in passwd_file, profile, use_session_token, ecs, ibm_iam_auth, ibm_iam_endpoint, imdsv1only and iam_role options are not available when using these options.  

The shared library needs to export the following functions(There are required and optional):  
- Required  
```
const char* VersionS3fsCredential(bool detail);
bool UpdateS3fsCredential(char** ppaccess_key_id, char** ppserect_access_key, char** ppaccess_token, long long* ptoken_expire, char** pperrstr);
```
- Optional  
```
bool InitS3fsCredential(const char* popts, char** pperrstr);
bool FreeS3fsCredential(char** pperrstr);
```
See the `s3fs_extcred.h` file for the contents of each function.  

### Reference
This PR is one of the solutions for #1068.

### Implementation example
I made the credential processing of `aws-sdk-cpp` into this shared library.
https://github.com/ggtakec/s3fs-fuse-awscred-lib

I think anyone can use it and try it out.  
(We are also considering transferring this library to the s3fs-fuse organization if agreed.)


gaul:
If I understand correctly, this uses dynamic linking to include a wrapper for the AWS C++ SDK.  This allows the main s3fs which is GPLv2 or later to retain its effective license, except for users who redistribute both s3fs and the wrapper where the combined s3fs GPLv2 or later and AWS SDK Apache license effectively upgrades to GPLv3 or later (this is my best understanding).

I don't think we should go down this path.  s3fs has an easy deployment mechanism for users since distributions package it.  I don't think that distributions will include this dynamic linking option and few users will understand it.  Instead I think we should either directly depend on the SDK and take the formal license upgrade hit.  Or we could pursue an external mechanism like #1932.

I also opened an issue asking AWS to relicense the SDK under both Apache v2 and MIT which would address one of the issues: https://github.com/aws/aws-sdk-cpp/issues/2018.

ggtakec:
As you pointed out, I think that what we should provide should be included in one package.
And even in this PR, there is no deviation from that rule.
I think that the authentication pattern currently implemented by s3fs-fuse should be maintained and its functionality should be improved.
And I take it for granted to maintain this functionality now.

- NOTE:  
This PR only provides an option to load the DSO of the authentication logic.   
The s3fs package does not need to include DSO.  
If the user requires this DSO, the DSO package or binary will be prepared at the user's responsibility and specified in this option.  
Therefore,  the user or vendor provides DSO package or binary which is a separate from s3fs package.  

So for most users it will not be necessary to specify this DSO by the option.
(Specifying DSO as an option is considered a special case.)

_We should strengthen the current authentication logic of s3fs for AWS S3.(the STS support pointed out in #1932 is another issue...)_
_To do this, we will be able to (1)implement it on the s3fs side, or (2)use the AWS-CPP-SDK._
_I would like to continue to respond in (1) as much as possible._
_However, we will also consider (2), but we will have to consider the issue of licensing. (Mixed licenses are problematic and I want to avoid them)_

The purpose of the external DSO load supported by this PR is to be able to provide other logic that s3fs authorization logic does not support.
I believe this is useful for user environments and vendor hardware dependencies.
Common authentication logic is implemented inside s3fs, but in cases that deviate from it(such as support for new authentication logic), it would be useful to load them with DSO.

I have prepared a DSO implemented in the AWS-CPP-SDK as a sample, but likewise this PR could create the possibility of providing IBM and other vendor-dependent authentication logic in the DSO.
This extension of the authentication process by DSO is only one means.

Normally, it is not assumed that the user will use this option.
if it does, it will be the user's responsibility to look for the DSO and the DSO provider will support the logic.

Finally, I believe that having(loading) this optional authentication method by DSO will increase the availability of s3fs.

ggtakec:

This PR does not change how packages are delivered, nor does it affect licensing.
Also, this is just an option, and will not affect existing logic.
Then, I'm merging this PR to move forward with code updates.

PR #1929, opened 2022-04-03
Upgrade to S3Proxy 2.0.0

gaul:
Release notes:

https://github.com/gaul/s3proxy/releases/tag/s3proxy-2.0.0

This enables Java 17 compatibility.  Also update some Linux CI
runners to JDK 17.


gaul:
I will update the CI runners in a future PR.  This will take a few commits but we should be able to test objects than memory at the end of the process.

ggtakec (approved):

PR #1937, opened 2022-04-21
Removed unnecessary debug options for aws command

ggtakec:
#1933

### Details
With the fix in #1933, I left the aws command with the unnecessary `--debug` option.
It's an option I don't need except when debugging test cases and I forgot to remove it.


gaul:
Sorry I missed this earlier.

Issue #1935, opened 2022-04-19
[ERR] fdcache.cpp:CleanupCacheDirInternal(819)


fikipollo:
#### Version of s3fs being used (s3fs --version)
V1.91

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
2.9.2

#### Kernel information (uname -r)
3.13.0-52-generic

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
Ubuntu 14.04.2 LTS

#### s3fs command line used, if applicable
```
s3fs archive-s3fs-bucket /archive -o endpoint=eu-west-1 -o use_sse=1 -o use_cache=/s3fs-cache -o ensure_diskfree=2000 -o del_cache -o multipart_size=512 -o multipart_copy_size=512 -o logfile=/var/log/s3fs.log -o dbglevel=warn -o parallel_count=6 -o allow_other
```
#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
_if you execute s3fs with dbglevel, curldbg option, you can get detail debug messages_
```
2022-04-19T07:58:07.264Z [ERR] fdcache.cpp:CleanupCacheDirInternal(819): could not get fd_manager_lock when clean up file(/somefile/analysis/results/whatever/sge/a.out)
2022-04-19T07:58:07.264Z [ERR] fdcache.cpp:CleanupCacheDirInternal(819): could not get fd_manager_lock when clean up file(/somefile/analysis/results/whatever/sge/fb.err)
2022-04-19T07:58:07.264Z [ERR] fdcache.cpp:CleanupCacheDirInternal(819): could not get fd_manager_lock when clean up file(/somefile/analysis/results/whatever/sge/cd.err)
2022-04-19T07:58:07.264Z [ERR] fdcache.cpp:CleanupCacheDirInternal(819): could not get fd_manager_lock when clean up file(/somefile/analysis/results/whatever/sge/a.err)
2022-04-19T07:58:07.264Z [ERR] fdcache.cpp:CleanupCacheDirInternal(819): could not get fd_manager_lock when clean up file(/somefile/analysis/results/whatever/sge/fg.err)
2022-04-19T07:58:07.264Z [ERR] fdcache.cpp:CleanupCacheDirInternal(819): could not get fd_manager_lock when clean up file(/somefile/analysis/results/whatever/sge/o.err)
[...]

PR #1946, opened 2022-05-18
Changed the message level(ERR to INFO) of skip cleanup cache

ggtakec:
#1935 

### Details
Change the level of the message to be output in the event of a conflict that occurs during cache cleanup from ERR to INFO.
This part is not an error(processing interruption etc.), and I think that there is no problem at the INFO level(skips the target file due to a conflict).


ggtakec:
I'm getting the following error when building on OpenSUSE. I will reconfirm it later.
```
Repository 'Main Update Repository' is invalid.
```
_This is not a problem with this PR. (It seems to be a problem with the package repository)_

ggtakec:
@gaul  Sorry before the review, I'll merge this.
If you find a problem, please let me know.

Issue #1890, opened 2022-02-15
s3fs 1.90 terminates with segfault 


nkkashyap:
### Additional Information
_The following information is very important in order to help us to help you. Omission of the following details may delay your support request or receive no attention at all._
_Keep in mind that the commands we provide to retrieve information are oriented to GNU/Linux Distributions, so you could need to use others if you use s3fs on macOS or BSD_

#### Version of s3fs being used (s3fs --version)
_Version: V1.90 (commit:cd466eb)_
```
# s3fs --version
Amazon Simple Storage Service File System V1.90 (commit:cd466eb) with OpenSSL
Copyright (C) 2010 Randy Rizun <rrizun@gmail.com>
License GPL2: GNU GPL version 2 <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
```

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
_Fuse Version: 2.9.7-1ubuntu1_
```
# dpkg -s fuse
Package: fuse
Status: install ok installed
Priority: optional
Section: utils
Installed-Size: 104
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Architecture: amd64
Version: 2.9.7-1ubuntu1
```
#### Kernel information (uname -r)
_Kernel: 4.15.0-163-generic_
```
# uname -r
4.15.0-163-generic
```

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
_Distro: Ubuntu 18.04.6 LTS_
```
# cat /etc/os-release 
NAME="Ubuntu"
VERSION="18.04.6 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.6 LTS"
VERSION_ID="18.04"
```
#### s3fs command line used, if applicable
```
[...]

PR #1948, opened 2022-05-23
fix for segfault while reseting handle for retrying

nkkashyap:
Signed-off-by: Neeraj Kumar Kashyap <nkashyap@in.ibm.com>

s3fs is terminating with Segmentation fault
The issue is occurring with s3fs release v1.91
 
### Details
In one situation, with `ibm_iam_auth` option `s3fs` tries to re-connect `private.iam.cloud.ibm.com` and getting `NSS error -5938 (PR_END_OF_FILE_ERROR)`. To handle this, `s3fs` resets the handle by calling `RemakeHandle` which in-turn calls `curl_slist_remove` to remove the key `Authorization` and finally hitting with `SIGSEGV`.

```
Found bundle for host private.iam.cloud.ibm.com: 0x7f0f6e1cd410
Connection 22 seems to be dead!
Closing connection 22 
About to connect() to private.iam.cloud.ibm.com port 443 (#23)
    Trying 166.9.251.2... 
    
Connected to private.iam.cloud.ibm.com (166.9.251.2) port 443 (#23)
NSS error -5938 (PR_END_OF_FILE_ERROR)
Encountered end of file
Closing connection 23 
curl.cpp:RequestPerform(2450): ### CURLE_SSL_CONNECT_ERROR  
kernel: s3fs[54514]: segfault at 8 ip 0000000000457e61 sp 00007f0f827f1e50 error 4 in s3fs[400000+bb000]
```

The code change in this PR provides the fix for the issue by terminating the `for loop` in `curl_slist_remove` once the key is removed from `requestHeaders`.

### DIFF
--- a/src/curl_util.cpp
+++ b/src/curl_util.cpp
@@ -122,6 +122,7 @@ struct curl_slist* curl_slist_remove(struct curl_slist* list, const char* key)
             struct curl_slist *tmp = *p;
             *p = (*p)->next;
             free(tmp);
+            break;
         }
     }
 


erapetti:
Works for me.

With s3fs-fuse version 1.91 and "-o ibm_iam_auth" it exits with SIGSEGV whenever there's a network error.
I tried this patched version on centos 7 and it keeps retrying, as expected, until the network is restored.

src/curl_util.cpp
```diff
@@ -122,6 +122,7 @@ struct curl_slist* curl_slist_remove(struct curl_slist* list, const char* key)
             struct curl_slist *tmp = *p;
             *p = (*p)->next;
```
gaul, line 123: Does this segfault because the pointer is advanced twice, one here and once on line 112?  So this logic is wrong?

Also is it possible to handle the situation with multiple list nodes with the name key?

Could you also add a unit test?

gaul (changes requested):

src/curl_util.cpp
ggtakec: @gaul I commented, but this seems to be due to a simple bug.
First of all, I think it's a good idea to have my suggested code tried.

erapetti:
@ggtakec  to reproduce the issue you need to get authorization using IBM API key, and then some error that trigger RemakeHandle.
For now you can put my API key in a "passwd-s3fs" file:

```
# cat passwd-s3fs
glpi-backup:ServiceId-7a673941-397d-40b6-a328-32fb13a40f81:UtRJTF_b1QWuZARh-U_eG06X4bPChDIPgZKH_3sVnF2v
```

Then this command will end with SIGSEGV:

```
# s3fs -d -d -f -s glpi-backup /mnt/ -o https://www.google.com -o passwd_file=passwd-s3fs -o ibm_iam_auth
```

Regarding your second code proposal (30 minutes ago), it fails with segfault:

```
2022-05-24T12:57:00.801Z [ERR] curl.cpp:CurlProgress(356): timeout now: 1653397020, curl_times[curl]: 1653396898, readwrite_timeout: 120
2022-05-24T12:57:00.801Z [ERR] curl.cpp:RequestPerform(2434): ### CURLE_ABORTED_BY_CALLBACK
2022-05-24T12:57:04.801Z [INF] curl.cpp:RequestPerform(2523): ### retrying...
2022-05-24T12:57:04.801Z [INF]       curl.cpp:RemakeHandle(1969): Retry request. [type=5][url=https://glpi-backup.s3.direct.us-south.cloud-object-storage.appdomain.cloud/][path=/]

Thread 1 "s3fstest" received signal SIGSEGV, Segmentation fault.
curl_slist_remove (list=0x0, key=key@entry=0x46a178 "Authorization") at curl_util.cpp:113
113	    for(;pcur; pprev = pcur, pcur = pcur->next){
#0  curl_slist_remove (list=0x0, key=key@entry=0x46a178 "Authorization") at curl_util.cpp:113
#1  0x0000000000427b6a in S3fsCurl::RemakeHandle (this=this@entry=0x7fffffffe160) at curl.cpp:1986
#2  0x00000000004286dc in S3fsCurl::RequestPerform (this=this@entry=0x7fffffffe160, 
    dontAddAuthHeaders=dontAddAuthHeaders@entry=false) at curl.cpp:2525
#3  0x000000000042b3cd in S3fsCurl::CheckBucket (this=this@entry=0x7fffffffe160) at curl.cpp:3513
#4  0x000000000040f5b9 in s3fs_check_service () at s3fs.cpp:3523
#5  0x000000000040fd15 in s3fs_init (conn=0x6b2f74) at s3fs.cpp:3410
#6  0x00007ffff7ba36af in fuse_fs_init () from /lib64/libfuse.so.2
#7  0x00007ffff7ba7969 in do_init () from /lib64/libfuse.so.2
#8  0x00007ffff7ba8818 in fuse_ll_process_buf () from /lib64/libfuse.so.2
#9  0x00007ffff7ba5013 in fuse_session_loop () from /lib64/libfuse.so.2
#10 0x00007ffff7b9d0b6 in fuse_loop () from /lib64/libfuse.so.2
#11 0x00007ffff7bada87 in fuse_main_common () from /lib64/libfuse.so.2
#12 0x0000000000406bfa in main (argc=13, argv=<optimized out>) at s3fs.cpp:4711
```

src/curl_util.cpp
nkkashyap: @gaul @ggtakec 
Thanks for your comments.
This is a bug in the code, after removing the key from the `requestHeaders` the pointer `*p` is getting advanced twice. 
Currently my code change is breaking the loop assuming there is always single instance of a key in the nodes of `requestHeaders` list.

erapetti:
@ggtakec Your code advances pcur inside "if(0 == result)" and then again with the for's next step.

If the searched key is in the first element and there are no more elements, then you got segfault in the for.
If there are more elements then you skip the second.

erapetti:
Just for the record, this PR is related to issue #1890, which happens at least in v1.90 and v.191 when retrying a connection with "-o ibm_iam_auth"

erapetti:
@nkkashyap what about if you change curl_slist_remove as follows:

```
struct curl_slist* curl_slist_remove(struct curl_slist* list, const char* key)
{
    if(!key){
        return list;
    }

    std::string strkey = trim(std::string(key));
    struct curl_slist **p = &list;
    while(*p){
        std::string strcur = (*p)->data;
        size_t pos;
        if(std::string::npos != (pos = strcur.find(':', 0))){
            strcur.erase(pos);
        }

        int result = strcasecmp(strkey.c_str(), strcur.c_str());
        if(0 == result){
            free((*p)->data);
            struct curl_slist *tmp = *p;
            *p = (*p)->next;
            free(tmp);
        } else {
            p = &(*p)->next;
        }
    }

    return list;
}
```

It works for me and I think it also address @gaul concerns:
- the pointer is advanced only once per iteration
- multiple list nodes with the name key are removed in a single call to curl_slist_remove

Also, the empty list is handled by the condition in the while.

ggtakec:
@erapetti Thank you for taking your time.
> Your code advances pcur inside "if(0 == result)" and then again with the for's next step.

You're right, my code had still bugs.

In your suggested code, if the beginning of `list` is freed, this function will return `list` as is, so the problem remains.

I think it's better to sort these out a bit (including the bug you found in my code) and rewrite the function as follows:
```
struct curl_slist* curl_slist_remove(struct curl_slist* list, const char* key)
{
    if(!list || !key){
        return list;
    }

    std::string strkey = trim(std::string(key));
    struct curl_slist *pprev, *pcur, *pnext;
    for(pprev = NULL, pcur = list, pnext = NULL; pcur; pcur = pnext){
        pnext = pcur->next;

        std::string strcur = pcur->data;
        size_t pos;
        if(std::string::npos != (pos = strcur.find(':', 0))){
            strcur.erase(pos);
        }

        int result = strcasecmp(strkey.c_str(), strcur.c_str());
        if(0 == result){
            if(pcur == list){
                list = pnext;
            }
            if(pprev){
                pprev->next = pnext;
            }
            free(pcur->data);
            free(pcur);
        }else{
            pprev = pcur;
        }
    }
    return list;
}
```

Sorry to trouble you, but can you give it a try?

erapetti:
@ggtakec you code works for me

nkkashyap:
@erapetti Thanks
@ggtakec It is fine for me, you can deliver the code change and I will close this pr :). Thanks

ggtakec:
@nkkashyap @erapetti Thanks for your kindness
I made #1950 based on this PR.

PR #1951, opened 2022-05-26
Fix double-iteration in curl_slist_remove

gaul:
Also backfill tests.  Fixes #1948.

### DIFF
--- a/src/curl_util.cpp
+++ b/src/curl_util.cpp
@@ -109,7 +109,7 @@ struct curl_slist* curl_slist_remove(struct curl_slist* list, const char* key)
 
     std::string strkey = trim(std::string(key));
     struct curl_slist **p = &list;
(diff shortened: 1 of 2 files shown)


gaul:
@nkkashyap can you test this against IBM to see if it resolves your symptoms?

@ggtakec I think this is a simpler approach that includes more test coverage.

ggtakec (approved):

ggtakec:
@gaul Thank you for the test code. (I forgot test_curl_util.cpp)
I don't think there is any problem with this test.
Also, this PR is simpler than the fix for #1950, so once it's merged, #1950 will close.

nkkashyap (approved):
PR looks good

ggtakec:
@nkkashyap Thanks for your help.

Issue #1947, opened 2022-05-22
Permission denied when run sed with the --in-place option


ggtakec:
#### Version of s3fs being used (s3fs --version)
Amazon Simple Storage Service File System V1.91 (commit:a30bede) with OpenSSL

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
2.9.2

#### Kernel information (uname -r)
5.10.96-90.460.amzn2.x86_64

#### s3fs command line used, if applicable
```
s3fs <bucket> /mnt/s3 -o allow_other,nonempty,enable_noobj_cache,enable_content_md5,use_cache=/tmp,url=https://s3.amazonaws.com,multireq_max=50,parallel_count=10,dbglevel=dbg,curldbg
```
### Details about issue
Replacing a file in a mount with the `--in-place` option of the `sed` command results in `Permission denied`.

I prepare a 1002 byte file and execute the `sed` command as shown below, an error will occur.
```
$ stat testfile
    File: ‘testfile’
    Size: 1002            Blocks: 2          IO Block: 4096   regular file
  Device: 2bh/43d Inode: 25          Links: 1
  Access: (0644/-rw-r--r--)  Uid: ( 1001/nakatani)   Gid: (  100/   users)
  Access: 2022-05-22 09:04:00.000000000 +0000
  Modify: 2022-05-22 09:04:00.000000000 +0000
  Change: 2022-05-22 09:04:00.000000000 +0000
   Birth: -

$ sed -i -e 's/target_string/TARGET_STRING/g' testfile
  sed: couldn't close ./sedFgcMLn: Permission denied
```

This issue was discovered during the #1939 investigation. (It is still unknown if it is related to that Issue)

PR #1954, opened 2022-05-29
Fixed a bug could not change the mode while the file was opened

ggtakec:
#1947 

### Details
When using the `--in-place(-c)` option of the sed command, sed seems to behave as follows:
- Create a new temporary file(create the file mode as a `0000`)
- Read the original file and write the replacement result to the temporary file
- Changed temporary file mode to `0644`
- Close file, and replace file

In s3fs, the stat information of a file is internally cached until flush is executed, during the file is opened.
In sed `--in-place` mode, a `Permission denied(EACCESS)` error occurred because s3fs forgot to update the open file mode to the stat cache.
So this PR fixed this bug.

#### NOTE
This PR added a test case.
Added test case was implemented with the sed command.
I made a shell script and a simple C test code and tried bug reproduction, but I couldn't reproduce the same error.
So the test is implemented with the sed command.

And in addition to the file mode, the file owner and group cache may also change, but I couldn't find a pattern that would cause an error while doing these things.
Therefore, this fix is limited to the file mode only.

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -412,6 +412,9 @@ bool StatCache::AddStat(const std::string& key, headers_t& meta, bool forcedir, 
 // Updates only meta data if cached data exists.
 // And when these are updated, it also updates the cache time.
 //
+// Since the file mode may change while the file is open, it is
+// updated as well.
+//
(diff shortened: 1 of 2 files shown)


ggtakec:
@gaul Thanks!

Issue #1934, opened 2022-04-18
s3fs1.91 : can't touch file using NFS


shiwu515:
#### GNU/Linux Distribution, if applicable
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"

#### s3fs command line used, if applicable
s3fs test /mnt/s3fs -o passwd_file=/etc/.passwd-s3fs -o url=http://**.**.**.** -o use_path_request_style -o noxmlns -o dbglevel=debug -f -o curldbg | tee -a /root/log.txt

#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
_if you execute s3fs with dbglevel, curldbg option, you can get detail debug messages_
2022-04-18T01:41:39.697Z [INF] s3fs.cpp:s3fs_create(905): [path=/7.c][mode=100644][flags=0xc1]
2022-04-18T01:41:39.697Z [DBG] s3fs.cpp:check_parent_object_access(627): [path=/7.c]
2022-04-18T01:41:39.697Z [DBG] s3fs.cpp:check_object_access(519): [path=/]
2022-04-18T01:41:39.697Z [DBG] s3fs.cpp:check_object_access(524): [pid=32142,uid=0,gid=0]
2022-04-18T01:41:39.697Z [DBG] s3fs.cpp:get_object_attribute(350): [path=/]
2022-04-18T01:41:39.697Z [DBG] s3fs.cpp:check_object_access(519): [path=/7.c]
2022-04-18T01:41:39.697Z [DBG] s3fs.cpp:check_object_access(524): [pid=32142,uid=0,gid=0]
2022-04-18T01:41:39.697Z [DBG] s3fs.cpp:get_object_attribute(350): [path=/7.c]
[...]

PR #1957, opened 2022-06-10
Fixed a bug that regular files could not be created by mknod

ggtakec:
#1934

### Details

The touch command fails when the directory on which s3fs is mounted is mounted with NFS, by #1934.
As a result of investigating this, it turned out to be a bug.

#### Cause

FUSE calls the `create`(`s3fs_sreate`) hook when creating a regular file with mknod.
After that, `flush`(`s3fs_flush`) is not called because mknod is the trigger, and `release`(`s3fs_release`) closes the file handle.
After this, file attribute changes such as `utimens` are called.

Also, when creating a file such as a `touch` command, NFS does not call `open`/`close`, but calls `mknod`.

Due to the above operation of FUSE and NFS, to create a new file with zero file size by the `touch` command, `create`(`s3fs_create`) is called, and `release`(`s3fs_release`) is called without `flush`(`s3fs_flush`).

s3fs delays the creation of new zero file size files until `flush`(`s3fs_flush`). (you know #1640, #1655, #1780, etc.)
This does not take into account the call flow by mknod, so the pending file creation(and meta update) is not done.
After that, calling `utimens` will result in an error because the file has not been created.
Therefore, the touch command failed via NFS.

#### Fix
When `release`(`s3fs_release`) is called, if it is in the pending state(new file creation, meta information update), it has been modified to update them.
This allows you to `create`(`touch` command) regular files with `mknod`(from NFS).

I also added a test for `mknod`.
FUSE is called by `create`(`s3fs_create`) in the case of regular file creation, but by `mknod`(`s3fs_mknod`) in other cases.
Including these, I am making it possible to inspect with this test.
All `mknod` tests are bypassed on macos (because root privileges are required).
It will also be bypassed if you create a BLOCK device as well.

#### Remaining issues
It looks like mknod's non-regular file creation is supported, but it doesn't work correctly.
I think that `BLOCK`, `FIFO` and `SOCK` will not work.
As for `CHARACTER`, it may work, but it has not been implemented as a block device.
I will post new issue about it.

### DIFF
--- a/src/fdcache_entity.h
+++ b/src/fdcache_entity.h
@@ -105,7 +105,7 @@ class FdEntity
         int GetPhysicalFd() const { return physical_fd; }
         bool IsModified() const;
         bool MergeOrgMeta(headers_t& updatemeta);
-        int UploadPending(int fd = -1);
+        int UploadPending(int fd);
 
         bool GetStats(struct stat& st, bool lock_already_held = false);
         int SetCtime(struct timespec time, bool lock_already_held = false);
--- a/src/fdcache_entity.h
+++ b/src/fdcache_entity.h
@@ -32,6 +32,17 @@
 class FdEntity
 {
     private:
+        // [NOTE]
+        // Distinguish between meta pending and new file creation pending,
+        // because the processing(request) at these updates is different.
+        // Therefore, the pending state is expressed by this enum type.
+        //
+        enum pending_status_t {
+            NO_UPDATE_PENDING = 0,
+            UPDATE_META_PENDING,        // pending meta header
+            CREATE_FILE_PENDING         // pending file creation and meta header
+        };
+
         static bool     mixmultipart;   // whether multipart uploading can use copy api.
 
         pthread_mutex_t fdent_lock;
@@ -49,7 +60,7 @@ class FdEntity
         std::string     cachepath;      // local cache file path
                                         // (if this is empty, does not load/save pagelist.)
         std::string     mirrorpath;     // mirror file path to local cache file path
-        bool            is_meta_pending;
+        pending_status_t pending_status;// status for new file creation and meta update
         struct timespec holding_mtime;  // if mtime is updated while the file is open, it is set time_t value
 
     private:
@@ -73,7 +84,6 @@ class FdEntity
         ssize_t WriteNoMultipart(PseudoFdInfo* pseudo_obj, const char* bytes, off_t start, size_t size);
         ssize_t WriteMultipart(PseudoFdInfo* pseudo_obj, const char* bytes, off_t start, size_t size);
         ssize_t WriteMixMultipart(PseudoFdInfo* pseudo_obj, const char* bytes, off_t start, size_t size);
-        int UploadPendingMeta();
 
     public:
         static bool GetNoMixMultipart() { return mixmultipart; }
@@ -95,6 +105,7 @@ class FdEntity
         int GetPhysicalFd() const { return physical_fd; }
         bool IsModified() const;
         bool MergeOrgMeta(headers_t& updatemeta);
+        int UploadPending(int fd = -1);
 
         bool GetStats(struct stat& st, bool lock_already_held = false);
         int SetCtime(struct timespec time, bool lock_already_held = false);
--- a/test/mknod_test.c
+++ b/test/mknod_test.c
@@ -82,7 +82,7 @@ bool TestMknod(const char* basepath, mode_t mode)
             break;
         case S_IFBLK:
             str_mode = str_mode_blk;
-            dev      = makedev((long long)(259), 0);     // temporary value
+            dev      = makedev((unsigned int)(259), 0);     // temporary value
             sprintf(filepath, "%s.%s", basepath, str_ext_blk);
             break;
         case S_IFIFO:
@@ -93,7 +93,8 @@ bool TestMknod(const char* basepath, mode_t mode)
         case S_IFSOCK:
             str_mode = str_mode_sock;
             dev      = 0;
-            sprintf(filepath, "%s.%s", basepath, str_ext_sock);
+            snprintf(filepath, S3FS_TEST_PATH_MAX, "%s.%s", basepath, str_ext_sock);
+            filepath[S3FS_TEST_PATH_MAX - 1] = '\0';    // for safety
             break;
         default:
             fprintf(stderr, "[ERROR] Called function with wrong mode argument.\n");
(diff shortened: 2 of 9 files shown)


giulianoc:
Hi, 

just for your information,
I had the same scenario:
- I have mounted through s3fs an s3 bucket
- this folder is exported via nfs

and the same issue:
- if I touch a file into the nfs mounted folder I get an error and the file does not appear into the folder mounted via s3fs
- If I edit a file using vi, the file does not appear into the folder mounted via s3fs

I just downloaded this fix/commit and now it works fine.
May I know if and when this fix will be released in order to be installed using the (apt install s3fs) command?

Best regards
Giuliano

giulianoc:
Hi,
I just saw an issue with this fix/commit:

scenario:
- I have mounted through s3fs an s3 bucket
- this folder is exported via nfs

steps to reproduce the issue:
1. a file is created into the folder nfs mounted (test.txt)
2. same file is removed from the s3fs mounted folder (rm test.txt). Correctly, the file is removed from the s3fs mounted folder.
3. listing the nfs folder directory, I see:

root@ip-172-31-41-170:/home/mms# ll /mmsStorageTest/
ls: cannot access '/mmsStorageTest/test.txt': No such file or directory
total 8
drwxrwx---  1 mms  root    0 Jan  1  1970 ./
drwxr-xr-x 28 root root 4096 Jun 11 09:42 ../
-?????????  ? ?    ?       ?            ? test.txt

Best regards
Giuliano

ggtakec:
@giulianoc Thanks for checking this PR code.
I also checked `vi` command via NFS mount.
It seems that `vi` usually creates a `.swp` file and then recreates the file when saving.
Therefore via NFS, `mknod` was called on both of them, causing a problem.
With this PR fix, I've confirmed that this `mknod` process works correctly and solves your problem.

After merging this PR into master, I think it will be included in the next release.
I try to release s3fs about every six months(although not a rule), so I consult with @gaul.
Please wait a little more.

giulianoc:
hi @ggtakec,
I find another issue using yout fix/commit:

1. I have an s3 bucket mounted on a server using s3fs
2. same folder is exported through nfs
3. I have 2 nfs client server mounting both the folder through nfs, let's call them nfsClient_1 and nfsClient_2
4. I create a file into the folder on nfsClient_1

I expect to see the same file on nfsClient_2 but it is not
Let me know if you need more information
Best regards

ggtakec:
@giulianoc 
I was able to reproduce it without having to start two NFS clients.
However, that seems to be a slightly different cause than this Issue/PR.
I created it as another Issue #1961, so I will consider it there.

test/mknod_test.c
```diff
@@ -0,0 +1,177 @@
...
+//---------------------------------------------------------
+int main(int argc, char *argv[])
+{
+    // Parse parameters
+    if(2 != argc){
+        fprintf(stderr, "[ERROR] No paraemter is specified.\n");
+        fprintf(stderr, "%s\n", usage_string);
+        exit(EXIT_FAILURE);
+    }
+    if(0 == strcasecmp("-h", argv[1]) || 0 == strcasecmp("--help", argv[1])){
```
gaul, line 144: Should this be the simpler `strcmp`?

test/mknod_test.c
```diff
@@ -0,0 +1,177 @@
...
+    // Test
+    //
+    // [NOTE]
+    // Privilege is required to execute S_IFBLK.
+    //
+    if(!TestMknod(argv[1], S_IFREG)  ||
+       !TestMknod(argv[1], S_IFCHR)  ||
+       !TestMknod(argv[1], S_IFIFO)  ||
+       !TestMknod(argv[1], S_IFSOCK) ||
+       (0 == geteuid() && !TestMknod(argv[1], S_IFBLK)))
```
gaul, line 162: Should this warn, "skipping mknod due to missing root privileges"?

test/mknod_test.c
```diff
@@ -0,0 +1,177 @@
...
+            break;
+        case S_IFIFO:
+            str_mode = str_mode_fifo;
+            dev      = 0;
+            sprintf(filepath, "%s.%s", basepath, str_ext_fifo);
+            break;
+        case S_IFSOCK:
+            str_mode = str_mode_sock;
+            dev      = 0;
+            sprintf(filepath, "%s.%s", basepath, str_ext_sock);
```
gaul, line 96: `snprintf` or use `std::string`.

src/fdcache_entity.h
```diff
@@ -95,6 +105,7 @@ class FdEntity
         int GetPhysicalFd() const { return physical_fd; }
         bool IsModified() const;
         bool MergeOrgMeta(headers_t& updatemeta);
+        int UploadPending(int fd = -1);
```
gaul, line 108: Better to remove the default parameter and just let the caller specify -1 when needed?

gaul (changes requested):

test/mknod_test.c
```diff
@@ -0,0 +1,177 @@
...
+            sprintf(filepath, "%s.%s", basepath, str_ext_reg);
+            break;
+        case S_IFCHR:
+            str_mode = str_mode_chr;
+            dev      = makedev(0, 0);
+            sprintf(filepath, "%s.%s", basepath, str_ext_chr);
+            break;
+        case S_IFBLK:
+            str_mode = str_mode_blk;
+            dev      = makedev((long long)(259), 0);     // temporary value
```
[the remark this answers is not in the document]
ggtakec: If I didn't cast it, I got a build error on macos, so I used it as `long long`, but it was `unsigned int`.

test/mknod_test.c
```diff
@@ -0,0 +1,177 @@
...
+}
+
+//---------------------------------------------------------
+// Main
+//---------------------------------------------------------
+int main(int argc, char *argv[])
+{
+    // Parse parameters
+    if(2 != argc){
+        fprintf(stderr, "[ERROR] No paraemter is specified.\n");
```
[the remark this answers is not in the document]
ggtakec: Sorry, it will be fixed soon.

test/mknod_test.c
ggtakec: Either is fine, but only lowercase letters are fine, so I'll modify it to strcmp.

test/mknod_test.c
ggtakec: I agree. Let's output the warning in an easy-to-understand manner.

test/mknod_test.c
ggtakec: This is a C source, so change it to snprintf.

src/fdcache_entity.h
ggtakec: Since `fd` was omitted only when calling from inside the `FdEntity` class, there is no problem by specifying `-1`.
Delete the default value.

ggtakec:
@gaul please re-review this.

PR #1960, opened 2022-06-12
Avoid extended initializer lists in C++03

gaul:
C++11 introduces these.


ggtakec (approved):
I'm sorry that I forgot that there was a warning in the macos build log before.

PR #1984, opened 2022-07-16
Re-re-re-fix propagating the return code

gaul:
This shell incantation is necessary to have errexit take effect for
test bodies.  Previous only the _last_ command's exit code was
propagated to the test runner, masking real failures.  Reverts
ea3c21f270db3c3298ba9b7d04e61146c712e9bb.


gaul:
This was hiding failures I introduced in #1970. :3rd_place_medal:

gaul:
CI shows a real failure that I will investigate.

ggtakec (approved):

PR #1988, opened 2022-07-16
Fixed a bug in the test_external_creation test

ggtakec:
#1984

### Details
Fixed a test error that occurred in #1984.
The error with test_external_creation is because the file stat(the file is not existed)  is not cached when noobj_cache is enabled.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -397,11 +397,18 @@ function test_external_modification {
 function test_external_creation {
     describe "Test external creation of an object ..."
     local OBJECT_NAME; OBJECT_NAME=$(basename "${PWD}")/"${TEST_TEXT_FILE}"
+
+    # [NOTE]
+    # If noobj_cache is enabled, register that cache here.
+    #
+    [ ! -e "${TEST_TEXT_FILE}" ]
+
+    # [NOTE]
+    # If noobj_cache is enabled, we cannot be sure that it is registered in that cache.
+    # That's because an error will occur if the upload by aws cli takes more than 1 second.
+    #
     echo "data" | aws_cli s3 cp - "s3://${TEST_BUCKET_1}/${OBJECT_NAME}"
-    # shellcheck disable=SC2009
-    if ! ps u -p "${S3FS_PID}" | grep -q disable_noobj_cache; then
-        [ ! -e "${TEST_TEXT_FILE}" ]
-    fi
+
     sleep 1
     [ -e "${TEST_TEXT_FILE}" ]
     rm -f "${TEST_TEXT_FILE}"


test/integration-test-main.sh
```diff
@@ -397,6 +397,15 @@ function test_external_modification {
 function test_external_creation {
     describe "Test external creation of an object ..."
     local OBJECT_NAME; OBJECT_NAME=$(basename "${PWD}")/"${TEST_TEXT_FILE}"
+
+    # [NOTE]
+    # This test requires that "TEST_TEXT_FILE" does not exist.
+    # (If it exists, it is forgetting to remove it somewhere.)
+    # Also, if noobj_cache is enabled, do the following test first to
+    # ensure that noobj_cache caches it.
```
gaul, line 405: `run_suite` sets a new name for each test so this should always not exist:

```
TEST_TEXT_FILE="test-s3fs-${RANDOM}.txt"
```

test/integration-test-main.sh
```diff
@@ -397,6 +397,15 @@ function test_external_modification {
...
+    #
+    [ ! -e "${TEST_TEXT_FILE}" ]
+
     echo "data" | aws_cli s3 cp - "s3://${TEST_BUCKET_1}/${OBJECT_NAME}"
     # shellcheck disable=SC2009
     if ! ps u -p "${S3FS_PID}" | grep -q disable_noobj_cache; then
```
gaul, line 411: The following line is the one that fails.  I think that this test is the same as before but since we now use a different directory for every function that it has a different behavior than before?

gaul (changes requested):

test/integration-test-main.sh
gaul: Sorry -- tests use different _files_ for every run but do not `cd` into a new directory.

test/integration-test-main.sh
ggtakec: Some of this comment was unnecessary. It has been deleted.

test/integration-test-main.sh
ggtakec: I'm sorry, I'm a little unsure what's wrong.
The meaning of testing TEST_TEXT_FILE at the beginning of this test is to cache the non-existence of the file.
(If disable_noobj_cache is not specified, that cache will be needed on the next line)

ggtakec: There are cases where the cache expiration date has passed, and it seems that it may fail.
I try to fix it.

ggtakec: `echo "data" | aws_cli s3 cp - "s3://${TEST_BUCKET_1}/${OBJECT_NAME}"`
This process seems to take more than 700ms.
Depending on the situation, it seems to take more than 1 second.
In that case, the test here will fail.
Therefore, when testing noobj_cache, the results will be unstable.

ggtakec: When noobj_cache was enabled, this test was checking that the file was cached for non-existence.
However, it depends on the processing time of aws cli as mentioned above.
The processing time may take more than 1 second depending on the situation, causing an error.

I have concluded that this cannot be avoided.
As a result, it has been fixed so that it does not check that the file is cached for non-existence.

ggtakec:
@gaul Please review again.
Changed not to run the test of whether noobj_cache is valid.
(Check the comments from the previous review)

test/integration-test-main.sh
```diff
@@ -397,11 +397,18 @@ function test_external_modification {
...
+    # [NOTE]
+    # If noobj_cache is enabled, we cannot be sure that it is registered in that cache.
+    # That's because an error will occur if the upload by aws cli takes more than 1 second.
+    #
     echo "data" | aws_cli s3 cp - "s3://${TEST_BUCKET_1}/${OBJECT_NAME}"
-    # shellcheck disable=SC2009
-    if ! ps u -p "${S3FS_PID}" | grep -q disable_noobj_cache; then
-        [ ! -e "${TEST_TEXT_FILE}" ]
-    fi
+
```
gaul, line 411: This also works:

```
     echo "data" | aws_cli s3 cp - "s3://${TEST_BUCKET_1}/${OBJECT_NAME}"
-    # shellcheck disable=SC2009
-    if ! ps u -p "${S3FS_PID}" | grep -q disable_noobj_cache; then
-        [ ! -e "${TEST_TEXT_FILE}" ]
-    fi
-    sleep 1
     [ -e "${TEST_TEXT_FILE}" ]
     rm -f "${TEST_TEXT_FILE}"
```

I'm not sure what this test should be checking?

ggtakec: I have detected that the aws cli call here takes somtimes more than a second.
It occurred several times on any OS(Linux system) in Github Actions.
After trying many times on my HOST, this call actually takes 600-1000ms(over).

Issue #1991, opened 2022-07-20
Deadlock with `-o streamupload`


gaul:
I notice that s3fs deadlocks when copying a ~300 MB file to AWS S3 when using the new `-o streamupload`.  I will try to collect some ThreadSanitizer debugging info.  References #1802.

ggtakec:
I still can't reproduce this deadlock problem.
Could you give me more information for reproducing?
(I think it won't take long to fix if it can be reproduced)

gaul:
I'm not sure I understand this race:

```
WARNING: ThreadSanitizer: data race (pid=1001103)
  Write of size 8 at 0x7b0800047000 by thread T50 (mutexes: write M0):
    #0 free <null> (s3fs+0x4f1c17) (BuildId: 9d87e7c5947222380e818fada77a1921171cc970)
    #1 OPENSSL_sk_free <null> (libcrypto.so.3+0x1fe4f9) (BuildId: 1cc34c0eccf5c38e231a0926e6acc94da2d9bc94)
    #2 S3fsCurl::HeadRequest(char const*, std::map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, header_nocase_cmp, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > >&) /home/gaul/work/s3fs-fuse/src/curl.cpp:3172:87 (s3fs+0x58f94e) (BuildId: 9d87e7c5947222380e818fada77a1921171cc970)
    #3 get_object_attribute(char const*, stat*, std::map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, header_nocase_cmp, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > >*, bool, bool*, bool) /home/gaul/work/s3fs-fuse/src/s3fs.cpp:385:28 (s3fs+0x54ce97) (BuildId: 9d87e7c5947222380e818fada77a1921171cc970)
[...]

PR #1992, opened 2022-07-20
Protect pending_status in UploadPending

gaul:
This requires avoiding double-locking in `RowFlush`.  References #1991.


gaul:
This passes locally.  I will look more closely this weekend.

ggtakec (approved):

PR #1993, opened 2022-07-22
Protect instruct_count in ParallelMultipartUpload

gaul:
References #1991.

### DIFF
--- a/src/fdcache_fdinfo.cpp
+++ b/src/fdcache_fdinfo.cpp
@@ -433,6 +433,8 @@ bool PseudoFdInfo::ParallelMultipartUpload(const char* path, const mp_part_list_
             delete thargs;
             return false;
         }
+
+        AutoLock auto_lock(&upload_list_lock);
         ++instruct_count;
     }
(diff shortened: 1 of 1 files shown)


src/fdcache_fdinfo.cpp
```diff
@@ -433,6 +433,8 @@ bool PseudoFdInfo::ParallelMultipartUpload(const char* path, const mp_part_list_
             delete thargs;
             return false;
         }
+
+        AutoLock auto_lock(&upload_list_lock);
```
gaul, line 437: Why do we even need a thread pool?  Creating a thread is much cheaper than the network RPC.

gaul:
This is incorrect.

src/fdcache_fdinfo.cpp
ggtakec: It is lighter than network RPC as you say, but thread pools are different.
Since it only issues instructions(semaphore) to threads that have been started(waiting) in advance, there is no startup overhead.

Issue #1987, opened 2022-07-16
Sometimes the test_truncate_cache fails on macos


ggtakec:
### Version of s3fs being used (s3fs --version)
v1.91

### Details about issue
On macos, sometimes test_truncate_cache fails.
I'm currently investigating this issue, but I don't yet know the exact conditions.

Normally, when deleting a directory, FUSE (osxfuse) calls s3fs in the following order.
- Delete each file in the directory
- Delete directory
When s3fs receives a directory deletion, it will fail the operation if the directory is not empty.

If it fails on macos, directory deletion is called during file deletions.
At this time, some file deletion are called even after directory deletion is called.
(I'm not sure, but it seemed to be called in parallel)

I'm investigating the conditions under which this phenomenon occurs.

PR #1994, opened 2022-07-22
Fixed a test_truncate_cache failure on macos

ggtakec:
#1987

### Details
About #1987, I found out where `test_truncate_cache` sometimes fails on macos.

It happens with `rm -rf $ (seq 2)` in `test_truncate_cache` test.
When deleting two directories, `s3fs_rmdir` fails because sometimes files remain in the directories.

When FUSE receives a system call to delete a directory, it lists the files in the directory.
It then deletes each of the listed files in turn(`s3fs_unlink`), deletes them all, and then calls `s3fs_rmdir`.
For macos(which may also happen on Linux?), any files were missing from the list of files in this directory at deleting each one.

The location of the problem has been identified and it has been fixed (as much as possible) with this PR code.
_(However, it remains unclear why the value of that buffer changed.)_

This PR should be applied early to avoid failing on macos tests.

### DIFF
--- a/src/s3fs_xml.cpp
+++ b/src/s3fs_xml.cpp
@@ -147,19 +147,20 @@ static char* get_object_name(xmlDocPtr doc, xmlNodePtr node, const char* path)
         S3FS_PRN_ERR("could not get object full path name..");
         return NULL;
     }
-    // basepath(path) is as same as fullpath.
-    if(0 == strcmp(reinterpret_cast<char*>(fullpath), path)){
-        xmlFree(fullpath);
+    std::string strfullpath = reinterpret_cast<char*>(fullpath);
+    xmlFree(fullpath);
+
+    // basepath(path) is as same as strfullpath.
+    if(0 == strcmp(strfullpath.c_str(), path)){
         return (char*)c_strErrorObjectName;
     }
 
     // Make dir path and filename
-    std::string   strdirpath = mydirname(std::string(reinterpret_cast<char*>(fullpath)));
-    std::string   strmybpath = mybasename(std::string(reinterpret_cast<char*>(fullpath)));
+    std::string   strdirpath = mydirname(strfullpath.c_str());
+    std::string   strmybpath = mybasename(strfullpath.c_str());
     const char* dirpath = strdirpath.c_str();
     const char* mybname = strmybpath.c_str();
     const char* basepath= (path && '/' == path[0]) ? &path[1] : path;
-    xmlFree(fullpath);
 
     if('\0' == mybname[0]){
         return NULL;


src/s3fs_xml.cpp
```diff
@@ -147,19 +147,20 @@ static char* get_object_name(xmlDocPtr doc, xmlNodePtr node, const char* path)
         S3FS_PRN_ERR("could not get object full path name..");
         return NULL;
     }
-    // basepath(path) is as same as fullpath.
-    if(0 == strcmp(reinterpret_cast<char*>(fullpath), path)){
-        xmlFree(fullpath);
+    std::string strfullpath = reinterpret_cast<char*>(fullpath);
```
gaul, line 150: I'm confused why storing this in a new std::string changes anything?  Doe libxml2 have some kind of race condition?

ggtakec: I couldn't figure out yet what caused this problem occasionally on macos(naybe not have problem on linux).
To reproduce this problem, I modified `test_truncate_cache` as follows and tried it.
(If you make the following changes, there is a high probability that macos will fail.)
```
function test_truncate_cache() {
    describe "Test make cache files over max cache file size ..."
    for dir in $(seq 10); do
        mkdir "${dir}"
        for file in $(seq 120); do
            touch "${dir}/${file}"
        done
        ls "${dir}"
    done
    # shellcheck disable=SC2046
    rm -rf $(seq 10)
}
```
When this test failed, the `get_object_name` function returned `c_strErrorObjectName` for one of files.
The reason was that it did not match `0 == strncmp (dirpath, basepath, strlen (basepath) -1)` in the function.

It happened in only one of the 120 files.
(I don't think it's related to `dirname`/`basename`...)
It seems that `fullpath(xmlChar*)` is a movement that has changed on the way.

If I assume it changed(memory corruption?), I still couldn't figure out the cause.(ex, libxml2 and multithread problem???)

There is a possibility that the cause will be hidden, but by copying the `fullpath` to `std::string` first, the phenomenon can be suppressed, so I think this PR should be merge at first.

ggtakec: I try to actually raise this error and present the log(those varibales value) at that time.(please wait a moment)

ggtakec: I have created a branch(in my repo) that outputs debug logs for testing.
(You can check https://github.com/ggtakec/s3fs-fuse/compare/master...test_pr1994 for the differences.)

This test gave an error(after running it several times).
The test log(https://github.com/ggtakec/s3fs-fuse/runs/7477874519?check_suite_focus=true) is large, so download it if you see it.

From this log, you can see that the error failed due to getting the file stat in the directory when removing the directory.

From this log,  s3fs was trying to remove a directory(`testrun-30516/2`) and it failed.
The reason was that when s3fs got a stat of the files in that directory, removed each files, and tried to remove the directory, but the files remained.
The remaining file was `testrun-30516/2/105`.

At this time, the contents obtained by `list_bucket` from the file list of `testrun-30516/2` are as follows(excerpt): 
`11_Test suite.txt` - L:126094
```
...
...
<Contents>
  <Key>testrun-30516/2/105</Key>
  <LastModified>2022-07-23T02:01:54Z</LastModified>
  <ETag>"d41d8cd98f00b204e9800998ecf8427e"</ETag>
  <Size>0</Size>
  <StorageClass>STANDARD</StorageClass>
  <Owner>
    <ID>75aa57f09aa0c8caeab4f8c24e99d10f8e7faeebf76c078efc7c6caea54ba06a</ID>
    <DisplayName>CustomersName@amazon.com</DisplayName>
  </Owner>
</Contents>
<Contents>
  <Key>testrun-30516/2/106</Key>
  <LastModified>2022-07-23T02:01:54Z</LastModified>
  <ETag>"d41d8cd98f00b204e9800998ecf8427e"</ETag>
  <Size>0</Size>
  <StorageClass>STANDARD</StorageClass>
  <Owner>
    <ID>75aa57f09aa0c8caeab4f8c24e99d10f8e7faeebf76c078efc7c6caea54ba06a</ID>
    <DisplayName>CustomersName@amazon.com</DisplayName>
  </Owner>
</Contents>
...
...
```
However, when `list_bucket` passed this XML to `append_objects_from_xml` and parsed each file with `get_object_name`, the value of a variable is unexpected.

`11_Test suite.txt` - L:126155
```
...
...
s3fs_xml.cpp:get_object_name(155): ============ get_object_name variables {
s3fs_xml.cpp:get_object_name(156):     before fullpath = testrun-30516/2/105
s3fs_xml.cpp:get_object_name(165):     dirpath         = /testrun-30516
s3fs_xml.cpp:get_object_name(166):     mybname         = 105
s3fs_xml.cpp:get_object_name(167):     basepath        = testrun-30516/2/
s3fs_xml.cpp:get_object_name(168):     after fullpath  = testrun-30516/2/105
s3fs_xml.cpp:get_object_name(169): }

s3fs_xml.cpp:get_object_name(155): ============ get_object_name variables {
s3fs_xml.cpp:get_object_name(156):     before fullpath = testrun-30516/2/106
s3fs_xml.cpp:get_object_name(165):     dirpath         = testrun-30516/2
s3fs_xml.cpp:get_object_name(166):     mybname         = 106
s3fs_xml.cpp:get_object_name(167):     basepath        = testrun-30516/2/
s3fs_xml.cpp:get_object_name(168):     after fullpath  = testrun-30516/2/106
s3fs_xml.cpp:get_object_name(169): }
...
...
```
The value of `dirpath` for `testrun-30516/2/105` file is strange.

I'm looking into the source code, but I haven't come up with the real cause yet.

ggtakec:
@gaul 
This fix didn't solve #1987 problem, so I changed this PR to draft .

I tried several fix patterns(including this PR code), but I couldn't resolve this error.
For example, I tried the following:
- Process only with strdirpath without using dirpath
- Call the arguments passed to `dirname` / `basename` in allocated memory
- Copy `Body string` and use it to libxml2 function after that.
- etc.

I think this problem may still be difficult to solve until we know the cause.
Additional, In macos11(not github action), this error hasn't occurred yet.

ggtakec:
@gaul 
This PR code turned out to be useless for fixing #1987.
I submitted #1996, which solves this problem.
So this PR is closed.

PR #1996, opened 2022-07-24
Fixed race condition in dirname and basename call

ggtakec:
#1987
#1994

### Details
The cause of `test_truncate_cache` test error was a race between threads in the `dirname` call.(Only confirmed with macos10)

This happened because in `test_truncate_cache` test, `s3fs_getattr` was called by another thread while calling `list_bucket`.
Specifically, another `dirname` call was made shortly during receiving a static pointer from `dirname` and setting it to `std::string` buffer.
Therefore, the value set in `std::string` was replaced with the result called by another thread.

To fix this, we introduced `mutex` for exclusive control of `dirname` and `basename`(which return pointers to static variables).

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -4529,6 +4529,14 @@ int main(int argc, char* argv[])
         exit(EXIT_FAILURE);
     }
 
+    // mutex for basename/dirname
+    if(!init_basename_lock()){
+        S3FS_PRN_EXIT("could not initialize mutex for basename/dirname.");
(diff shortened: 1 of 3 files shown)


gaul:
It seems like these functions are considered thread-safe yet can return static memory:

> These functions may return pointers to statically allocated memory which  may  be  overwritten  by  subsequent calls.   Alternatively, they may return a pointer to some part of path, so that the string referred to by path should not be modified or freed until the pointer returned by the function is no longer required.

PR #2001, opened 2022-07-25
Set mtime/ctime/atime of all objects as nanosecond

ggtakec:
#2000 

### Details
Changed the stat time of the object for testing in #2000 to nanosecond units.
However, there was a part where the time of some objects(directories) created/modified by s3fs remained time_t/timeval unit.
Some tests could fail instability because of using time as second/usec units.

This PR code changed them to nanosecond units.
As a result, `mtime`/`ctime`/`atime` of all objects created and modified by s3fs are set in nanosecond units.
With the fix, I prepared some common utility functions and decided to use them.

#### NOTE
On macos10, we have detected that `test_mtime_file` fails.
The reason is that when copying a file with `cp -p`, the `mtime` argument of the `s3fs_utimens` call should pass the `mtime` of the original file if it is in the correct state, but this value was different.
In macos10, this `mtime` is rounded down to microseconds instead of nanoseconds.
I could not know how to fix this.(in the first place, it is hard to think that it can be fixed with s3fs)
Therefore, I changed the `test_mtime_file` test result scrutiny only slightly for macos.

### DIFF
--- a/src/s3fs_util.cpp
+++ b/src/s3fs_util.cpp
@@ -447,6 +447,129 @@ void print_launch_message(int argc, char** argv)
     S3FS_PRN_LAUNCH_INFO("%s", message.c_str());
 }
 
+//-------------------------------------------------------------------
+// Utility for nanosecond time(timespec)
+//-------------------------------------------------------------------
+const struct timespec S3FS_OMIT_TS = {0, UTIME_OMIT};
+
+//
+// result: -1  ts1 <  ts2
+//          0  ts1 == ts2
+//          1  ts1 >  ts2
+//
+int compare_timespec(const struct timespec& ts1, const struct timespec& ts2)
+{
+    if(ts1.tv_sec < ts2.tv_sec){
+        return -1;
+    }else if(ts1.tv_sec > ts2.tv_sec){
+        return 1;
+    }else{
+        if(ts1.tv_nsec < ts2.tv_nsec){
+            return -1;
+        }else if(ts1.tv_nsec > ts2.tv_nsec){
+            return 1;
+        }
+    }
+    return 0;
+}
+
+//
+// result: -1  st <  ts
+//          0  st == ts
+//          1  st >  ts
+//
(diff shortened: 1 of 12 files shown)


src/s3fs_util.cpp
```diff
@@ -447,6 +447,129 @@ void print_launch_message(int argc, char** argv)
...
+//
+int compare_timespec(const struct stat& st, stat_time_type type, const struct timespec& ts)
+{
+    struct timespec st_ts;
+    set_stat_to_timespec(st, type, st_ts);
+
+    return compare_timespec(st_ts, ts);
+}
+
+void set_timespec_to_stat(struct stat& st, stat_time_type type, const struct timespec& ts)
```
gaul, line 489: Is it better to return a `struct stat` instead?

src/s3fs_util.cpp
```diff
@@ -447,6 +447,129 @@ void print_launch_message(int argc, char** argv)
...
+           ts.tv_nsec = st.st_ctimespec.tv_nsec;
+        #else
+           ts         = st.st_ctim;
+        #endif
+    }else{
+        S3FS_PRN_ERR("unknown type(%d), so use 0 as timespec.", type);
+        ts.tv_sec     = 0;
+        ts.tv_nsec    = 0;
+    }
+    return &ts;
```
gaul, line 548: Return by value?

gaul (changes requested):

src/s3fs_util.cpp
ggtakec: I wanted to simply call this function in `s3fs_str_realtime` function, so this function returns the pointer of the argument variable(ts).

src/s3fs_util.cpp
ggtakec: This function could return a pointer to the argument(st), but so far this usage has no chance to use now.
If necessary in the future, I will consider modifying it in that way.

PR #2003, opened 2022-07-27
Fixed deadlock in S3fsCurl::DestroyCurlHandle

ggtakec:
n/a

### Details
In the case where `S3fsCurl::DestroyCurlHandle` is called from `S3fsCurl::CreateCurlHandle`, there was a flow that caused a deadlock.

#### NOTE
Now, the build of macos10 for Github Actions is blocked for 24H.
I'll try again later.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -1920,7 +1920,7 @@ bool S3fsCurl::CreateCurlHandle(bool only_pool, bool remake)
     AutoLock lock(&S3fsCurl::curl_handles_lock);
 
     if(hCurl && remake){
-        if(!DestroyCurlHandle(false)){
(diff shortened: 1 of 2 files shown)


src/curl.cpp
```diff
@@ -1920,7 +1920,7 @@ bool S3fsCurl::CreateCurlHandle(bool only_pool, bool remake)
     AutoLock lock(&S3fsCurl::curl_handles_lock);
 
     if(hCurl && remake){
-        if(!DestroyCurlHandle(false)){
+        if(!DestroyCurlHandle(false, true, true)){
```
gaul, line 1923: Should this pass `AutoLock::ALREADY_LOCKED` instead?

ggtakec: It has changed.

PR #2006, opened 2022-07-28
Remove more uses of `const_cast`

gaul:
Follows on to #2004.


src/s3fs.cpp
```diff
@@ -3759,6 +3759,7 @@ static bool set_mountpoint_attribute(struct stat& mpst)
 //
 static int set_bucket(const char* arg)
 {
+    // TODO: Mutates input.  Consider some other tokenization.
```
ggtakec, line 3762: It should use strdup for bucket_name from arg.

ggtakec (approved):

PR #2011, opened 2022-07-29
Remove more unneeded headers identified by IWYU

gaul:


ggtakec (approved):

gaul:
This shows about a 6% improvement when compiling on a Raspberry Pi 4 with `-O0` and `make -j4`:

3928a7e359053aefb370d222306307e46cc8e9b1:
```
real    0m39.212s
user    2m1.274s
sys     0m16.591s
```

48e9e51f4fd9042b5a83d59aa00042fa8c68b043:
```
real    0m36.800s
user    1m56.508s
sys     0m15.287s
```

I suspect that there are some further gains if we reorganize headers.

Issue #2022, opened 2022-08-12
A subdirectory is not accessible just after mounting S3 bucket


Mage152:
### Additional Information
_The following information is very important in order to help us to help you. Omission of the following details may delay your support request or receive no attention at all._
_Keep in mind that the commands we provide to retrieve information are oriented to GNU/Linux Distributions, so you could need to use others if you use s3fs on macOS or BSD_

#### Version of s3fs being used (s3fs --version)
```
# s3fs --version | head -1
Amazon Simple Storage Service File System V1.91 (commit:unknown) with OpenSSL
```

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
```
# rpm -q fuse
fuse-2.9.2-11.el7.x86_64
```

#### Kernel information (uname -r)
```
# uname -r
3.10.0-957.27.2.el7.x86_64
```

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
```
# cat /etc/redhat-release 
CentOS Linux release 7.6.1810 (Core)
```

#### s3fs command line used, if applicable
```
```
#### /etc/fstab entry, if applicable
```
# tail -1 /etc/fstab
bucket_name     /s3/my_directory   fuse.s3fs del_cache,allow_other,uid=1021,gid=1020,umask=000,_netdev,logfile=/var/log/s3fs.log,dbglevel=information,use_cache=/tmp/ 0 0
```

#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
I specified into /etc/fstab the options

- logfile=/var/log/s3fs.log
- dbglevel=information

### Details about issue
Just after mount /s3/my_directory directory, the following command does not work
```
$ ll /s3/my_directory/DATA/SUBDIR/
ls: cannot access /s3/my_directory/DATA/SUBDIR/: Not a directory
```

It works, if I do this:
```
myuser-myserver$ ll /s3/my_directory/DATA/SUBDIR/
ls: cannot access /s3/my_directory/DATA/SUBDIR/: Not a directory
myuser-myserver$ ll /s3/my_directory/DATA/
total 4
drwxrwxrwx 1 
drwxrwxrwx 1 
-rwxrwxrwx 1 

myuser-myserver$ ll /s3/my_directory/DATA/SUBDIR/
total 432274
drwxrwxrwx 1 
drwxrwxrwx 1 
-rwxrwxrwx 1 
[...]

PR #2023, opened 2022-08-14
Support the object under no directory object path by compat_dir

ggtakec:
#2022 

### Details
About #2022
If user uploads a file that includes a non-existent directory path from the aws command, etc., the stat information(`dir/` object) of the intermediate directory does not exist.
So s3fs can not list those directories.
However, user can access files by specifying a file path that includes a directory.
Specifying `compat_dir` does not change this phenomenon.

This fixes:
Fixed so that when `compat_dir` is specified, the directory can be listed even if the stat information(`dir/` object) of the intermediate directory does not exist.
I have added the test code, but a part of test has not been run because `compat_dir` is not included in the s3fs boot options(on the test).
However, my manual testing has been performed and it was succeed.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -48,6 +48,7 @@
 #include "s3fs_util.h"
 #include "mpu_util.h"
 #include "threadpoolman.h"
+#include "autolock.h"
 
 //-------------------------------------------------------------------
 // Symbols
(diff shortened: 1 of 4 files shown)


src/s3fs.cpp
```diff
@@ -2746,6 +2781,55 @@ static int readdir_multi_head(const char* path, const S3ObjList& head, void* buf
...
+        dummy_header["x-amz-meta-uid"]   = str(is_s3fs_uid ? s3fs_uid : geteuid());
+        dummy_header["x-amz-meta-gid"]   = str(is_s3fs_gid ? s3fs_gid : getegid());
+        dummy_header["x-amz-meta-mode"]  = str(S_IFDIR | (~dirmask & (S_IRWXU | S_IRWXG | S_IRWXO)));
+        dummy_header["x-amz-meta-atime"] = str(0);
+        dummy_header["x-amz-meta-ctime"] = str(0);
+        dummy_header["x-amz-meta-mtime"] = str(0);
+
+        for(s3obj_list_t::iterator reiter = notfound_param.notfound_list.begin(); reiter != notfound_param.notfound_list.end(); ++reiter){
+            int dir_result;
+            if(0 == (dir_result = directory_empty(reiter->c_str()))){
```
gaul, line 2804: Does this call `directory_empty` and thus `list_bucket` once per not found object?  This will have bad performance.

ggtakec: Yes, as you pointed out, this will degrade performance.
(However, only when the `compat_dir` option is specified and the directory is not found.)

For simple way, conditions that affect performance can be mitigated by specifying a level in the option of `compat_dir`(ex, old_type_dir, no_dir_type, both...) or preparing another option.

In the future, performance can be improved by not calling `directory_empty` and executing Head requests in parallel or run `directory_empty` in parallel.

However, this performance loss may not have a widespread impact, given that `compat_dir` is disabled by `default`.
(Since it is accumulated in the stat cache, this serial requests does not occur every time.)

ggtakec:

Due to the default value of `compat_dir`, the performance degradation caused by this code fix will not be affected by default.
And the bug cannot be ignored, so we will merge it.

Issue #2027, opened 2022-09-02
CPU usage and I/O are high 


bhuvanp1305:
### Details about the issue
When we call s3fs with the following options,
```
mountS3fs(){
    log "Attempting to mount s3fs fuse"
    s3fs "${s3Bucket}" "${mountPath}" \
    -o ro \
    -o noatime \
    -o passwd_file="${s3cred},use_path_request_style,url=${s3Endpoint}"
}
```
As per the load-model applied, file collection is happening every 5 mins, so when it again tries next time it fetches the object that CPU usage and I/O became high

<img width="726" alt="s3fs-cpu-usage-high" src="https://user-images.githubusercontent.com/15911292/188074206-3d8e68f0-ddf0-4d8d-af1a-67ca47bc2b44.png">

#### Version of s3fs being used (s3fs --version)
_sh-4.4# s3fs --version
Amazon Simple Storage Service File System V1.91 (commit:unknown) with OpenSSL
Copyright (C) 2010 Randy Rizun <rrizun@gmail.com>
License GPL2: GNU GPL version 2 <[https://gnu.org/licenses/gpl.html>](https://gnu.org/licenses/gpl.html%3E)
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law._

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
_sh-4.4# rpm -qi fuse
Name        : fuse
Version     : 2.9.7
Release     : 15.el8
Architecture: x86_64
Install Date: Fri Jul 22 20:33:52 2022
Group       : Unspecified
Size        : 208300
License     : GPL+
Signature   : RSA/SHA256, Fri Feb 25 18:38:22 2022, Key ID 199e2f91fd431d51
Source RPM  : fuse-2.9.7-15.el8.src.rpm
Build Date  : Thu Feb 24 17:57:08 2022
Build Host  : x86-vm-55.build.eng.bos.redhat.com
Relocations : (not relocatable)
Packager    : Red Hat, Inc. <[http://bugzilla.redhat.com/bugzilla>](http://bugzilla.redhat.com/bugzilla%3E)
Vendor      : Red Hat, Inc.
URL         : http://fuse.sf.net/
Summary     : File System in Userspace (FUSE) v2 utilities
Description :
With FUSE it is possible to implement a fully functional filesystem in a
userspace program. This package contains the FUSE v2 userspace tools to
[...]

Issue #2019, opened 2022-08-10
When tested with fio and the -numjobs parameter is added, performance drops considerably compared to not adding it


zhijian-pro:
### Additional Information
_The following information is very important in order to help us to help you. Omission of the following details may delay your support request or receive no attention at all._
_Keep in mind that the commands we provide to retrieve information are oriented to GNU/Linux Distributions, so you could need to use others if you use s3fs on macOS or BSD_

#### Version of s3fs being used (s3fs --version)
Amazon Simple Storage Service File System V1.90 (commit:unknown) with GnuTLS(gcrypt)
Copyright (C) 2010 Randy Rizun <rrizun@gmail.com>
License GPL2: GNU GPL version 2 <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
```
root@tmp-fio-bench:~# find / -name libfuse.so*
/snap/lxd/22923/lib/x86_64-linux-gnu/libfuse.so.2
/snap/lxd/22923/lib/x86_64-linux-gnu/libfuse.so.2.9.9
/usr/lib/x86_64-linux-gnu/libfuse.so.2.9.9
/usr/lib/x86_64-linux-gnu/libfuse.so.2
```

#### Kernel information (uname -r)
_command result: 5.15.0-1015-aws

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
```
PRETTY_NAME="Ubuntu 22.04 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

#### s3fs command line used, if applicable
```
s3fs zjtest1 /tmp/s3fs1 -o endpoint=ap-southeast-1 -o passwd_file=${HOME}/.passwd-s3fs -o dbglevel=info -f -o curldbg
```

### Details about issue

[...]

PR #2028, opened 2022-09-04
Improved fdcache_page performance

ggtakec:
#2019 #2027

### Details
The following issues were reported.
- High CPU load when running s3fs (#2027)
- Extremely poor performance in `fio` test with multi number of job (#2019)

Both of these seem to be caused by the processing by the `page list` of `fdcache` operation in the `PageList` class.
There was a problem with `page list` compress and data extraction logic, resulting in poor performance.

Below are the values when measured with `gprof`:
```
 %   cumulative   self              self     total
time   seconds   seconds    calls  us/call  us/call  name
71.51      1.33     1.33    49576    26.83    26.83  raw_compress_fdpage_list(std::__cxx11::list<fdpage, std::allocator<fdpage> > const&, bool, bool, bool, bool) [clone .constprop.56]
13.44      1.58     0.25    12672    19.73    19.73  PageList::BytesModified() const
10.22      1.77     0.19    13184    14.41    14.41  PageList::GetTotalUnloadedPageSize(long, long, long) const
 3.76      1.84     0.07    49567     1.41    28.24  PageList::Compress()
 0.54      1.85     0.01    24964     0.40    28.22  PageList::SetPageLoadedStatus(long, long, PageList::page_status, bool)
...
...
```

Reviewed related functions including `raw_compress_fdpage_list` in problem.(Reduced useless copies, etc.)
I have confirmed that this fix does not cause a high load on the CPU.

### DIFF
--- a/src/fdcache_page.cpp
+++ b/src/fdcache_page.cpp
@@ -39,12 +39,14 @@ static const int CHECK_CACHEFILE_PART_SIZE = 1024 * 16;    // Buffer size in Pag
 // fdpage_list_t utility
 //------------------------------------------------
 // Inline function for repeated processing
-inline void raw_add_compress_fdpage_list(fdpage_list_t& pagelist, fdpage& page, bool ignore_load, bool ignore_modify, bool default_load, bool default_modify)
+inline void raw_add_compress_fdpage_list(fdpage_list_t& pagelist, const fdpage& orgpage, bool ignore_load, bool ignore_modify, bool default_load, bool default_modify)
 {
-    if(0 < page.bytes){
+    if(0 < orgpage.bytes){
         // [NOTE]
         // The page variable is subject to change here.
         //
+        fdpage page = orgpage;
+
         if(ignore_load){
             page.loaded   = default_load;
         }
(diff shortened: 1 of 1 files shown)


bhuvanp1305:
@ggtakec @gaul any timeline for the merge of this PR?

ggtakec:
@bhuvanp1305 Sorry to keep you waiting, please wait a little longer.
Is it possible for you to try this code?

SanjeetKr7:
@ggtakec 
Yes we tried this code and installed all the required dependencies in our container.
```dependencies = automake gcc-c++ make git tar libcurl-devel libxml2-devel openssl-devel mailcap fuse```

but It is not able to mount the file from s3 bucket and we are getting this error
```
sftp_user2@sftp-server-84c69445fd-p87lr tmp]$ cat tmp__home_sftp_user2_nedata_perfs.log
2022-09-08T05:37:16.750Z [INF] curl.cpp:InitMimeType(428): Loaded mime information from /etc/mime.types
2022-09-08T05:37:16.754Z [INF] s3fs.cpp:s3fs_init(3502): init v1.91(commit:1946ac4) with OpenSSL
2022-09-08T05:37:38.357Z [ERR] s3fs_xml.cpp:get_base_exp(113): marker_xp->nodesetval is empty.
2022-09-08T05:42:31.517Z [ERR] s3fs_xml.cpp:get_base_exp(113): marker_xp->nodesetval is empty.```

ggtakec:
@SanjeetKr7 
I think this error in `get_base_exp` has nothing to do with the modified code in this PR.
The part fixed in this PR is related to the local cache of files (objects).
I think the `get_base_exp` error is related to `listbucket`, so is it occurring even if the PR code is not adapted?
(If so, I'm sure it's something else.)

SanjeetKr7:
> @SanjeetKr7 I think this error in `get_base_exp` has nothing to do with the modified code in this PR. The part fixed in this PR is related to the local cache of files (objects). I think the `get_base_exp` error is related to `listbucket`, so is it occurring even if the PR code is not adapted? (If so, I'm sure it's something else.)

It was not occurring before the PR code adopted

SanjeetKr7:
@ggtakec @gaul 
We were not facing any type of file mounting issue(get_base_exp error) with `v1.89` and `v1.91` of s3fs-fuse. This issue only appears after taking the source of this PR
``` 
Able to mount the files
	ocp lab + alpine image Dockerfile + s3fs v1.89  => successful
	ocp lab + alpine image Dockerfile + s3fs v1.91  => successful
Not able to mount the files
	ocp lab + alpine image Dockerfile + s3fs master => failing
patch is on top of master
        ocp lab + alpine image Dockerfile + s3fs master + patch => failing
```
Due to this issue, we deviated from our original issue which is CPU high usage. Please help us to resolve this issue so that we can focus on our main concern of CPU high usage.

ggtakec:
@SanjeetKr7 
I think your error has a different cause than the fix code in this PR.
Probably, the current master code (changed from v1.91) seems to be the cause.

Is it possible to check a little more detailed log by adding `dbglevel=dbg` and `curldbg` options?
I have created a separate issue(#2036), so please let me know the detailed results there.

Issue #2033, opened 2022-09-16
Using S3FS with OpenSSH SFTP Chroot causes EACCES despite correct permissions


michaelsmoody:
### Additional Information
_The following information is very important in order to help us to help you. Omission of the following details may delay your support request or receive no attention at all._
_Keep in mind that the commands we provide to retrieve information are oriented to GNU/Linux Distributions, so you could need to use others if you use s3fs on macOS or BSD_

#### Version of s3fs being used (s3fs --version)
s3fs 1.91

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
FUSE 2.9.2

#### Kernel information (uname -r)
5.10.135-122.509.amzn2.x86_64

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
Amazon Linux 2

#### s3fs command line used, if applicable
```
s3fs -o url=https://s3-us-gov-west-1.amazonaws.com -o endpoint=us-gov-west-1 -o allow_other -o mp_umask=022 -o dev -o suid -o use_xattr cust-s3-transfer-storage:/FOLDER /mnt/folder
```
#### /etc/fstab entry, if applicable
```
```
#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
_if you execute s3fs with dbglevel, curldbg option, you can get detail debug messages_
```
```
### Details about issue

With an identical working setup on EFS (works perfectly with the following) in `/etc/ssh/sshd_config`:

```
Subsystem sftp internal-sftp -f LOCAL7 -l INFO
ForceCommand internal-sftp -d /mnt/folder -u 0007 -f LOCAL7 -l INFO
# Chroot the connection into the specified directory
ChrootDirectory /mnt/folder
```

And with relevant groups setup on the remote filesystem (EFS/NFS), S3FS fails to work the same way. We were investigating this as a possible "swap" for AWS Transfer Family (which can also use an S3 storage backend, but maintaining POSIX).

Logs in the chroot SFTP show:

`error: chdir to /mnt/folder failed: No such file or directory`

Additionally, and much more importantly!

[...]

Issue #775, opened 2018-06-07
ACL doesn't fully work?


mateusz-:
#### Version of s3fs being used (s3fs --version)
1.82

#### Version of fuse being used (pkg-config --modversion fuse)
2.9.4

#### System information (uname -r)
4.14.42-52.37.amzn1.x86_64

#### Distro (cat /etc/issue)
Amazon Linux AMI release 2018.03

#### /etc/fstab entry (if applicable): 
```
btr-ftp     /ftp        fuse.s3fs _netdev,allow_other,use_xattr 0 0
```
### Details about issue

Below, I created a new dir in my home directory, set default ACL permissions for that dir, and then created a new file in that dir. Lastly, I verified ACL on that new file.

I then proceeded to do the same in the /ftp directory, which is a s3fs mount. The result is not the same.

```sh
$ mkdir foo
$ setfacl -d -m g:btrst:rwx foo
$ touch foo/bar
$ getfacl foo/bar 
# file: foo/bar
# owner: ec2-user
# group: ec2-user
user::rw-
group::rwx      #effective:rw-
group:btrst:rwx     #effective:rw-
mask::rw-
other::r--
$ mkdir /ftp/foo
$ setfacl -d -m g:btrst:rwx /ftp/foo
$ touch /ftp/foo/bar
$ getfacl /ftp/foo/bar
getfacl: Removing leading '/' from absolute path names
# file: ftp/foo/bar
# owner: ec2-user
# group: ec2-user
user::rw-
group::rw-
other::r--
$ getfacl /ftp/foo/
getfacl: Removing leading '/' from absolute path names
# file: ftp/foo/
# owner: ec2-user
# group: ec2-user
user::rwx
group::rwx
other::r-x
default:user::rwx
default:group::rwx
default:group:btrst:rwx
default:mask::rwx
default:other::r-x
```

tam116:
Seeing the same thing, also sticky bit doesn't seem to work either

gaul:
Related to #844 and #1517.

michaelsmoody:
This is one of an unfortunate _number_ of blocking issues that is stopping me from using S3FS in a production implementation and has, well, (and I really hate to say this, because it sounds....unforgiving) caused a waste of around 3 days of work chasing various bugs and problems. I'll be opening bug reports about them, but, this one is a deeply blocking issue (POSIX compatibility being a complete necessity).

[...]

Issue #1769, opened 2021-10-03
Upgrade macOS CI 


gaul:
I saw this warning today:

> macOS-latest workflows will use macOS-11 soon. For more details, see https://github.com/actions/virtual-environments/issues/4060

But I believe that osxfuse does not support macOS 11 or requires some rebooting or other difficulties.

ggtakec:
@gaul Thanks for your infromation.
I would be happy if following problem was solved.
I will try it once around 11/3.
https://github.com/s3fs-fuse/s3fs-fuse/blob/master/.github/workflows/ci.yml#L120-L126

ggtakec:
I noticed the following error for macos in the Schedule build of Github Actions on this Sunday.
```
configure: error: Package requirements (fuse >= 2.7.3 libcurl >= 7.0 libxml-2.0 >= 2.6 libcrypto >= 0.9 ) were not met:
No package 'libcrypto' found
```

ggtakec:
**The macOS 10.15 Actions runner image will be fully unsupported by 8/30/22 in GitHub**

I've picked up this issue from the Issue at https://github.com/actions/virtual-environments/issues/5583.
And today I saw an errors about this several times, when the test run on macOS runner.

Currently, Github Actions for s3fs test has no choice but to use macOS 10.15. That's because we get an error when installing osxfuse if we use another macOS version.
See: https://github.com/actions/virtual-environments/issues/598 , https://github.com/actions/virtual-environments/issues/4731

We need to install osxfuse on macOS11 runner, but we can not have done it yet.

ggtakec:
Currently, in Github Action, the following error is displayed and it can not be solved.
> This is a scheduled macOS-10.15 brownout. The macOS-10.15 environment is deprecated and will be removed on August 30th, 2022. For more details, see https://github.com/actions/virtual-environments/issues/5583

ggtakec:
Since `macos-12` Github Actions runner is now supported, I checked again with both `macos-11` and `12`.

In conclusion, it is unlikely that you can install and use `macfuse` with Github Actions Runner.
This is because `macfuse` is not allowed in macOS device policies, so it cannot be used without rebooting.

For reference, if your macOS device policies are allowed, you can view them like this:
```
$ sudo sqlite3 /private/var/db/SystemPolicyConfiguration/KextPolicy "select * from kext_policy;"
[...]

PR #2039, opened 2022-09-25
Added support for xattr as POSIX ACL

ggtakec:
#775 #2033

### Details
s3fs' xattr support only worked as Extended attributes.(such as setxattr)
We did not implement the function as `POSIX ACL` pointed out in the issues, so I implemented this.(such as setfacl)

I have also created a test script and it is passing the test without any issues.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -1309,6 +1309,196 @@ function test_rm_rf_dir {
    fi
 }
 
+function test_posix_acl {
+    describe "Testing posix acl function ..."
+
+    #------------------------------------------------------
+    # Directory
+    #------------------------------------------------------
+    local POSIX_ACL_TEST_DIR1="posix_acl_dir1"
+    local POSIX_ACL_TEST_DIR2="posix_acl_dir2"
+    mkdir "${POSIX_ACL_TEST_DIR1}"
+
+    #
+    # Set posix acl(not default)
+    #
+    setfacl -m "u:${USER}:rwx" "${POSIX_ACL_TEST_DIR1}"
+    if ! getfacl "${POSIX_ACL_TEST_DIR1}" | grep -q "^user:${USER}:rwx"; then
+       echo "Could not set posix acl(not default) to ${POSIX_ACL_TEST_DIR1} directory"
+       return 1
+    fi
+
+    #
+    # Set posix acl(default)
+    #
+    setfacl -d -m "u:${USER}:rwx" "${POSIX_ACL_TEST_DIR1}"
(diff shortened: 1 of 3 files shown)


test/small-integration-test.sh
```diff
@@ -42,7 +42,7 @@ export CACHE_DIR
 export ENSURE_DISKFREE_SIZE 
 if [ -n "${ALL_TESTS}" ]; then
     FLAGS=(
-        "use_cache=${CACHE_DIR} -o ensure_diskfree=${ENSURE_DISKFREE_SIZE} -o fake_diskfree=${FAKE_FREE_DISK_SIZE}"
+        "use_cache=${CACHE_DIR} -o ensure_diskfree=${ENSURE_DISKFREE_SIZE} -o fake_diskfree=${FAKE_FREE_DISK_SIZE} -o use_xattr"
```
gaul, line 45: Should we have a separate flag for xattr instead of including it in the cache tests?

test/integration-test-main.sh
```diff
@@ -1309,6 +1309,196 @@ function test_rm_rf_dir {
...
 
+function test_posix_acl {
+    describe "Testing posix acl function ..."
+
+    #------------------------------------------------------
+	# Directory
+    #------------------------------------------------------
+	local POSIX_ACL_TEST_DIR1="posix_acl_dir1"
+	local POSIX_ACL_TEST_DIR2="posix_acl_dir2"
+    mkdir "${POSIX_ACL_TEST_DIR1}"
```
gaul, line 1320: Could you fix the indentation in this file?  I think some tabs should be expanded to spaces.

gaul (changes requested):
Please make small changes.

test/small-integration-test.sh
ggtakec: _Sorry for my late reply_
It's best to have separate tests for xattr, but I don't think they affect each other when tested with other options.
(If we add another test pattern(option pair), the test time will be longer, so I don't think it's necessary to do that.)

test/integration-test-main.sh
ggtakec: It's my fault, thus it is fixed.

ggtakec:
I removed macos10 runner from required.
https://github.com/s3fs-fuse/s3fs-fuse/issues/1769#issuecomment-1271660587

test/integration-test-main.sh
ggtakec: _additional comment_
There was still a missing fix. Fixed and rebased.

ggtakec:

I'm merging this PR to move forward with code updates.

PR #2052, opened 2022-11-05
Updated comments in curl.cpp (related to #1827)

ggtakec:
#1827 

### Details
I have chnaged the comment and added it as a note.


ggtakec:
@gaul 
Since it's only a comment update, I'll merge without waiting for reviews.

PR #2053, opened 2022-11-05
Updated actions/checkout for gha from v2 to v3

ggtakec:
n/a

### Details
Github Actions has the following warning, so I upgraded the version of `actions/checkout.`
```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/checkout
```


ggtakec:
@gaul I'm sorry, since it is only an action/checkout version upgrade, I will merge it without waiting for a review.

Issue #2051, opened 2022-11-04
Improve performance of readdir


huntersman:
#### Version of s3fs being used (s3fs --version)
1.91
#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
2.9.2
#### Kernel information (uname -r)
5.4.211-1.el7.elrepo.x86_64
#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
CentOS 7
#### s3fs command line used, if applicable
```
s3fs bucket /mount/bucket -o passwd_file=${HOME}/.passwd-s3fs -o url=http://ip -o use_path_request_style -o noxmlns -o dbglevel=error -o default_acl=public-read -o logfile=/var/log/s3fs.log -o allow_other -o multireq_max=1000
```
### Details about issue
It takes about 30 seconds for `ls` 100 thousands files, maybe it could be faster.

gaul:
s3fs 1.92 reduces `readdir` S3 operations and thus performance by 33%.  However something like #1482 is necessary for better performance.

PR #2056, opened 2022-11-15
Use cache to improve readdir performance

huntersman:
#2051

### Details
When a directory has a lot of files, `s3fs_readdir` will take a lot of time because of `list_bucket` and `readdir_multi_head` is time-consuming. One possible solution is that we maintain a map of the path and files under that path, and files stat information should also be cached. When we `readdir` from a certain path, we can get data from the map instead of sending request to s3.

However, there are two defects. Firstly, files should only be modified via `s3fs`, any modification on s3 server will not change the map. In other words, files are read only on s3. Secondly, map is not the best way because when we reboot the server, data will disappear, maybe we could use a database to store it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -34,7 +34,7 @@ AC_CHECK_HEADERS([attr/xattr.h])
 AC_CHECK_HEADERS([sys/extattr.h])
 AC_CHECK_FUNCS([fallocate])
 
-CXXFLAGS="$CXXFLAGS -Wall -fno-exceptions -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2"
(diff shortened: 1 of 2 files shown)


ggtakec:
@huntersman Thank you for this PR.
This fix maybe make it faster.
However, due to the history of s3fs, I think s3fs should not cache every file path accessing.
It assumes that all access is done via s3fs, as you mentioned in your PR comment, so we are not using that logic.
Instead, s3fs has the ability to cache stat information(that is controled by timeout, etc.).
Of course, the first access does not make sense to cache, but from the second access onwards, the cache works effectively.
If you assume all accessing via s3fs, would it be possible to increase the size of the stat cache and lengthen the cache timeout?

I'm planning to fix the readdir(HEAD requests) and stat cache related fix after the next release.
I think that will improve the performance a little more.

huntersman:
Close due to s3fs should not cache every file path accessing.

PR #2016, opened 2022-07-30
Added the function to update mtime/ctime of the parent directory

ggtakec:
### Relevant Issue (if applicable)
#1961 #1962 #1964 #2024

### Details
This fix, the `mtime`/`ctime` of the parent directory is updated when the `i-node` of the parent directory is updated(create/delete files and directories, etc.).

The operation to update `mtime`/`ctime` in the parent directory is shown below:
| Operation                                                                  | ctime/mtime  |
| -------------------------------------------------------------------------- | ------------ |
| Create file/symbolic file/directory(`touch`, `cp`, `ln -s`, `mkdir`, etc)  | update       |
| Rename file/symbolic file/directory(`mv`, etc)                             | update       |
| Update file/symbolic file/directory(`touch`, `chmod`, `chown`, chgrp`, etc)| not update   |

I am adding a test(`test_update_parent_directory_time`) for the above pattern.

This fix will perform updating stats of the parent directory for each operation.
And this updating stats is enabled by default.
However, some users may not need this update stats process, and I have added the option `disable_update_parent_dir_stat` to suppress it.

This PR will resolve the issues in #1961 and #1962.
After #1964 will be merged, this branch will be rebased and release Draft.

This PR contains the same code as #1964.
(This is necessary and dependent it for conducting the test.)

PR #2057, opened 2022-11-19
Fixed a bug when end of path for mount point is multi slash

ggtakec:
#2016

### Details
s3fs expects that the last character of the path is not `/` when the mount point is specified by `<bucket name> + <path(ex. /path)>`.
If a charactor `/` terminated path(ex. '/path/') will be specified, it will be excluded during option parsing.
But  there was a bug that multiple `/` terminated path(ex. `/path//`).
This PR fixed this bug.

- Impact of this bug
If `/path//` and a mount point were specified, the files under the mount point were invisible.  
(Because it searches the path `<bucket>/path//`)  
Also, although files can be manipulated, a file with a path that s3fs cannot recognize has been created.  
For example, creating a file named `sample` with this mount point created a file named /path//sample`.  

In addition, I think that there were very few cases where the mount point was specified as `path//`, so I imagine that there is almost no impact on users.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -4179,10 +4179,8 @@ static int set_bucket(const char* arg)
                 return -1;
             }
             mount_prefix = pmount_prefix;
-            // remove trailing slash
-            if(mount_prefix[mount_prefix.size() - 1] == '/'){
-                mount_prefix.erase(mount_prefix.size() - 1);
-            }
+            // Trim the last consecutive '/'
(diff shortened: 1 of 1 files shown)


ggtakec:
@gaul 
This fix was minor and passed the test.
As for the impact on the user, if the user specifies the wrong path, I think that it is already not working properly, so I think the impact of the fix is minimal.
(But this bug affects more than PR #2016.)
I apologize in advance for merging without waiting for permission.

PR #2062, opened 2022-11-22
Fix man page inaccuracies

Pashugan:
Subj


gaul:
Thank you for your contribution @Pashugan!

Issue #2061, opened 2022-11-21
Support to compatible with OpenSSL 3.0


ggtakec:
### Details about issue
Fedora:37 and RockyLinux:9 are based on OpenSSL 3.0.
Therefore, it outputs the following warning  when compiling.(the build is successful.)

```
openssl_auth.cpp: In function 'bool s3fs_init_global_ssl()':
openssl_auth.cpp:54:25: warning: 'int ERR_load_BIO_strings()' is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
   54 |     ERR_load_BIO_strings();
      |     ~~~~~~~~~~~~~~~~~~~~^~
In file included from /usr/include/openssl/cryptoerr.h:17,
                 from /usr/include/openssl/crypto.h:38,
                 from /usr/include/openssl/bio.h:30,
                 from /usr/include/openssl/evp.h:30,
                 from openssl_auth.cpp:27:
/usr/include/openssl/cryptoerr_legacy.h:31:27: note: declared here
   31 | OSSL_DEPRECATEDIN_3_0 int ERR_load_BIO_strings(void);
      |                           ^~~~~~~~~~~~~~~~~~~~
openssl_auth.cpp: In function 'unsigned char* s3fs_md5_fd(int, off_t, off_t)':
openssl_auth.cpp:263:13: warning: 'int MD5_Init(MD5_CTX*)' is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  263 |     MD5_Init(&md5ctx);
      |     ~~~~~~~~^~~~~~~~~
In file included from openssl_auth.cpp:29:
/usr/include/openssl/md5.h:49:27: note: declared here
   49 | OSSL_DEPRECATEDIN_3_0 int MD5_Init(MD5_CTX *c);
      |                           ^~~~~~~~
openssl_auth.cpp:278:19: warning: 'int MD5_Update(MD5_CTX*, const void*, size_t)' is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  278 |         MD5_Update(&md5ctx, buf, bytes);
      |         ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
/usr/include/openssl/md5.h:50:27: note: declared here
   50 | OSSL_DEPRECATEDIN_3_0 int MD5_Update(MD5_CTX *c, const void *data, size_t len);
      |                           ^~~~~~~~~~
[...]

PR #2065, opened 2022-11-23
Compatible with OpenSSL 3.0

ggtakec:
#2061

### Details
Fixed the warning that appeared when building OpenSSL 3.0.
MD5_*** and etc has been deprecated and cased to use EVP.

### DIFF
--- a/src/openssl_auth.cpp
+++ b/src/openssl_auth.cpp
@@ -51,7 +51,14 @@ const char* s3fs_crypt_lib_name()
 bool s3fs_init_global_ssl()
 {
     ERR_load_crypto_strings();
+
+    // [NOTE]
+    // OpenSSL 3.0 loads error strings automatically so these functions are not needed.
+    //
+    #ifndef USE_OPENSSL_30
     ERR_load_BIO_strings();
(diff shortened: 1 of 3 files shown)


src/openssl_auth.cpp
```diff
@@ -238,8 +245,66 @@ bool s3fs_HMAC256(const void* key, size_t keylen, const unsigned char* data, siz
...
+        char        buf[len];
+        bytes = len < (size - total) ? len : (size - total);
+        bytes = pread(fd, buf, bytes, start + total);
+        if(0 == bytes){
+            // end of file
+            break;
+        }else if(-1 == bytes){
+            // error
+            S3FS_PRN_ERR("file read error(%d)", errno);
+            return NULL;
```
gaul, line 291: Should this call `EVP_MD_CTX_free`?

gaul (changes requested):

src/openssl_auth.cpp
ggtakec: Sorry for my fault, it is fixed and updated soon.

Issue #2069, opened 2022-11-29
Cannot set `http_proxy` environment variable in fstab


gaul:
s3fs may need to add a flag to directly support mounting via fstab while setting a proxy.  Context:

https://serverfault.com/questions/1019109/mount-aws-s3-on-boot-while-behind-proxy-using-s3fs

ggtakec:
The same issue was also pointed out in #1861. I'm sorry for not start to work for it yet.

ggtakec:
Now that we have completed the implementation of proxy options, I will close this issue.

PR #2075, opened 2022-12-04
Added proxy and proxy_cred_file option

ggtakec:
#2069

### Details
This PR is implemented the 'proxy' and 'proxy_cred_file' options with reference to the 'curl' command.
It is almost equivalent to curl's `--proxy`, `--proxy-user` option.

Currently we can't have all the complex options like curl does, so we only implement simple functionality.
The credential for HTTP proxy authentication is loaded username and passphrase from a file instead of options(for secure).

@gaul Testing of this fix is not ready.
Do you have any ideas? (ex. using haproxy)


gaul:
We can use [Chaos HTTP Proxy](https://github.com/bouncestorage/chaos-http-proxy) to test this.  You just need to unset `http_proxy` environment variable and pass the flag instead.  In fact this might be a better approach for the tests.

ggtakec:
@gaul 
For example, I modify the test scripts to add a flag similar to the `CHAOS_HTTP_PROXY` flag, such as `TEST_PROXY_OPTION`, is it ok?
I think it's possible, we can start `Chaos HTTP Proxy` and inject s3fs with proxy settings.
(since you already provided most of it in `integration-test-common.sh`)

However, I am trying following command on this PR branch now, but I am getting an error.
```
$ CHAOS_HTTP_PROXY=1 S3_URL=http://127.0.0.1:8080 make check -C test
```
I got following errors.
```
Connection was closed before we received a valid response from endpoint URL: "http://127.0.0.1:8080/s3fs-integration-test".
make_bucket failed: s3://s3fs-integration-test Connection was closed before we received a valid response from endpoint URL: "http://127.0.0.1:8080/s3fs-integration-test".
```
Do you know the reason for this?

ggtakec:
I could test using the `Chaos HTTP Proxy`.
I've found that this code still needs fixing, so I'll keep testing and fixing it.

ggtakec:
@gaul 
It has been fixed to be able to test with `CHAOS HTTP PROXY`.

Running tests with `CHAOS_HTTP_PROXY_OPT=1` (ex. `CHAOS_HTTP_PROXY_OPT=1 make check`) will run the tests via the `CHAOS HTTP PROXY` using new `proxy` option.
The existing `CHAOS_HTTP_PROXY=1` (ex. `CHAOS_HTTP_PROXY=1 make check`) option has not changed.
It tests with the `http_proxy` environment variable(not changed).

#### Fixed a bug
In `integration-test-common.sh`, there was the following problem with the `CHAOS_HTTP_PROXY` option, so it has been fixed.
I don't think `CHAOS HTTP PROXY` supports HTTPS, so `s3proxy` must be started with `HTTP`.
For this I added the `s3proxy_http.conf` file.
_This is probably because a fix leaked when I changed `s3proxy` to start with `HTTPS`._
_This fix overlaped with the processing for testing the code in this PR, so I fixed it together in this PR._

PR #2078, opened 2022-12-16
Make some methods and parameters const

gaul:
This requires making some locks mutable.


ggtakec (approved):
Certainly, it may be better to qualify methods and variables with const.
Let's unify lock(mutex) with mutable modifier from now on.

PR #2086, opened 2023-01-04
Fixed a conflict between curl and curl-minimal on RockyLinux 9

ggtakec:
n/a

### Details
`Gihutb Actions` on `Rocky Linux 9` had the following error.
```
Error: 
(try to add '--allowerasing' to command line to replace conflicting packages or '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
 Problem: problem with installed package curl-minimal-7.76.1-19.el9.x86_64
  - package curl-minimal-7.76.1-19.el9.x86_64 conflicts with curl provided by curl-7.76.1-19.el9.x86_64
  - conflicting requests
```

The `Rocky Linux 9` image includes `curl-minimal`, which conflicts with installing the `curl` package.
(the conflict may not have occurred before `7.76.1-19`)

For `Rocky Linux 9` only, this PR adds the `--allowerasing` option when installing `curl` to avoid this.

In the `linux-ci-helper.sh` file, I defined `PACKAGE_INSTALL_ADDITIONAL_OPTIONS` and modified it to specify the above options only for `RockyLinux9`.

We should remove this option if it is no longer needed.

### DIFF
--- a/.github/workflows/linux-ci-helper.sh
+++ b/.github/workflows/linux-ci-helper.sh
@@ -68,6 +68,7 @@ CONFIGURE_OPTIONS="CXXFLAGS='-O -std=c++03 -DS3FS_PTHREAD_ERRORCHECK=1' --prefix
 # Default values
 #
 PACKAGE_ENABLE_REPO_OPTIONS=""
+PACKAGE_INSTALL_ADDITIONAL_OPTIONS=""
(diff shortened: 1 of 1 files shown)


ggtakec:
@gaul Sorry, currently builds on RockyLinux9 keep failing, so I'm merging this PR without waiting for your review.

Issue #2063, opened 2022-11-22
CheckBucket fails with restrictive IAM policy and mounting a path


francisATgwn:
#### Version of s3fs being used (s3fs --version)
```
Amazon Simple Storage Service File System V1.91 (commit:d98fdf4) with OpenSSL
```

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
`2.9.9`

#### Kernel information (uname -r)
`5.15.0-1023-aws`

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
```
PRETTY_NAME="Ubuntu 22.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.1 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

#### s3fs command line used, if applicable
```
sudo s3fs -f REDACTED:/path/to/redacted /mnt/foo -o ro,_netdev,passwd_file=REDACTED,allow_other,umask=0007,nodev,dbglevel=info,curldbg=body
```

#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
```
2022-11-22T20:25:42.934Z [CRT] s3fs_logger.cpp:LowSetLogLevel(240): change debug level from [CRT] to [INF]
2022-11-22T20:25:42.934Z [INF]     s3fs.cpp:set_mountpoint_attribute(4136): PROC(uid=0, gid=0) - MountPoint(uid=0, gid=0, mode=40755)
2022-11-22T20:25:42.938Z [INF] curl.cpp:InitMimeType(428): Loaded mime information from /etc/mime.types
2022-11-22T20:25:42.938Z [INF] fdcache_stat.cpp:CheckCacheFileStatTopDir(78): The path to cache top dir is empty, thus not need to check permission.
2022-11-22T20:25:42.938Z [INF] threadpoolman.cpp:StopThreads(195): Any threads are running now, then nothing to do.
[...]

Issue #1460, opened 2020-10-29
s3fs requires directory object to exist to mount via path


gaul:
Why does s3fs require this?  It can lead to user confusion:

https://stackoverflow.com/questions/64584917/s3fs-bucket-name-contains-illegal-character/64585933

ValentinNikin:
Hello! In my case, on the contrary, it is required that an attempt to mount gives an error if the path does not exist in S3 bucket. Maybe there is some option to do this when trying to mount?

ggtakec:
@ValentinNikin Thanks for the bug report.
It's an unexpected glitch with a series of fixes, so I posted PR #2114 to fix this bug.
Please wait a little longer until this is merged.

PR #2087, opened 2023-01-05
Retry BucketCheck containing directory paths

ggtakec:
#2063 #1728 #1687 #1460

### Details
Fixed a bug where a bucket has no permissions and a directory has permissions, and a mount with that directory fails.
This PR is related to #1728 which was previously fixed.

#### Specification
Here's a bucket check on s3fs startup and a description of how it works in this PR.

The following patterns for mount point are supported by s3fs:

1. Mount the bucket top
2. Mount the directory(folder) under the bucket. In this case, there are the following cases:
(   2A) Directories created by clients other than s3fs
(   2B) Directory created by s3fs

At first, if user has access to the bucket, the checking access to the bucket succeeds and this function returns success.
However, if user does not have access to the bucket and has permissions to the directory, this first check will fail.
But if user specifies the directory for mount point, this function retries to check with the path containing the directory.
And it will be success.

In the case of (2A), the check will succeed if the bucket allows to access, but will fail if permissions are granted only to the directory, as it is not a directory recognized by s3fs.
This combination is not supported by s3fs, so make sure user create the directory before starting s3fs.
In case (2B), if user does not have access to bucket, the first check(to bucket) fails, but the retry check(with path) succeeds.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -3529,10 +3529,13 @@ int S3fsCurl::GetObjectRequest(const char* tpath, int fd, off_t start, off_t siz
     return result;
 }
 
-int S3fsCurl::CheckBucket()
+int S3fsCurl::CheckBucket(const char* check_path)
 {
     S3FS_PRN_INFO3("check a bucket.");
 
+    if(!check_path || 0 == strlen(check_path)){
+        return -EIO;
+    }
     if(!CreateCurlHandle()){
         return -EIO;
     }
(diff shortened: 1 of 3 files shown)


adamqqqplay:
@ggtakec Hi, ggtakec. I'm curious whether we can give priority to the check in (2) when the user specifies a subdirectory to mount? This can avoid error reporting and retrying in (1) when the user has only subdirectory permissions but no bucket permissions.

ggtakec:
@adamqqqplay 
I believe that #2155 is in line with your request.
Please check it, thanks in advance for your help.

Issue #2085, opened 2023-01-03
File copies using multipart upload fail with SSE-C 


icassassin:
#### Version of s3fs being used (`s3fs --version`)
Amazon Simple Storage Service File System V1.91 (commit:14eb1a7) with OpenSSL
Also tried with v1.90 that comes with Ubuntu 22.04; same issue

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)
example: 2.9.9

#### Kernel information (`uname -r`)
5.15.0-56-generic

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)
Ubuntu 22.04.1 LTS

#### How to run s3fs, if applicable
from /etc/fstab:
[bucket name] [mount point] fuse.s3fs _netdev,allow_other,url=https://s3.wasabisys.com,passwd_file=<path>,use_sse=custom:<path2>,use_path_request_style,dbglevel=info 0 0

#### s3fs syslog messages (`grep s3fs /var/log/syslog`, `journalctl | grep s3fs`, or `s3fs outputs`)
I think the relevant syslog lines are:
Jan  3 20:41:06 rdserv-staging s3fs[10246]: curl_multi.cpp:MultiRead(228): failed a request(400: https://s3.wasabisys.com/[bucket name]/Godot_v3.5.1-stable_x11.64.zip?partNumber=3&uploadId=kVWQFryLsXmfu__SQmTnhVLwq2gL6s7CYhMAXzlVQZ9vSXF4ceiFE9FIRxleQExNdgwIvIo_x7HkjCPJAt6lj9XwAfTxynL5ED2otPBmmyTbY40_HjZXwS9Kkq-lT6AH)
Jan  3 20:41:06 rdserv-staging s3fs[10246]:       [tpath=/Godot_v3.5.1-stable_x11.64.zip][start=20971520][size=18752962][part=3]
Jan  3 20:41:06 rdserv-staging s3fs[10246]:       URL is https://s3.wasabisys.com/[bucket name]/Godot_v3.5.1-stable_x11.64.zip?partNumber=3&uploadId=kVWQFryLsXmfu__SQmTnhVLwq2gL6s7CYhMAXzlVQZ9vSXF4ceiFE9FIRxleQExNdgwIvIo_x7HkjCPJAt6lj9XwAfTxynL5ED2otPBmmyTbY40_HjZXwS9Kkq-lT6AH
[...]

Issue #1816, opened 2021-12-10
rsync IO error on ftruncate


skepticalwaves:
#### Version of s3fs being used (s3fs --version)
Amazon Simple Storage Service File System V1.90 (commit:v1.90) with OpenSSL

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
```
# dpkg -s fuse
Package: fuse
Status: install ok installed
Priority: optional
Section: utils
Installed-Size: 113
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Architecture: amd64
Version: 2.9.9-3
Depends: libc6 (>= 2.28), libfuse2 (= 2.9.9-3), adduser, mount (>= 2.19.1), sed (>= 4)
Conffiles:
 /etc/fuse.conf 298587592c8444196833f317def414f2
Description: Filesystem in Userspace
 Filesystem in Userspace (FUSE) is a simple interface for userspace programs to
 export a virtual filesystem to the Linux kernel. It also aims to provide a
 secure method for non privileged users to create and mount their own filesystem
 implementations.
Original-Maintainer: Laszlo Boszormenyi (GCS) <gcs@debian.org>
Homepage: https://github.com/libfuse/libfuse/wiki

```

#### Kernel information (uname -r)
5.4.0-89-generic

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
NAME="Ubuntu"
VERSION="20.04.3 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.3 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal

#### s3fs command line used, if applicable
```
NA
```
#### /etc/fstab entry, if applicable
```
<REDACTED>:/matrix /matrix/s3-media-store/ fuse.s3fs x-systemd.automount,x-systemd.requires=network-online.target,allow_other,url=<REDACTED>,uid=997,gid=1000,use_sse=custom:<REDACTED PATH> 0 0
```
[...]

Issue #1865, opened 2022-01-21
Flush of dirty data w/ SSE custom key fails after first round


lmlsna:
### Additional Information
_The following information is very important in order to help us to help you. Omission of the following details may delay your support request or receive no attention at all._
_Keep in mind that the commands we provide to retrieve information are oriented to GNU/Linux Distributions, so you could need to use others if you use s3fs on macOS or BSD_

#### Version of s3fs being used (s3fs --version)
1.89

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
2.9.9

#### Kernel information (uname -r)
5.13.0-25-generic

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
PRETTY_NAME="Ubuntu 21.10"
NAME="Ubuntu"
VERSION_ID="21.10"
VERSION="21.10 (Impish Indri)"
VERSION_CODENAME=impish
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=impish

#### s3fs command line used, if applicable
```
s3fs -d -f -o max_dirty_data=-1 -o use_sse=custom:/home/ubuntu/s3fs/sse-key -o load_sse_c=/home/ubuntu/s3fs/sse-key -o passwd_file=/home/ubuntu/s3fs/passwd my-bucket:/ /mount/point
```
#### /etc/fstab entry, if applicable
```
```
#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
_if you execute s3fs with dbglevel, curldbg option, you can get detail debug messages_
```
s3fs[3822577]: curl.cpp:RequestPerform(2318): HTTP response code 400, returning EIO. Body Text:
<?xml version="1.0" encoding="UTF-8"?><Error>
<Code>InvalidRequest</Code>
<Message>The multipart upload initiate requested encryption. Subsequent part requests must include the appropriate encryption parameters.</Message>
</Error>

s3fs[3822577]: curl_multi.cpp:MultiPerform(135): thread failed - rc(-5)
```
### Details about issue
[...]

PR #2088, opened 2023-01-06
Fixed a bug about multipart upload with SSE

ggtakec:
#2085 #1816

### Details
In the multipart upload logic with SSE, there was a bug that could not be processed correctly except for copy processing.
This PR is fixed these.


gaul:
Need better test coverage but S3Proxy does not support SSE-C.

ggtakec:
@lmlsna Sorry, I forgot to notify #1865.
We are waiting for your test result.

Issue #1778, opened 2021-10-13
Use with IRSA


shaunc:
### Additional Information
_The following information is very important in order to help us to help you. Omission of the following details may delay your support request or receive no attention at all._
_Keep in mind that the commands we provide to retrieve information are oriented to GNU/Linux Distributions, so you could need to use others if you use s3fs on macOS or BSD_

#### Version of s3fs being used (s3fs --version)
1.85

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
2.9.9

#### Kernel information (uname -r)
4.14.186-146.268.amzn2.x86_64

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
NAME="Ubuntu"
VERSION="20.04.3 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.3 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal

### Details about issue

I am looking for instructions for using with IRSA. In my pod container environment I have AWS_ROLE_ARN and AWS_WEB_IDENTITY_TOKEN_FILE. How can I get s3fs-fuse to consume these?

gaul:
Sorry I'm not familiar with with IRSA.  Later versions of s3fs do support IMDSv2 -- does this work for you?

@nmeyerhans do you know anything about this?

theis188:
Thanks for opening this issue, I would also find this feature useful.

aru-trackunit:
+1

bressanmarcos:
:+1: +1

mak-454:
+1

yao531441:
+1

yukinakanaka:
+1

ggtakec:
As one solution to this issue:

I have released the following source as a prototype few month ago.
https://github.com/ggtakec/s3fs-fuse-awscred-lib
_You can refer to the README.md for how to build/use it._
This library is a DSO that wraps the `aws-sdk-cpp` authentication functionality.
[...]

PR #2101, opened 2023-01-29
Fixed token update process using ext-cred library

ggtakec:
https://github.com/s3fs-fuse/s3fs-fuse/issues/1778#issuecomment-1407172269

### Details
Fixed some incorrect judgments when specifying an external credential authentication library(specified by `credlib` option).

This is a bug reported by @artificial-aidan.
Thanks @artificial-aidan for great help.


ggtakec:
@gaul This PR has no impact unless you have an authentication DSO loaded, so I merge this.
If you have any problems, please point them out.

PR #2114, opened 2023-02-19
Fixed to fail when the mount point directory is not existed

ggtakec:
https://github.com/s3fs-fuse/s3fs-fuse/issues/1460#issuecomment-1425279454 #1728 #2087

### Details
We fixed some mount point checking by #1728 and #2087.
But when specifying `<bucket>/<directory>(/...)` as the mount point, and if that mount point directory does not exist, it is not an error.
This bug was reported by @ValentinNikin.

If this error does not occur, a directory that does not exist will be mounted as a mount point, and directories/files can be manipulated under that directory.

Therefore, this case has been corrected to be an error.
(This fix is same behavior as v1.91 and below)

#### Note
_Even if the directory object as mount point does not exist, an error will not occur if an object containing that path exists._
_For example, if `<bucket>/<directory>/<file>` is created with a CLI tool, `<bucket>/<directory>/<file>` exists, but `<bucket>/<directory>` does not exist._
_And in the case of specifying `<bucket>/<directory>` as the mount point._

This fix will be merged soon.
If the review is not completed by the time of the merge, please let me know later.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -3365,11 +3365,20 @@ static int remote_mountpath_exists(const char* path)
     S3FS_PRN_INFO1("[path=%s]", path);
 
     // getattr will prefix the path with the remote mountpoint
-    if(0 != (result = get_object_attribute("/", &stbuf, NULL))){
+    if(0 != (result = get_object_attribute(path, &stbuf, NULL))){
         return result;
     }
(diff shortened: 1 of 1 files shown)


ggtakec:
@gaul I will merge this, please let me know if any problems.

Issue #2110, opened 2023-02-12
Request to add ALPINE to supported OS(Build/Test)


ggtakec:
### Correspond to ALPINE.
Support ALPINE(v3.17) in GithubActions(CI).
_I don't think there is much to fix._

The packages required for building and the contents that need to be modified are as follows:
- Requires OpenSSL 3.0 support  
already supported by master
- Dependent packages required for building  
automake, autoconf, util-linux-misc, fuse, fuse-dev, libxml2-dev, curl-dev
- autoconf  
autoconf is newer version, so we need to run autoupdate before running autogen.sh
- curl version  
Requires support for CURLOPTDEPRECATED warning
- etc

adamqqqplay:
This is a great question, hopefully the support on ALPINE will improve. Thanks!

ggtakec:
@adamqqqplay 
I have posted a PR code to build and test ALPINE with Github Actions.
That code and CI environment has been confirmed to work on ALPINE as well.

adamqqqplay:
> @adamqqqplay I have posted a PR code to build and test ALPINE with Github Actions. That code and CI environment has been confirmed to work on ALPINE as well.

Thanks, I will test it soon.

ggtakec:
This will be closed. If you still have problems, please reopen or post a new issue.

Issue #2046, opened 2022-10-15
AWS Alpine iam_role argument ignored when using fstab


RoyceTheBiker:
Compiled s3fs from source, version 1.91.
Mounting the S3 using the command and iam_role option works exactly as expected. Using **fstab** setting fails with an error 
```
s3fs: could not determine how to establish security credentials.
mount: mounting s3fs on /s3 failed: No such file or directory
```

### Additional Information
AWS Alpine Linux 3.15.6 running in AWS EC2 instance
"ImageId": "ami-0a31e7dfd86d40bad",
"ImageLocation": "538276064493/alpine-3.15.6-x86_64-bios-cloudinit-r0"

#### Version of s3fs being used (s3fs --version)
git clone --branch v1.91 --single-branch https://github.com/s3fs-fuse/s3fs-fuse.git

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
fuse-2.9.9-r1

#### Kernel information (uname -r)
5.15.59-0-virt

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.15.6
PRETTY_NAME="Alpine Linux v3.15"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://bugs.alpinelinux.org/"

#### s3fs command line used, if applicable
This works in Alpine and AL2
```
s3fs -o iam_role='S3_Access' ${S3_BUCKET_NAME} /s3/
```
#### /etc/fstab entry, if applicable
This works in AL2, but fails in Alpine
```
s3fs#${S3_BUCKET_NAME} /s3 fuse rw,iam_role=S3_Access 0 0
```
These also fail in Alpine
```
s3fs#${S3_BUCKET_NAME} /s3 fuse rw,iam_role=auto 0 0
s3fs#${S3_BUCKET_NAME} /s3 fuse rw,iam_role 0 0
```

#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
I have been testing with various levels of debugging. I may try to run in a gdb wrapper if no one knows a workaround.

### Details about the issue
[...]

PR #2116, opened 2023-02-19
Updated files for support ALPINE v3.17

ggtakec:
#2046 #2110

### Details
Enabled to execute Github Actions(CI) for ALPINE(v3.17).

This change includes the following changes:
#### `linux-ci-helper.sh`
Changed from `bash` to `sh`. (because the initial image of ALPINE does not include `bash`)
Created a condition for ALPINE and made modifications accordingly.

#### `autogen.sh`
ALPINE's `autoconf` is new and needs to be updated in `configure.ac`.
However, some supported older OS's don't support the new KEYWORDS in configure.ac, so I've added `autoupdated` for it.

#### `curl.cpp`
Changed `CURLOPT_PUT` to `CURLOPT_UPLOAD`.
`CURLOPT_PROGRESSFUNCTION` should be changed to `CURLOPT_XFERINFOFUNCTION`, but it was not changed because we have an OS that uses old curl.
Compiling on ALPINE will output a warning, which can be ignored.

#### integration-test-main.sh
In ALPINE, the `ENOTSUP` error is output as `Not supported` instead of `Operation not supported`, so it have been fixed.
`test_overwrite_existing_file_range` was doing file comparison with redirection, which often failed in ALPINE.
This error has been eliminated by creating a real file for this comparison, so this has been fixed.
(This fix does not affect testing.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,7 @@ jobs:
           - fedora:37
           - fedora:36
           - opensuse/leap:15
+          - alpine:3.17
 
     container:
       image: ${{ matrix.container }}
--- a/.github/workflows/linux-ci-helper.sh
+++ b/.github/workflows/linux-ci-helper.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 #
 # s3fs - FUSE-based file system backed by Amazon S3
 #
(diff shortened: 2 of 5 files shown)


ggtakec:
@gaul I will merge this, please let me know if any problems.

Issue #2090, opened 2023-01-08
Reading file is blocked if there is a dd command running on a s3fs mounted volume


creeew:
#### Version of s3fs being used (`s3fs --version`)
V1.91

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)
2.9.2

#### Kernel information (`uname -r`)
4.19.113-300.el7.x86_64

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)
CentOS7

#### How to run s3fs, if applicable
s3fs test-bucket /mnt/dir -o rw,allow_other,no_check_certificate,use_path_request_style,disable_noobj_cache,url=http://x.x.x.x:9000,passwd_file=/.passwd-s3fs,dev,suid

### Details about issue
Using s3fs command mount bucket "test-bucket" to dir "/mnt/dir". The bucket "test-bucket" has a object which name is "test.txt".
And change directory to /mnt/dir, running command `dd if=/dev/zero of=big-file bs=1024 count=102400`.
Try to read existed file "test.txt" while dd command is running.
The isssue is that "test.txt" reading operation was blocked, cannot get the file's content until dd command finished. Further more, the s3fs process memory was continuously growing.

Looks like s3fs reading operation "GetExistFdEntity" have to get the lock which was holded by "dd", maybe that's the reason why reading was blocked.
https://github.com/s3fs-fuse/s3fs-fuse/blob/222110e153248e85030ceb59501c2c73edc0f255/src/fdcache.cpp#L626-L634

ggtakec:
@creeew Thanks for this Issue post.
I just wanted to inform you about I'm going to investigate this matter soon.

ggtakec:
@creeew 
I've checked both the master branch code and the v1.91 code, just like you, but I haven't been able to reproduce the same problem.
_In each case, I tried specifying and not specifying the `use_path_request_style` option, but I could not reproduce it._

Locking in the `FdManager::GetExistFdEntity()` function is just a linear search of the `fdent_map_t variable` (I believe it is finite).
So it's hard to imagine blocking during the upload of a large file with the `dd` command.
And I'm also concerned that the used memory keeps increasing as you point.

Therefore, I have not yet determined whether this problem depends on the individual environment or is due to a flaw in the code.

If you can, could you try same test with the code in the master branch?
We have fixed some bugs since the v1.91 release, so it may show different results.
[...]

Issue #367, opened 2016-02-27
A feature request


zubryan:
Hi
I need  a feature like this:
After putting files in cache dir, asynchronously send files to S3, then I don't need to wait.
 
Is it possible? or it has already had this feature?

Thanks,
Bryan

ggtakec:
s3fs could not transfer a object asynchronously as normal, but you can specify ensure_diskfree option and it may help you.
This option means "sets MB to ensure disk free space".
So you set large value for this, s3fs probabry transfer a object asynchronously.

You can load latest codes from master branch for ensure_diskfree option.
Please try to use and set it.

Regards,

gaul:
@zubryan this is called a write-behind cache.  Generally network file systems do not implement this because there is no way to tell the application when the transfer fails, e.g., during network disconnection.

PR #2119, opened 2023-02-21
Make s3fs_flush async

huntersman:
When big file is multi uploading, s3fs will be blocked due to semaphore, we can not do anything until uploading finished. So I change `s3fs_flush` asynchronously.

I know it's not a good idea to flush asynchronously as #367 already pointed out, but if you want better performance, try it.

I am new to C++, maybe the code is not very good. 

#2090
### Details

A file is written through `s3fs_write` → `s3fs_flush` → `s3fs_release`. Use C++ thread to do part of `s3fs_flush`. Big file usually takes a lot of time to flush, so `s3fs_release` can't be called immediately, I use `doRelease` replaced it.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -25,6 +25,8 @@
 #include <dirent.h>
 #include <sys/types.h>
 #include <getopt.h>
+#include <thread>
+#include <mutex>
 
 #include "common.h"
 #include "s3fs.h"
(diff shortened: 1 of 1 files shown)


ggtakec:
@huntersman I'm sorry for my late reply.

#### About this PR codes:
When s3fs files are updated(flush), threads are used in the function called from `s3fs_flush()`.
With this fix, there are also cases where threads are called from threads.
Also, the new support for `streamupload` does partial uploads even before flush is called, and it internally starts a thread and waits.
Your code also keeps a `fuse_file_info`, but s3fs already manage  the value of `fi` in `fuse_file_info` internally, so it's not that simple.

Thus I think we should use another method(modify the existing code in some functions) to make this flush asynchronous.
So that, I don't think the code in this PR can be applied for the reasons mentioned above.

I will have taken note of your #2090 comments and will be working on it.
I appreciate your kindness.

PR #2123, opened 2023-03-06
Fixed segmentation fault caused by file write failure

garenchan:
### Details

I am currently using  s3fs with minio. When I try to write a file with a name greater than 255 in s3fs, s3fs exit with segmentation fault. Because minio uses an xfs file system that limits filename length to 255.

```
#0  0x00007f3ee1fbff4f in __strlen_sse42 () from /lib64/libc.so.6
Python Exception <type 'exceptions.ValueError'> Cannot find type const headers_t::_Rep_type:
#1  0x00000000004218a7 in get_mode (meta=std::map with 5 elements, path=path@entry=0x0, checkdir=checkdir@entry=false,
    forcedir=forcedir@entry=false) at metaheader.cpp:144
#2  0x000000000043ec11 in StatCache::UpdateMetaStats (this=<optimized out>,
    key="/host_e8b47009f661-host_diskio_used_percent_az_id_6fcabee9_f323_4904_9799_073ad3e552a3_cloud_id_6398743040_cluster_id_6398730244_device_sda1_instance_7009f65e_e8b4_b14f_eb11_5087b07a8c47_region_id_84c"..., meta=std::map with 5 elements = {...})
    at cache.cpp:452
#3  0x000000000041713a in s3fs_utimens (_path=<optimized out>, ts=<optimized out>) at s3fs.cpp:2215
#4  0x00007f3ee36cac2f in fuse_fs_utimens () from /lib64/libfuse.so.2
#5  0x00007f3ee36ccc1d in fuse_lib_setattr () from /lib64/libfuse.so.2
#6  0x00007f3ee36d142e in do_setattr () from /lib64/libfuse.so.2
#7  0x00007f3ee36d1bcb in fuse_ll_process_buf () from /lib64/libfuse.so.2
#8  0x00007f3ee36ce461 in fuse_do_work () from /lib64/libfuse.so.2
#9  0x00007f3ee224eea5 in start_thread () from /lib64/libpthread.so.0
#10 0x00007f3ee1f77b0d in clone () from /lib64/libc.so.6
```

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -449,7 +449,7 @@ bool StatCache::UpdateMetaStats(const std::string& key, headers_t& meta)
     SetStatCacheTime(ent->cache_date);
 
     // Update only mode
-    ent->stbuf.st_mode = get_mode(meta);
+    ent->stbuf.st_mode = get_mode(meta, key);
 
     return true;
 }
--- a/src/metaheader.cpp
+++ b/src/metaheader.cpp
@@ -125,7 +125,7 @@ mode_t get_mode(const char *s, int base)
     return static_cast<mode_t>(cvt_strtoofft(s, base));
 }
 
-mode_t get_mode(const headers_t& meta, const char* path, bool checkdir, bool forcedir)
+mode_t get_mode(const headers_t& meta, const std::string& strpath, bool checkdir, bool forcedir)
 {
     mode_t mode     = 0;
     bool   isS3sync = false;
@@ -141,7 +141,7 @@ mode_t get_mode(const headers_t& meta, const char* path, bool checkdir, bool for
     }else{
         // If another tool creates an object without permissions, default to owner
         // read-write and group readable.
-        mode = path[strlen(path) - 1] == '/' ? 0750 : 0640;
+        mode = (!strpath.empty() && '/' == *strpath.rbegin()) ? 0750 : 0640;
     }
 
     // Checking the bitmask, if the last 3 bits are all zero then process as a regular
@@ -163,7 +163,7 @@ mode_t get_mode(const headers_t& meta, const char* path, bool checkdir, bool for
                         if(strConType == "application/x-directory" || strConType == "httpd/unix-directory"){
                             // Nextcloud uses this MIME type for directory objects when mounting bucket as external Storage
                             mode |= S_IFDIR;
-                        }else if(path && 0 < strlen(path) && '/' == path[strlen(path) - 1]){
+                        }else if(!strpath.empty() && '/' == *strpath.rbegin()){
                             if(strConType == "binary/octet-stream" || strConType == "application/octet-stream"){
                                 mode |= S_IFDIR;
                             }else{
--- a/src/metaheader.h
+++ b/src/metaheader.h
@@ -46,7 +46,7 @@ struct timespec get_atime(const headers_t& meta, bool overcheck = true);
 off_t get_size(const char *s);
 off_t get_size(const headers_t& meta);
 mode_t get_mode(const char *s, int base = 0);
-mode_t get_mode(const headers_t& meta, const char* path = NULL, bool checkdir = false, bool forcedir = false);
+mode_t get_mode(const headers_t& meta, const std::string& strpath, bool checkdir = false, bool forcedir = false);
 uid_t get_uid(const char *s);
 uid_t get_uid(const headers_t& meta);
 gid_t get_gid(const char *s);


src/cache.cpp
```diff
@@ -449,7 +449,7 @@ bool StatCache::UpdateMetaStats(const std::string& key, headers_t& meta)
     SetStatCacheTime(ent->cache_date);
 
     // Update only mode
-    ent->stbuf.st_mode = get_mode(meta);
+    ent->stbuf.st_mode = get_mode(meta, key.c_str());
```
gaul, line 452: Could you remove the default value for `path` from `metaheader.h`?

src/metaheader.cpp
```diff
@@ -138,7 +138,7 @@ mode_t get_mode(const headers_t& meta, const char* path, bool checkdir, bool for
         isS3sync = true;
     }else if(meta.end() != (iter = meta.find("x-amz-meta-goog-reserved-posix-mode"))){ // for GCS
         mode = get_mode((*iter).second.c_str(), 8);
-    }else{
+    }else if(path != NULL){
```
gaul, line 141: If you remove the default value then this conditional isn't necessary.

gaul (changes requested):

gaul:
Could you explain the file write failure more fully?  We should add a test for overlong file names.  While these should not be allowed to be created via s3fs due to `NAME_MAX`, it is possible that existing objects in S3 have longer names and we should at least make sure we don't crash via a test.

src/metaheader.cpp
gaul: Let's at least remove the default argument.  These tend to have magical behaviors.  If we want to make the cost maximally defensive we should also consider when `strlen` returns 0.

ggtakec: OK, we should not use default arguments for this function.

@garenchan 
I would like to merge the sources in this PR, so could you edit the part where I commented and update the code?

ggtakec: Could you fix the following 3 files (including your modified part)?

- src/cache.cpp
```
diff --git a/src/cache.cpp b/src/cache.cpp
index 593760a..d538a1e 100644
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -449,7 +449,7 @@ bool StatCache::UpdateMetaStats(const std::string& key, headers_t& meta)
     SetStatCacheTime(ent->cache_date);

     // Update only mode
-    ent->stbuf.st_mode = get_mode(meta);
+    ent->stbuf.st_mode = get_mode(meta, key);

     return true;
 }
```

- src/metaheader.cpp
```
diff --git a/src/metaheader.cpp b/src/metaheader.cpp
index d0c28b5..fc6f630 100644
--- a/src/metaheader.cpp
+++ b/src/metaheader.cpp
@@ -125,7 +125,7 @@ mode_t get_mode(const char *s, int base)
     return static_cast<mode_t>(cvt_strtoofft(s, base));
 }

-mode_t get_mode(const headers_t& meta, const char* path, bool checkdir, bool forcedir)
+mode_t get_mode(const headers_t& meta, const std::string& strpath, bool checkdir, bool forcedir)
 {
     mode_t mode     = 0;
     bool   isS3sync = false;
@@ -141,7 +141,7 @@ mode_t get_mode(const headers_t& meta, const char* path, bool checkdir, bool for
     }else{
         // If another tool creates an object without permissions, default to owner
         // read-write and group readable.
-        mode = path[strlen(path) - 1] == '/' ? 0750 : 0640;
+        mode = (!strpath.empty() && '/' == *strpath.rbegin()) ? 0750 : 0640;
     }

     // Checking the bitmask, if the last 3 bits are all zero then process as a regular
@@ -163,7 +163,7 @@ mode_t get_mode(const headers_t& meta, const char* path, bool checkdir, bool for
                         if(strConType == "application/x-directory" || strConType == "httpd/unix-directory"){
                             // Nextcloud uses this MIME type for directory objects when mounting bucket as external Storage
                             mode |= S_IFDIR;
-                        }else if(path && 0 < strlen(path) && '/' == path[strlen(path) - 1]){
+                        }else if(!strpath.empty() && '/' == *strpath.rbegin()){
                             if(strConType == "binary/octet-stream" || strConType == "application/octet-stream"){
                                 mode |= S_IFDIR;
                             }else{
```

- src/metaheader.h
```
diff --git a/src/metaheader.h b/src/metaheader.h
index 5f78a42..44dd8a5 100644
--- a/src/metaheader.h
+++ b/src/metaheader.h
@@ -46,7 +46,7 @@ struct timespec get_atime(const headers_t& meta, bool overcheck = true);
 off_t get_size(const char *s);
 off_t get_size(const headers_t& meta);
 mode_t get_mode(const char *s, int base = 0);
-mode_t get_mode(const headers_t& meta, const char* path = NULL, bool checkdir = false, bool forcedir = false);
+mode_t get_mode(const headers_t& meta, const std::string& strpath, bool checkdir = false, bool forcedir = false);
 uid_t get_uid(const char *s);
 uid_t get_uid(const headers_t& meta);
 gid_t get_gid(const char *s);
```

ggtakec:
@gaul 
I try to implement a test case for file names (object names) that exceed `NAME_MAX`.
(I think that test case is separated from this PR and apply in another PR.)

ggtakec:
@ garenchan
Thanks for updating your code, I will merge this PR .

ggtakec (approved):

Issue #2120, opened 2023-02-22
File content lost after write failure


adamqqqplay:
When the user's IAM role does not have permission to write to the S3 Bucket, the file will lose its original data after overwriting.

#### Version of s3fs being used (`s3fs --version`)
V1.91 (commit:f8a825e)

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)
2.9.9

#### Kernel information (`uname -r`)
5.15.0-43-generic

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)
PRETTY_NAME="Ubuntu 22.04.1 LTS"

#### How to run s3fs, if applicable
command line  

```
sudo s3fs elune-s3-test -o check_cache_dir_exist -o use_cache=~/test/s3fs/cache -o passwd_file=~/test/s3fs/s3key.txt -o allow_other  -o auto_unmount -o entry_timeout=100000 -o attr_timeout=10000 -o dbglevel=info -f  ./mnt
```

#### s3fs syslog messages (`grep s3fs /var/log/syslog`, `journalctl | grep s3fs`, or `s3fs outputs`)

```
2023-02-22T07:40:19.093Z [INF] s3fs.cpp:s3fs_open(2650): [path=/test.txt][flags=0x8000]
2023-02-22T07:40:19.093Z [INF]       fdcache.cpp:GetFdEntity(484): [path=/test.txt][pseudo_fd=-1]
2023-02-22T07:40:19.093Z [INF]       cache.cpp:DelStat(596): delete stat cache entry[path=/test.txt]
2023-02-22T07:40:19.093Z [INF]       curl.cpp:HeadRequest(3295): [tpath=/test.txt]
2023-02-22T07:40:19.093Z [INF]       curl.cpp:PreHeadRequest(3255): [tpath=/test.txt][bpath=][save=][sseckeypos=18446744073709551615]
2023-02-22T07:40:19.093Z [INF]       curl_util.cpp:prepare_url(257): URL is https://s3-ap-southeast-1.amazonaws.com/elune-s3-test/test.txt
2023-02-22T07:40:19.093Z [INF]       curl_util.cpp:prepare_url(290): URL changed is https://elune-s3-test.s3-ap-southeast-1.amazonaws.com/test.txt
[...]

PR #2127, opened 2023-03-12
Added cache clearing on upload error

ggtakec:
#2120

### Details
After an error occurred on the server side during file upload(Flush) processing, there are reports(#2120) that the contents of the file cannot be read correctly.

If the upload fails in an environment that uses cache, the local cache may have been updated even though the object on the server side has not been updated.
Even in this state, subsequent reading may be performed normally (maybe reloading due to inconsistency in stat information), but Issue #2120 seems to cause inconsistency.

If the upload fails, the local cache is in a correct state (not matching the server side) and cannot be repaired.
Therefore, I modified it to clear the cache if it fails.

@gaul 
Since this PR only clears the cache, I think it won't impact anything else.
So, if the tests pass, I think I'll merge this PR in no time(probably after issuer's review).
If you have any problems please let me know.

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -1446,6 +1446,14 @@ int FdEntity::RowFlush(int fd, const char* tpath, AutoLock::Type type, bool forc
         result = RowFlushMultipart(pseudo_obj, tpath);
     }
 
+    // [NOTE]
+    // if something went wrong, so if you are using a cache file,
+    // the cache file may not be correct. So delete cache files.
+    //
+    if(0 != result && !cachepath.empty()){
+        FdManager::DeleteCacheFile(tpath);
+    }
+
     return result;
 }
 


adamqqqplay:
I tested this PR and it fixed our previous issue.

Regarding the cause of the error of this issue, my analysis is written [here](https://github.com/adamqqqplay/s3fs-fuse/commits/fix-cache-inconsistency):

> If the user's IAM role does not have permission to write to the S3
Bucket, the original data will be lost after the file is overwritten.
> 
> This is because overwrite writing will be divided into two steps:
truncation and writing. First, truncating the file will succeed, but
uploading will fail, so subsequent writing will not happen. But at this
time, the timestamp of the local file is newer than that on the server,
so S3FS will think that the local data is correct when it is read next
time, but the old data on the server is actually correct.
> 
> This patch will clear the timestamp of the local cache to invalidate the
cache after the upload fails, and force S3FS to re-download the data at
the next read to fix this problem.

liubogithub:
Hi, Is is possible to use pagelist to track which part is stale so that we can fix it in a fine-grained way?
@ggtakec

ggtakec:
@adamqqqplay 
Updating mtime/ctime(to 0) can complicate matters, although I think it's a very rare case.
If a different file descriptor is open than the failed write(flush), it may be overwritten with its ctime/ctime.
So I believe removing caches that may not be consistent with server-side data is a way to avoid complicating matters.

Thanks to your confirmation, I will merge this PR.

ggtakec:
@liubogithub 
> Hi, Is is possible to use pagelist to track which part is stale so that we can fix it in a fine-grained way?

I wish I could do this too, but I'm still worried.
_If it is not streamupload, it may be possible by unloading the part marked as edited in pagelist._
_However, in the case of streamupload, I'm not sure if that simple process is enough._

This correspondence is originally a workaround for the upload error, so I'm going to merge this PR code for now.
_We don't want that cost to be wasted by deleting the cache of large files that have already been loaded, but..._

adamqqqplay:
@ggtakec Thanks for everything!

Issue #2128, opened 2023-03-14
Download and delete a file at the same time will eventually not release the cache space


huntersman:
#### Version of s3fs being used (`s3fs --version`)

V1.91
#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)

2.9.2
#### Kernel information (`uname -r`)

5.4.213-1.el7.elrepo.x86_64
#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)

CentOS 7
#### How to run s3fs, if applicable

`s3fs demo /root/demo -o passwd_file=${HOME}/.passwd-s3fs -o url=http://ip -o use_path_request_style -o noxmlns -o dbglevel=error -o default_acl=public-read -o logfile=/var/log/s3fs.log -o allow_other -o multireq_max=500 -o nocopyapi -o use_cache="/root/s3fs"`
### Details about issue

1. Use a SFTP tool to download a file from s3fs (For example, a 2GB file).
2. When it's downloading, delete the file. `rm -f /root/demo/2GB`
3. `.fuse_hidden` file is generated by Linux.
4. After downloading finished, `.fuse_hidden` is deleted automatically, but the cache space is still 2 GB.

ggtakec:
@huntersman Thanks for PR.
But I would like to know more information about `"the cache space is still 2 GB."`
Does this mean there are 2GB files left under the `/root/s3fs` directory?
There is no direct relationship between the `.fuse_hidden` file and s3fs' own cache file, so I'm wondering.

huntersman:
There is no 2GB file in cache directory, neither in `.demo.stat` and `.demo.mirror`. The result of `df -h` shows that the space is not freed. And I notice that there is an error in the log.
```shell
[ERR] s3fs.cpp:s3fs_release(2963):could not find pseudo_fd(3) for path(/.fuse_hidden0000000500000002)
```
After I kill s3fs, the space will be freed.
```shell
kill -9 s3fsPid
```

ggtakec:
@huntersman 
I understood the cause of this problem.

FUSE temporarily changes the file name to `/.fuse_hiddenXXXYYY` if the open file is deleted by another process etc.
When the open file is closed, this temporarily renamed file is also deleted.

Above case, s3fs did not rename the cache for large files, but deleted them.(but, renamed except for large files)
This may have been leaked in past fixes.

[...]

PR #2130, opened 2023-03-15
Fix an issue that cache space will not be freed

huntersman:

#2128 
### Details

I think cache space will not be freed is because `s3fs_unlink` already delete the cache file, but `.fuse_hidden` file is generated which is also in the cache space. So I use a set to record file which is downloading, and only remove cache file when file is not in set.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -49,7 +49,7 @@
 #include "mpu_util.h"
 #include "threadpoolman.h"
 #include "autolock.h"
-
+#include <set>
 //-------------------------------------------------------------------
 // Symbols
 //-------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/s3fs.cpp
```diff
@@ -1166,9 +1166,12 @@ static int s3fs_unlink(const char* _path)
     }
     S3fsCurl s3fscurl;
     result = s3fscurl.DeleteRequest(path);
-    StatCache::getStatCacheData()->DelStat(path);
-    StatCache::getStatCacheData()->DelSymlink(path);
-    FdManager::DeleteCacheFile(path);
+    // delete stat cache when it's not downloading
+    if(fileReading.find(path) == fileReading.end()){
```
ggtakec, line 1170: When are cache files (s3fs' own) deleted?
It seems that there is no timing to delete unless it is executed in this part.

ggtakec:
@huntersman 
With your help, I was able to identify the cause of #2128.
The way to fix it is much simpler than this PR.
I created #2135 for that fix.
This PR will be closed and the fix for #2128 will be addressed with #2135.

Issue #2107, opened 2023-02-02
s3fs adds characters to file 


mabi08:
#### Version of s3fs being used (`s3fs --version`)
Amazon Simple Storage Service File System V1.91 (commit:unknown) with OpenSSL

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)
Name        : fuse
Version     : 2.9.2
Release     : 11.amzn2
Architecture: x86_64
Install Date: Thu 26 Jan 2023 10:49:48 PM UTC
Group       : System Environment/Base
Size        : 222809
License     : GPL+
Signature   : RSA/SHA256, Thu 06 Dec 2018 07:31:53 PM UTC, Key ID 11cf1f95c87f5b1a
Source RPM  : fuse-2.9.2-11.amzn2.src.rpm
Build Date  : Fri 16 Nov 2018 08:35:39 PM UTC
Build Host  : build.amazon.com
Relocations : (not relocatable)
Packager    : Amazon Linux
Vendor      : Amazon Linux
URL         : https://github.com/libfuse/libfuse
Summary     : File System in Userspace (FUSE) utilities

#### Kernel information (`uname -r`)
4.14.301-224.520.amzn2.x86_64

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)
NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"

#### How to run s3fs, if applicable

[] command line  
[x] /etc/fstab  

### Details about issue

Hi,

we are using s3fs to mount s3 buckets on AWS appliances (Amazon Linux 2). 
However, we are currently facing a bug where removing characters from a file seems to add weird characters. 
In Jupyterlab editor those are shown as red dots in the file, afterwards the file is not usable anymore. This is reproducible also with other type of files. We tried building s3fs from source but this does not fix the issue for us. I attached a gif showing the problem.

Kind regards
Maik

![s3fs_issue](https://user-images.githubusercontent.com/50304545/216338953-2722b72d-f1bd-4ffc-a836-a4959e257080.gif)

ggtakec:
@mabi08 
[...]

PR #2131, opened 2023-03-16
Flush after opened file size is shrinked

ggtakec:
#2107 

### Details
There was a bug in trauncate, which shrinks files.

This bug occurs under the following conditions:
- Open the file and shrink the file.
- After shrinking, do read before flushing(do not close or sync).

At this time, it got the old size before shrinking from the server(Content-Length) and set old size again.
So that the read data from the reduced size to the original size padded as 0x00, and read it.
When read all of file, the read size is old length before reduction.

I added C source code to test the condition that caused the failure and included it in this PR.
(Because it is difficult to realize with a shell script, I wrote it in C source)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ config.status
 config.sub
 configure
 configure.scan
+configure.ac~
 depcomp
 install-sh
 libtool
@@ -86,6 +87,7 @@ test/junk_data
 test/s3proxy-*
 test/write_multiblock
 test/mknod_test
+test/truncate_read_file
 
 #
 # Windows ports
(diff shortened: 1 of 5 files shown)


mabi08:
Thanks 4 your effort! However the issue still seems to exist.

These are the installation steps I am using for installation:
```sh sudo yum install -y git automake gcc-c++ {fuse,libxml2,openssl}-devel fuse
cd /tmp
sudo git clone https://github.com/s3fs-fuse/s3fs-fuse
cd s3fs-fuse/
sudo git fetch origin pull/2131/head:fix_filesize
sudo git checkout fix_filesize
sudo ./autogen.sh
sudo ./configure
sudo make
sudo make install
```

s3fs --version shows the following:
image

ggtakec:
I was unable to identify the cause of the problem that was not fixed in your environment.
However, instead of flushing immediately after truncate by this PR(a slightly easier fix), I have found a way to avoid the problem with the current logic.
I posted new PR#2133 that adapted them, so I will close this PR and work on another PR.

@mabi08 
I'm sorry, but it would be helpful if you could double check with the code in PR#2133.

@gaul
This PR will be closed and switched to #2133.

PR #2133, opened 2023-03-18
Fixed a bug when reading a reduced file without flushing

ggtakec:
#2107 #2131
_This PR changed the fix method because #2131 was not enough fix._

### Bug details
There was a bug in trauncate, which shrinks files.

This bug occurs under the following conditions:
- Open the file and shrink the file.
- After shrinking, do read before flushing(do not close or sync).

At this time, it got the old size before shrinking from the server(Content-Length) and set old size again.
So that the read data from the reduced size to the original size padded as 0x00, and read it.
When read all of file, the read size is old length before reduction.

### Fixes
Changed so that when opening a file that has already been opened and resized(without flushing), it is opened as the modified file size instead of the file size on the server side.

When reading(loading) a file, there was a problem that the modifying(shrink) flag was cleared after reading even though the file was reduced.
Therefore, there was a problem that flush(upload) of file size change did not occur, so it has been fixed.

Although it is not directly related to this case, there was a bug in the number of blocks used in the stat information, so it has been corrected.

I added C source code to test the condition that caused the failure and included it in this PR.
(Because it is difficult to realize with a shell script, I wrote it in C source)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ config.status
 config.sub
 configure
 configure.scan
+configure.ac~
 depcomp
 install-sh
 libtool
@@ -86,6 +87,7 @@ test/junk_data
 test/s3proxy-*
 test/write_multiblock
 test/mknod_test
+test/truncate_read_file
 
 #
 # Windows ports
--- a/src/fdcache_page.cpp
+++ b/src/fdcache_page.cpp
@@ -456,7 +456,13 @@ bool PageList::Resize(off_t size, bool is_loaded, bool is_modified)
     off_t total = Size();
 
     if(0 == total){
+        // [NOTE]
+        // The is_shrink flag remains unchanged in this function.
+        //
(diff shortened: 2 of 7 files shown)


ggtakec:
@mabi08 
I installed `Jupyter Notebook` on `Ubuntu 22.04` and inspected each s3fs upload type(`stermupload` / `multipartupload`) and with/without cache-derectory.
And Inspection performed file editing from the browser.

However, I have not yet been able to reproduce the bug with the code in this PR.(and I haven't been able to reproduce the same phenomenon with `v1.91`.)
The bug that caused this PR to fix was found without using `Jupyter Notebook`.
So bug #2107 may have a different cause than the fix in this PR.

I would like to merge this PR, not close Issue #2107, and reproduce the issue and identify the cause.
(I will also post this same comment on #2107)

ggtakec:
@gaul I'll merge this PR, and if you find any problem, please let me know.

PR #2134, opened 2023-03-18
Fixed configure.ac by autoupdate

ggtakec:
#2116 

### Details
We added `autoupdate` command to `autogen.sh` by #2116.
`autoupdate` will update `configure.ac` at build time if needed, but currently it always updates.
Therefore, this PR updates `configure.ac` as same as updating it by `autoupdate`.


ggtakec:
@gaul Please let me know if you have any problems.

Issue #2067, opened 2022-11-25
ls, permission denied for filenames containing CR (0x13) character


jalmasi:
Did you know that linux file names can contain CRLF (0x13) character? :>
I guess everybody knows about special characters, but 0x13 is even more special. If you manage to create a file containing 0x13, you're not going to see it in the file name. You may see it as <0x13>, ^S or DC3 in editors/viewers if you redirect ls to file, but linux console will just not render it.

Then you mount a bucket with s3fs, copy that file over to a directory there, issue ls, and get permission denied. Everything works, except that one directory containing the evil file(s).
But you can list the directory content using awscli, permission denied will not be there. And neither will 0x13, because it's so very special.
S3 UI for the bucket also does not display the character, and does not allow file removal. Yet it does allow to remove the character, everything works afterwards. You just need to know where it is. Only just :)

This isn't exactly a bug report. I don't expect anybody to do anything about this, I'm just trying to save time and nerves of the next guy that runs into it.
The above provides enough info to diagnose and reproduce the issue. Fix is obvious, renaming the file. 
Hard part is matching 0x13. I had a small java program written for that, but language doesn't matter. Couldn't use regex, because, like, what's the escape sequence? :) I guess utf escape sequences might do.
Well, HTH :>

simonfelding:
@jalmasi the regex way is to match 0x0D (it's character 13 in decimal, 0D in hexadecimal!). \r should do it too :)

ggtakec:
I did a quick check.
I created objects with filenames containing `\n` and `\r` in my S3 bucket and referenced them by mounting them with s3fs.
And they were the same as the result displayed by `ls` in the file Linux created on the local disk.
Both are displayed as `?`, such as `SAMPLE?`.
Inside s3fs, these special characters appear to be processed exactly as received from the S3 server.

@jalmasi 
What's wrong with s3fs compared to normal Linux commands?

jalmasi:
@ggtakec simple "find ." on the mounted bucket would end up with permission denied. But note that these weren't bucket objects created in the bucket, these were files on ext4 filesystem that contained special characters. Then the bucket was mounted, and files copied over.

ggtakec:
@jalmasi 
I found the cause of this bug and posted a PR #2136 to fix it.

The cause was that the character code was converted by libxml, `CR` (`0x0D`) characters were converted to `LF`(`0x0A`), and the file could not be detected.

If you can build this modified code, please try to check it.
[...]

PR #2136, opened 2023-03-21
Fixed a bug in handling file names containing CR(0x0D)

ggtakec:
#2067 

### Details
libxml2 converts `CR` code(`'\r'`=`0x0D`) to `LF`(`'\n'`=`0x0A`) according to the XML specification.

Therefore, if the object name(file name) contains `CR`, `ListBucket` will try to process the object name converted to `LF`.
And as a result, there was a problem that the user could not get the file list.

The fix is to escape(`%0D`) the `CR` code from `ListBucket` results before libxml2's XML parsing and restore it after parsing.
_The encoding uses the same technique as URL encoding._
Within XML content, HTML encoding(`&#13;`) is appropriate, but in this case escaped and unescaped strings are indistinguishable, so it was not possible.

In this fix, we added a unit test(in `test_string_util.cpp`) for the string manipulation function(`test_cr_encoding`).
Also added `cr_filename.c` for testing filenames containing `CR` code. (It was added with C code because it was difficult to create a `CR` code file name from the shell script)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,6 +88,7 @@ test/s3proxy-*
 test/write_multiblock
 test/mknod_test
 test/truncate_read_file
+test/cr_filename
 
 #
 # Windows ports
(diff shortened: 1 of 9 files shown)


ggtakec:
@gaul 
We've confirmed it, so I'm merging this.
Please let me know if you find any problems.

Issue #2111, opened 2023-02-14
Size of mounted filesystem is 16 EiB


balducciatix:
#### Version of s3fs being used (`s3fs --version`)

1.91
#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)

2.9.2
#### Minio version
RELEASE.2023-01-25T00-19-54Z

#### Kernel information (`uname -r`)

3.10.0-1160.81.1.el7.x86_64

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)

CentOS 7 (host on which I mount the bucket)
AlmaLinux 8.7 (host running minio server)

#### How to run s3fs, if applicable

command line:
```
s3fs my-second-bucket /mnt/bucket -f -o passwd_file=${HOME}/.passwd-s3fs -o url=http://10.0.5.17:9000/ -o use_path_request_style -o dbglevel=info -f -o curldbg
```  

#### s3fs syslog messages (`grep s3fs /var/log/syslog`, `journalctl | grep s3fs`, or `s3fs outputs`)

```
2023-02-13T16:21:33.957Z [INF] s3fs version 1.91(e715b77) : s3fs -f -o passwd_file=/root/.passwd-s3fs -o url=http://10.0.5.17:9000 -o use_path_request_style -o dbglevel=info -f -o curldbg my-second-bucket /mnt/bucket
2023-02-13T16:21:33.960Z [CRT] s3fs_logger.cpp:LowSetLogLevel(239): change debug level from [CRT] to [INF]
2023-02-13T16:21:33.960Z [INF]     s3fs.cpp:set_mountpoint_attribute(4372): PROC(uid=0, gid=0) - MountPoint(uid=993, gid=991, mode=40777)
2023-02-13T16:21:33.962Z [INF] curl.cpp:InitMimeType(431): Loaded mime information from /etc/mime.types
2023-02-13T16:21:33.962Z [INF] fdcache_stat.cpp:CheckCacheFileStatTopDir(78): The path to cache top dir is empty, thus not need to check permission.
2023-02-13T16:21:33.962Z [INF] threadpoolman.cpp:StopThreads(195): Any threads are running now, then nothing to do.
2023-02-13T16:21:33.962Z [INF]       threadpoolman.cpp:Worker(76): Start worker thread in ThreadPoolMan.
[...]

PR #2148, opened 2023-04-10
New option: bucket_size

OttaviaB:
### Relevant Issue

#2111
### Details

This pull request adds the new `bucket_size` option, which allows to set the size of the mounted bucket advertised by s3fs to a user-defined value.

While the previous behaviour is preserved as the default, this new option can be used when accurate metrics are needed.


src/s3fs.cpp
```diff
@@ -2808,15 +2810,15 @@ static int s3fs_statfs(const char* _path, struct statvfs* stbuf)
...
+    stbuf->f_blocks = bucket_size;
+    stbuf->f_bfree = stbuf->f_blocks;
 #elif defined(__APPLE__)
-    stbuf->f_blocks = UINT32_MAX;
-    stbuf->f_bfree  = UINT32_MAX;
-    stbuf->f_files  = UINT32_MAX;
-    stbuf->f_ffree  = UINT32_MAX;
+    stbuf->f_blocks = bucket_size;
+    stbuf->f_bfree  = stbuf->f_blocks;
+    stbuf->f_bfree  = stbuf->f_blocks;
```
ggtakec, line 2818: Isn't this a mistake in `stbuf->f_files`?

ggtakec (changes requested):
Please check comments

ggtakec:
@OttaviaB Thank you for your kindness!
I agree with this amendment, but it seems that one correction is required.

src/s3fs.cpp
OttaviaB: Yes, definitely. I have just pushed a new commit where f_files and f_ffree are left defined as UINT32MAX as before.

ggtakec:
@OttaviaB Thanks for your contribution.
I'll merge this PR soon.

Issue #2151, opened 2023-04-19
Issue mounting with prefix and federation token


nchaly:
#### Version of s3fs being used (`s3fs --version`)

Amazon Simple Storage Service File System V1.91 (commit:c4f95f1) with OpenSSL

(manually built from master).

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)

2.9.9

#### Kernel information (`uname -r`)

5.19.0-1022-aws

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)

PRETTY_NAME="Ubuntu 22.04.2 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.2 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy

### Details about issue

Scenario: one of our services issues to clients federation tokens to access specified prefix ("folder"). Credentials consists of key id, key, session token, and limit access only to specified prefix for certain amount of time. 

Client mounts specified prefix with the command:

```
s3fs -o ro -o endpoint=eu-west-1 -o dbglevel=info -o use_session_token -o use_cache=/tmp/s3fs -o del_cache -o curldbg ${BUCKET}:${PREFIX} ${MNT_PATH}
```

Credentials are passed via environment. Result is that bucket is not mounted. Checking s3fs logfiles shows the following sequence of requests (with retry added in https://github.com/s3fs-fuse/s3fs-fuse/pull/2087 ):

- request GET /
- response 403
- request GET ${PREFIX}
- response 404

(detailed logs below).

Note that using the same credentials, the `aws s3 ls ${BUCKET}${PREFIX}` returns list of files correctly. 

So the question is - is there something wrong with my scenario or permissions?
Otherwise - does this second retry with prefix work as expected? Because it does not match  S3 api methods:

- GetObject request expects valid key (i.e. full path to a file) - https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html (and naturally 404 is normal).
- ListObjects request should pass prefix as a parameter: https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html (and that's why aws cli works).

s3fs logs  (grep s3fs /var/log/syslog) (shortened):

```
init v1.91(commit:c4f95f1) with OpenSSL, credential-library(built-in)
check services.
      check a bucket.
      
> GET / HTTP/1.1
[...]

PR #2153, opened 2023-05-04
Allow mount points without directory objects by compat_dir

ggtakec:
#2151
(Related: #2063 #1728 #1687 #1460 )

### Details
Checking mount point still had issue #2151 after a series of fixes.
(It's almost identical to the issue mentioned in #1460.)

I would like to support the following cases:

User specifies a directory created by a client other than s3fs as the mount point.
(It means there is no object as this directory, and there is a file object etc under that path.)
And if user has no permission to the bucket and has permission to this mount point directory, s3fs does not start with an error.

This PR is code to support the above case.

There is already a `compat_dir` option to support directories created by clients other than s3fs and directories with no existing objects.
In the above case(no directory object for mount point), this PR code make s3fs start without an error by specifying the `compat_dir` option enabled.
This means consistent with the definition and intent of the `compat_dir` option.

@nchaly
I referred to your https://github.com/nchaly/s3fs-fuse/commit/5670dd9dab9962e55ab1d3457ddf8025fecc7653 code.
Thanks for your great help!

Testing of this PR code is difficult to automate, so I checked it manually.

### DIFF
--- a/doc/man/s3fs.1.in
+++ b/doc/man/s3fs.1.in
@@ -372,6 +372,8 @@ s3fs supports two different naming schemas "dir/" and "dir" to map directory nam
 S3fs uses only the first schema "dir/" to create S3 objects for directories.
 .TP
 The support for these different naming schemas causes an increased communication effort.
+.TP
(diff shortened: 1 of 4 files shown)


nchaly:
Thanks - it work for me. `s3fs/1.91 (commit hash b096c6e; OpenSSL)`

ggtakec:
@nchaly Thank you for confirming, thus this PR will be merged soon.

PR #2154, opened 2023-05-08
Disabled to run github action for macos

ggtakec:
#1769

### Details
As of May 2023, it is no longer possible to launch macOS 10.15 as a Runner for Github Actions.
After this, it is necessary to use macOS 11 or later, but Github Actions cannot be executed because the OS reboot is required to install macFUSE.
So I disabled Github Actions execution on macOS.

We will resume this once we are able to work around the macFUSE installation issue on macOS 11 and above.

@gaul 
I will merge this PR soon.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,13 +115,17 @@ jobs:
           make ALL_TESTS=1 check -C test || (test/filter-suite-log.sh test/test-suite.log; exit 1)
 
   # [NOTE]
-  # A case of "runs-on: macos-11.0" does not work,
-  # because load_osxfuse returns exit code = 1.
-  # Maybe it needs to reboot. Apple said 
-  # "Installing a new kernel extension requires signing in as an Admin user. You must also restart your Mac to load the extension".
-  # Then we do not use macos 11 on GitHub Actions now.
+  # This Job does not work for macOS 11 and later because load_osxfuse returns exit code = 1.
+  # Apple states "You must be signed in as an administrator user to install new kernel
+  # extensions, and your Mac must be rebooted for the extensions to load.", so it needs
+  # to reboot OS.
+  # As of May 2023, GitHub Actions are no longer able to launch macos 10.15 as runner,
+  # so we can not run this Job.
+  # In the future, if it is found a solution, we will resume this Job execution.
   #
   macos10:
+    if: false
+
     runs-on: macos-10.15
 
     steps:


adamqqqplay:
@ggtakec Hi, there are two very similar commits on the current master branch, I don't know if this is a github bug? The first commit is the change of this PR, but the second commit is empty.
image

Perhaps doing a `git reset` on the master branch will fix this.

ggtakec:
@adamqqqplay Thanks for letting me know.
I think it's not a github bug, maybe I did something wrong.(I still don't know why)
I understand that `git reset` can fix that, but since it has already been reflected to the master branch, I think it should not be executed. I'm worried about impacting other forked repositories, so I'll leave it as is.
(If it's my personal repository, I'll do it, but this s3fs-fuse repository has a lot of referrers, so I think the change will cause confusion.)

adamqqqplay:
@ggtakec You're welcome, from the output on this page, there are indeed some problems here. What is merged is an empty commit, followed by a commit with actual changes that is merged into master after the branch is deleted.
image

ggtakec:
@adamqqqplay It's not good if the reason isn't clear, so I contacted Github support to see if it was a bug or not.
So we will wait for his reply.

ggtakec:
@adamqqqplay 
The reason why this commit occurred is not yet clear, but it seems that it occurred in the following conditions.
- merge
- delete personal branch
- **the page before the merge was redisplayed**
- click the merge button **again**

I have also contacted Github Support about this matter. They will get back to me if they have any new information on this matter.
(Why was the page redisplayed, and why was the meaningless re-merge performed?)

That's all I can do, but I'll try to be more careful in the future.

adamqqqplay:
> @adamqqqplay The reason why this commit occurred is not yet clear, but it seems that it occurred in the following conditions.
> 
> * merge
> * delete personal branch
> * **the page before the merge was redisplayed**
> * click the merge button **again**
> 
> I have also contacted Github Support about this matter. They will get back to me if they have any new information on this matter. (Why was the page redisplayed, and why was the meaningless re-merge performed?)
> 
> That's all I can do, but I'll try to be more careful in the future.

@ggtakec Thanks for the detailed description, hopefully this sort of thing will be fixed by github in the future.

ggtakec:
@adamqqqplay 
I got a reply from Github.

The problem seems to be that after doing the first PR merge the merge was successful, but a timeout occurred at the time when the backend marked it as PR done.
Therefore, it was a phenomenon caused by me again instructing the merge process from the Github Web.
They are working on a fix, so I'm looking forward to that.
I believe that this phenomenon is not very frequent, so it is unlikely that it will recur.

That's all we can do, now we just have to wait for them to fix it.

adamqqqplay:
> @adamqqqplay I got a reply from Github.
> 
> The problem seems to be that after doing the first PR merge the merge was successful, but a timeout occurred at the time when the backend marked it as PR done. Therefore, it was a phenomenon caused by me again instructing the merge process from the Github Web. They are working on a fix, so I'm looking forward to that. I believe that this phenomenon is not very frequent, so it is unlikely that it will recur.
> 
> That's all we can do, now we just have to wait for them to fix it.

Thanks for your reply, it looks like this is a small bug caused by network.

Now that github has discovered the problem, the probability of this kind of thing happening in the future is still relatively small.

PR #2155, opened 2023-05-08
Directly and simplify requests in mount point checks

ggtakec:
#2153
(Related: #2151 #2063 #1728 #1687 #1460 )

#### Simplified mount point checking
If user specifies a directory path for the mount point and user doesn't have access to the bucket top, the check is done in two steps(the bucket top and the directory).
(This is that @adamqqqplay pointed out by https://github.com/s3fs-fuse/s3fs-fuse/pull/2087#issuecomment-1537834762.)

Instead of this unnecessary two-step check, only the directory is checked when a directory path is specified.

I also reviewed and simplified the logic of the modified functions accordingly.

#### Fix output messages in checking mount point
Changed to output the recommended option when an error(warning) occurs in the mount point check.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -4267,23 +4267,13 @@ static bool check_endpoint_error(const char* pbody, size_t len, std::string& exp
 //     (2A) Directories created by clients other than s3fs
 //     (2B) Directory created by s3fs
 //
-// At first this function checks for access to the bucket and returns success
-// if the user has access to the bucket.
-// This first check will fail if the user does not have access to the bucket.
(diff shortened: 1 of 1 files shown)


adamqqqplay:
Thanks for your quick fix! I have tested this PR, it works for me. 
After this PR, the startup time of S3FS when mounting bucket subdirectories should be further shortened.

ggtakec:
@adamqqqplay Thanks for confirming.
I'll wait a bit longer before merging this PR, please wait for a while.

ggtakec:
@gaul Imerge this, if you have any problem. please post issues.

Issue #2156, opened 2023-05-11
[feature request] request to support write file use O_DIRECT flags


tanguofu:
when write the big file for example  the file is tens of gigabytes in size. the active_file cache in memory of system because very big， which will be trigger the oom when the s3fs run in a pod which has  poor mem limit.

so it is possible to add the options  make s3fs write use O_DIRECT flags to reduce the system memory cache for write  large file

many thanks！

ggtakec:
@tanguofu 
Could you tell us what version of s3fs you are using and the contents of the command line(or fstab entry) when starting s3fs?
I'm wondering why the s3fs process gets tens of gigabytes in a single object download.
(If this is occurred, I believe it can be avoided with an option, so please tell me how to start it.)
Thanks in advance for your assistance.

gaul:
If s3fs has unbounded memory then this is something we should fix.  This has nothing to do with `O_DIRECT` which limits the kernel page cache.

tanguofu:
the memory is used by  page cache of system memory， so i add the fdatasync to flush cache fix this.

ggtakec:
@tanguofu 
_(Please let us continue the discussion of comment #2157 in this issue.)_

The problem is that when running s3fs in a container(kubernetes/docker), downloading files of tens of gigabytes will increase the ative_file cache(page cache), which will hit the OOM threshold, isn't it?

And since it is in a Container, the active_file cache increases within the free memory area of host/node, and it becomes over the OOM limit.

Ant you know that the solution is to either `call sync/fsync/fdatasync`(#2157), `use the O_DIRECT flag`, or `set drop_caches`.

I think `drop_caches` is the only means of flushing the cache from outside the process.
How bad was the performance when you tried it? (Was there a difference with or without the sync command?)

Modifying s3fs itself is either using fdatasync like #2157 or using the `O_DIRECT` flag, but personally if it can be handled with the `O_DIRECT` flag, that method is acceptable.
@gaul How about do you think?

I understand that this issue is due to Container's Limit and OOM behavior, so it is different from bare metal, VM, etc. environments.

gaul:
Absolutely not.  `O_DIRECT` is strictly worse than your previous PR to `fdatasync` at some interval, instead flushing *on every write*.  I don't believe that you understand how operating systems work so I will try to explain but you should start with something like https://www.linuxatemyram.com/.

[...]

PR #2157, opened 2023-05-12
Update curl.cpp:  reduce memory cache use when copy the big file from s3fs mount path

tanguofu:
flush the file and clean the page cache when filepart download done

### Details

add a fdatasync when curl download the filepart finish this to make memory cache of this file to flush to disk.

if not add fdatasync , when copy the large files  this will take many active file cache .

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -703,7 +703,12 @@ size_t S3fsCurl::DownloadWriteCallback(void* ptr, size_t size, size_t nmemb, voi
     }
     pCurl->partdata.startpos += totalwrite;
     pCurl->partdata.size     -= totalwrite;
-
+    
+    // flush the file and clean the page cache
+    if (pCurl->partdata.size == 0){
+        fdatasync(pCurl->partdata.fd);
+    }
+	
     return totalwrite;
 }
 


src/curl.cpp
```diff
@@ -703,7 +703,12 @@ size_t S3fsCurl::DownloadWriteCallback(void* ptr, size_t size, size_t nmemb, voi
     }
     pCurl->partdata.startpos += totalwrite;
     pCurl->partdata.size     -= totalwrite;
-
+    
+    // flush the file and clean the page cache
+    if (pCurl->partdata.size == 0){
+        fdatasync(pCurl->partdata.fd);
```
gaul, line 709: Why is this desirable?  The kernel should flush its cache as needed and due to memory pressure.  Maybe you should configure your system differently if this is the behavior your want.

ggtakec: Is this a fix for #2156?

tanguofu: yes,  this fix make the s3fs use less  system page cache to write the big file to download， in my test， if do not add fdatasync， the cache use  will be 8GB-11GB when  copy a 13GB file from the dir of s3fs mount to  local disk。

ggtakec: @tanguofu 
(I may be slightly wrong)
When an object is downloaded with s3fs, the size of the temporary file used for the CURL buffer is small because it should be downloaded in pieces instead of downloading the entire object size at once.
So I don't think this fix will make a big difference in memory usage.
(But, it recycles CURL handles, so that may be affecting it.)
However, other issues also point out memory usage, so I can't deny the cause of this correction.
Also, what are your s3fs boot options? (I want to confirm if it is related)

tanguofu: i start s3fs use with options: `-o tmpdir=/tmpdir -o use_cache=/cachedir`; the memory is used by  `active_file` cache, this is crop from the oom log of the pod:
oom log

ggtakec: @tanguofu Thanks for your reply.
Let me ask you some questions.
I assume you are using s3fs in a docker (or kubernetes) container (pod) by #2156.
Is the partition specified for this `tmpdir` and `use_cache` a disk using `tmpfs`?

What worries me is the following:
The s3fs cache file will continue to be created unless you specify a space limit for it, so `tmpfs`=`memory` will keep increasing.

Would calling `fdatasync` like in this PR code reduce the memory size used?
(I think the dirty disk cache can be reduced, but the s3fs cache file remains in the cache diretory(`tmpfs`), so I think the memory size will eventually overflow.)

Could you clear the OOM by specifying the upper limit of the usage capacity of this cache file, such as by using the `ensure_diskfree` option, or by not using the cache file(or deleting it periodically by an external process)?

tanguofu: the `tmpdir` and `use_cache`  just store at the disk of the pod dir  instead of `tmpfs`;  which can be very by `df`;
 
i think this is easy to reproduction if  copy the large file in the pod  from mount s3fs dir to local dir; i test in the centos 7.9 os and k8s 1.18 version;

by call `fdatasync` in the download write function of curl write callback , the memory use of the pod is reduced from 10GB to the 10MB approximately, this is  verified   by `kubectl top pod -n xxx  xxx`；

our project we just use this patch and now the memory use is very little right now，i think use `fdatasync` to clear cache  take less  negative effects； at my test this make little effect on performance.

as far as know another choice is drop system cache by `sysctl -w vm.drop_caches=1` but this has more effect on performance.

ggtakec: @tanguofu Thanks for the explanation, I'm thinking of merging this.
(Because there are issue reports that the amount of memory used increases, in addition to your case)

However, after the merge I will add an option(ex. `need_datasync`).
It will make this process will only be enabled when explicitly specified that option, rather than the default behavior.

ggtakec (approved):

ggtakec:
@tanguofu 
#2169 will add the `need_datasync` option.
If you use this feature, please specify that option.
Thanks for your great help, and kineness.

src/curl.cpp
gaul: I strongly disagree with this unnecessary behavior.  Which other tools do this kind of proactive syncing?  I think it is misleading to users even if added as an option.  Suggest reverting before 1.93.

gaul: If you want to hack your programs to unnecessarily sync you should do this via an `LD_PRELOAD` hack and not pollute random programs.

ggtakec:
@tanguofu 
This PR was marked as "merged" because it was merged and but reverted.
However, this issue continues to be investigated.
After that, if you decide that this PR code is necessary, please request another PR.

I'm sorry for the inconvenience.

ggtakec:
@tanguofu 
Please continue the discussion in #2156 for OOM related memory issues including this PR.

tanguofu:
ok, many thanks @ggtakec

PR #2160, opened 2023-05-13
Add Fedora 38 and remove 36 from CI

gaul:


ggtakec (changes requested):
Need to change `linux-ci-helper.sh` file.
To add `fedora:38` instead of `fedora:36`.

gaul:
> Need to change `linux-ci-helper.sh` file.

Done.

ggtakec (approved):

PR #2162, opened 2023-05-14
Add Ubuntu 23.04 and remove 18.04

gaul:
18.04 is EOL:

https://ubuntu.com/blog/ubuntu-18-04-eol-for-devices

This removes a FUSE 3 blocker.  References #1159.

PR #2163, opened 2023-05-14
Address cppcheck 2.10 warnings

gaul:
References #2162.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,8 +101,8 @@ jobs:
 
       - name: Cppcheck
         run: |
-          # work around resource leak false positives on older Linux distributions
-          if cppcheck --version | awk '{if ($2 <= 1.86) { exit(1) } }'; then
+          # work around resource leak false positives on older and newer Linux distributions
+          if cppcheck --version | awk '{if ($2 <= 1.86 || $2 >= 2.10) { exit(1) } }'; then
             make cppcheck
           fi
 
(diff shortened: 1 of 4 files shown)


gaul:
This is a bit tricky -- old versions of cppcheck don't accept some kinds of suppressions but newer versions require them.  Maybe we could use a per-version suppression file?

ggtakec:
@gaul
It seems that cppcheck's false positives and cppcheck's `--suppress` syntax are not set well, so how about fixing it as follows?
(When I checked, it was able to build without problems.)

- The `Makefile.am` only adds `useStlAlgorithm` and does not change the `unmatchedSuppression` syntax:  
```
--suppress=unmatchedSuppression \
--suppress=useStlAlgorithm \
```

- In `src/fdcache_page.cpp`, cppcheck seems to be confused about the one-liner, so write it with an if statement as follows:  
```
PageList::page_status pstatus = PAGE_NOT_LOAD_MODIFIED;
if(is_loaded && is_modified){
    pstatus = PageList::PAGE_LOAD_MODIFIED;
}else if(!is_loaded && is_modified){
    pstatus = PageList::PAGE_MODIFIED;
}else if(is_loaded && !is_modified){
    pstatus = PageList::PAGE_LOADED;
}
```  
It looks different because it removes the one-liner, but the content is the same, so it shouldn't be a problem.

gaul:
Let's disable newer cppcheck to unbreak CI then we can revisit proper fixes.

ggtakec:
Acknowledged to postpone cppcheck's useStlAlgorithm error handling.
We will discuss this matter later.

PR #1932, opened 2022-04-13
Add AWS SSO authentication and option aws_home_dir (s3fs-fuse#1931)

VivienMoretto87:
This pull request add a new authentication AWS SSO. I have although add the option aws_home_dir to configure the home directory (only available for the authentication AWS SSO)

Issue #1931, opened 2022-04-08
Add AWS SSO Authentifcation 


VivienMoretto87:
Hi, 

I need to integrate the AWS SSO authentifcation in s3fs. I have just checked the opensource project aws-cli (SSOCredentialFetcher in credentials.py in the project https://github.com/aws/aws-cli/tree/v2/awscli ) and the s3fs seems not to be compatible with this kind of authentication. 

I have modified the code : 
- get the cache json file with SSO credentials (it is the result sha1 of a json file { 'startUrl': self._start_url, 'roleName': self._role_name, 'accountId': self._account_id }) 
- get credentials for the cache json file (parse the json with nlohmann library )

It seems to work. 

Can you tell if it is possible to contribute the community by pushing my code and give some advises to have a clearer implementation ? 

Regards,

ggtakec:
@VivienMoretto87 Thanks for your kindness.
s3fs-fuse doesn't currently support SSO authentication, so we'd appreciate it if you could make a PullRequest.
After that, @gaul and I will be able to review it.

As a solution to this problem, I am now submitting another PR (# 1927).
This PR is the code to delegate the authentication process to an external DSO.
I think that combining this PR code with the DSO( https://github.com/ggtakec/s3fs-fuse-awscred-lib) should enable AWS SSO authentication.
This DSO just calls aws-sdk-cpp and it does the same authentication process as awscli and can get the tokens it needs.

If you are interested, I would be grateful if you could try # 1927 as well.

VivienMoretto87:
Thank you for the answer, I will push my code. I tested it Friday and it seems to work (other authentication too). 

Regards, 
Vivien

Emru1:
Hey, can we expect this feature any time soon?
Looking forward to it

ggtakec:
#1927 adds `credlib` and `credlib_ops` options.
Authentication not currently supported by s3fs can be supported by specifying a dedicated DSO for these options.
Please try to create your own DSO or use a trusted one.
I provide https://github.com/ggtakec/s3fs-fuse-awscred-lib which can cover all AWS authentication.

We will continue to add support for AWS and other cloud-based authentication, but for now please try to use the above DSO for AWS authentication which are not supported by s3fs.

PR #2165, opened 2023-05-15
Add AWS SSO authentication and option aws_home_dir (s3fs-fuse#1931) 

darioackermann:
#1931 
already open MR: #1932 

### Details
* Adds credentials grabbing from SSO

This is not very different from #1932 
I have made the integration tests work (except for centos and rockylinux, for which the header is not available as package - what should we do there?) and adjusted stuff such that cppcheck doesn't complain, and resolved merge conflicts.

We could default that SSO support is not enabled automatically for centos/rockylinux builds. (and add a flag for enabling it).


ggtakec:
@darioackermann 
I haven't been able to comment all of them yet, but let me ask you a quick question.

First of all, I don't think there is any processing when the SSO session token expires, but will it be updated outside of s3fs?
In general, is it enough that the cache file exists and is available to s3fs only for the duration of the token's validity period?
I am concerned that SSO(external authentication) generally has an expiration date.

The openssl code (SHA1) should be implemented in a separate file.
You will also need to implement it with gnuTLS and NSS.

I use nlohmann-json as my JSON parser, but I don't want to use it because it deviates from c++03 (I know it's old).
(exmaple, libyaml can also support c++03.)
However, I wonder if it should be included just to parse this SSO cache file.

Thanks for your help.

ggtakec:
@darioackermann Thenks for your reply.

Regarding SSO support, I think @sqlbot's [comment](https://github.com/s3fs-fuse/s3fs-fuse/pull/1932#issuecomment-1104339561) in the previous PR makes sense, and I also agree with him.

> but the user would notice since the share is unavailable then (?)

Since s3fs usually runs on the server-side(background), I don't think the user will notice it easily if the token expires.
The behavior of s3fs and the file system at that time is presented to the user as errors such as `file cannot be read`, `file cannot be written`, `file not found` and etc.
I think it is difficult for the user to judge that SSO token is expired.

It would be useful if there was another application(i don't know if it exists) that would notice the SSO token update and automatically redo the SSO login.
But I think it only makes sense on the `not server-side` (personal desktops, laptops, etc.).

@gaul
Please give me your opinion.

Please let me discuss in this PR whether this AWS SSO support should be done or not.

#### NOTE
If we proceed with this PR(if we decide to support SSO), I would like you to make the following modifications to the source code.

1. `SHA1` code requires version of `OpenSSL`/`gnuTLS`/`NSS`.  
I could provide you with a sample implementing these.
1. Don't use `nlohmann-json`, please change to code by `libyaml` or `coding simple parser`.  
`libyaml` is a library that can also parse `json`.(not `yaml` only)  
I'm sure I can provide a sample, either libyaml or a parser specifically for AWS files.

darioackermann:
Yes: Auth by SSO is definitely not a server side feature and nothing that will work on an unmonitored system.

>It would be useful if there was another application(i don't know if it exists) that would notice the SSO token update and automatically redo the SSO login.

You mean initiate the SSO login? There's always at least a manual click required (if the user is already logged in at AWS)



Automating the whole process would mean that we'd somehow have to store the credentials what is not the point of it..

> libyaml is a library that can also parse json
Thanks for the clarification, got misled by the library name :))

darioackermann:
Second thought: it would be nice to have a possibility to inform the user about the expired SSO token, but I don't really see a way that can be used?

ggtakec:
@darioackermann Thanks for your reply.
I understand the point of using IAM keys only when necessary.

However, the problem with expiring tokens is still that the cause of the error is hidden for the end user.

As you know, when the SSO token expires, s3fs tries to use that token and fails.
And since s3fs can only report the failure with errno, it cannot tell the party that performed the file manipulation operation that the token has expired.
Even if we can identify the cause from the s3fs log output, it is difficult to tell the end user.

It would be nice if there was some way to do it, but if this feature were built in, it would likely cause more confusion when the token expires than the inconvenience of not being able to use SSO token.
Our concerns are focused on this point.

mgaunard:
If the credentials expire, it's the same as the drive failing or being hot-removed.

This feature is necessary for a lot of people, I see no harm in adding it. If you don't like the concept of temporary credentials, just use permanent ones instead.

bemyak:
Hei,
I've tested the @darioackermann 's fork in combination with `libs3fsawscred.so` and it works!
A bit slowish maybe, but it's much better than having no possibility to authenticate with SSO at all.

> Since s3fs usually runs on the server-side(background) …

This is not the case at least for me, I use s3fs-fuse to browse files on S3 from my computer. I think it's a valid use case, and kindly ask you to reconsider merging this, @ggtakec :pray:

darioackermann:
There should be no need for using additional libraries when using the `aws sso login --profile profile_name` command.

I am still willing to make the needed adjustments, but haven't heard back that it would be merged eventually.

bemyak:
@darioackermann If I run it without the plugin (after `aws sso login`), I get `s3fs: Not found access key/secret key in passwd file.` error, even though `aws s3 ls` works :shrug:

kuza55:
Hi,

I wanted to ping this as another user invested in this as an option.

I want to use s3fs on my development machine and keep using SSO for auth.

I understand that FUSE cannot report detailed errors, but you get used to having to reauth every morning and that errors you experience before then have a clear cause.

gaul:
Is there a path forward with this PR?

ggtakec:
Regarding AWS TOKEN-specific processing(SSO), even if it is incorporated into s3fs itself, users will be confused because it cannot reissue TOKEN.
But narrowing the scope of support for that reason would be denying availability.

Therefore, I have provided the following external DSO that can delegate the authentication used by s3fs. (This wraps the authentication function of the AWS SDK)
https://github.com/ggtakec/s3fs-fuse-awscred-lib

I think it would be better to have users use the above DSO if necessary, rather than proceeding with this PR.
(Maintenance is not keeping up, but I will start if necessary)

gaul:
Closing due to inactivity please reopen if you want to continue.

PR #2168, opened 2023-05-27
Specify `_FORTIFY_SOURCE=3`

gaul:
This can find more kinds of buffer overflows:

https://developers.redhat.com/articles/2022/09/17/gccs-new-fortification-level


ggtakec (approved):
@gaul Thanks for the suggestion, I believe this can make s3fs even more secure.

PR #2169, opened 2023-05-27
Added need_datasync option

ggtakec:
#2157

### Details
By merging #2157, it is now possible to sync data when downloading.
However, I want to limit this call to `fdatasync` with this `need_datasync` option(which is disabled by default).
But we will check in the long term if this option will be no longer necessary.

@gaul 
This PR ties in with #2157 and will be merged soon.


gaul:
This is the wrong approach as discussed in https://github.com/s3fs-fuse/s3fs-fuse/pull/2157/files#r1207868619.

ggtakec:
I stop to request to merge this.

PR #2175, opened 2023-05-30
Remove wait and check loop from mk_test_file

gaul:
This appears to be some kind of eventual consistency check.  This should have no effect given S3Proxy and recent AWS strong consistency. Also it is likely ineffective given the other test object creation operations.


ggtakec (approved):
It seems this patch is a thing of the past as you said.
I could confirm that there is no problem now.
(I also manually tested on macOS to confirm that there were no issues with this change.)
Thanks.

PR #2177, opened 2023-05-31
Use smart pointer to manage pcfstat object

eryugey:
Previously pcfstat points to a raw pointer, and it may be leaked if
function returned before deleting it.

So use smart pointer to automatically release the object.

Note that currently s3fs only uses c++03, so we use auto_ptr here, not
unique_ptr, which requires c++11.

Fixes: https://github.com/s3fs-fuse/s3fs-fuse/commit/6ca5a24a7f29718967de1b8cc78dfe68c839729a ("Fix two inconsistency issues between stat cache and cache file (https://github.com/s3fs-fuse/s3fs-fuse/pull/2152)")
Signed-off-by: Eryu Guan <eguan@linux.alibaba.com>

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -24,6 +24,7 @@
 #include <unistd.h>
 #include <limits.h>
 #include <sys/stat.h>
+#include <memory>
 
 #include "common.h"
 #include "fdcache_entity.h"
@@ -485,7 +486,7 @@ int FdEntity::Open(const headers_t* pmeta, off_t size, const struct timespec& ts
         bool  need_save_csf = false;  // need to save(reset) cache stat file
         bool  is_truncate   = false;  // need to truncate
 
-        CacheFileStat* pcfstat = NULL;
+        std::auto_ptr<CacheFileStat> pcfstat;
 
         if(!cachepath.empty()){
             // using cache
@@ -500,7 +501,7 @@ int FdEntity::Open(const headers_t* pmeta, off_t size, const struct timespec& ts
             }
 
             // open cache and cache stat file, load page info.
-            pcfstat = new CacheFileStat(path.c_str());
+            pcfstat.reset(new CacheFileStat(path.c_str()));
 
             // try to open cache file
             if( -1 != (physical_fd = open(cachepath.c_str(), O_RDWR)) &&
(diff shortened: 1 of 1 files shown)


gaul:
This requires C++11 and s3fs currently only requires C++03.  We haven't really talked about this before since C++11's benefits are modest but #1068 has discussion about a library we would like to use which requires C++14.

src/fdcache_entity.cpp
```diff
@@ -643,8 +644,8 @@ int FdEntity::Open(const headers_t* pmeta, off_t size, const struct timespec& ts
                 S3FS_PRN_WARN("failed to save cache stat file(%s), but continue...", path.c_str());
             }
         }
-        if(pcfstat){
-            delete pcfstat;
+        if(pcfstat.get() != nullptr){
+            delete pcfstat.release();
```
gaul, line 648: This seems to be an unusual way to use `unique_ptr`, instead lifetimes should automatically delete this.

gaul (changes requested):
We are definitely leaking on error paths, probably in other parts of the code as well.  I don't have a strong opposition to this but let's discuss it in the related issue.

gaul:
Alternatively we could use a `unique_ptr` reimplementation that I experimented with in https://github.com/gaul/s3fs-fuse/pull/new/c++/unique_ptr.  However these usually have bugs.

src/fdcache_entity.cpp
eryugey: Removed.

eryugey:
> We are definitely leaking on error paths, probably in other parts of the code as well. I don't have a strong opposition to this but let's discuss it in the related issue.

OK, I switched to auto_ptr this time, and we could switch it back to unique_ptr once s3fs adopts c++11 or above.

ggtakec (approved):
@eryugey Thanks for the fix code.
_(This code was recently modified and created many leaks.)_

@gaul Please check.
I think we can merge this.

gaul:
Discussing this in #2195.

ggtakec:
@gaul Would you like to close this PR and contain this to #2195?

gaul:
#2195 is a better approach but let's wait until we sort out 1.93.

PR #2181, opened 2023-06-06
test: Exclude sys/sysmacros.h on FreeBSD

diizzyy:
Fixes build issue on FreeBSD

### DIFF
--- a/test/mknod_test.c
+++ b/test/mknod_test.c
@@ -28,7 +28,10 @@
 #include <string.h>
 #include <errno.h>
 #ifndef __APPLE__
+#elif
+#ifndef __FreeBSD__
 #include <sys/sysmacros.h>
+#endif
 #endif
 
 //---------------------------------------------------------


test/mknod_test.c
```diff
@@ -28,8 +28,11 @@
 #include <string.h>
 #include <errno.h>
 #ifndef __APPLE__
+#elif
+#ifndef __FreeBSD__
```
gaul, line 32: I am skeptical of code that CI does not exercise which will rot over time.  Is there some other way to approach this?

diizzyy: I'm not sure what you mean, that file is Linux-kernel specific.

gaul: Could we call `system` to invoke `mknod` instead?

gaul: Thinking about this some more I think the `#ifdef` is fine.  But CI is broken so we cannot merge this.

gaul: @diizzyy Could you move forward with this PR?

diizzyy: I'll try to have a look again next week, unfortunately I don't have a Linux box available so I'm not sure why it fails.

gaul:
Closing this due to inactivity.  Please reopen if still relevant.  If we can improve FreeBSD support I would like to do that and especially get its packages on the latest version.

PR #2191, opened 2023-06-21
Add support for FUSE-T on macos

macos-fuse-t:
Add support for user space fuse (FUSE-T) on macos. 
FUSE-T doesn't require installation of kernel extensions (KEXT) as opposed to osxfuse.
Details of the project can be found here: https://github.com/macos-fuse-t/fuse-t
To install FUSE-T use brew:
  brew tap macos-fuse-t/homebrew-cask 
  brew install fuse-t

### Details

Change autoconf to discover and use FUSE-T if present, otherwise fallback to osxfuse

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -39,6 +39,7 @@ CXXFLAGS="$CXXFLAGS -Wall -fno-exceptions -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOUR
 dnl ----------------------------------------------
 dnl For macOS
 dnl ----------------------------------------------
+found_fuse_t=no
 case "$target" in
    *-cygwin* )
       # Do something specific for windows using winfsp
@@ -48,6 +49,7 @@ case "$target" in
    *-darwin* )
       # Do something specific for mac
       min_fuse_version=2.7.3
+      min_fuse_t_version=1.0.20
       ;;
    *)
       # Default Case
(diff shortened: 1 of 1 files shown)


gaul:
@macos-fuse-t Please investigate CI failures.

@ggtakec fuse-t is an alternate FUSE implementation based on the macOS NFS client.  In addition to a newer and maintained code base, this may give us a better path forward in CI which cannot use kexts.

macos-fuse-t:
please re-run CI

gaul:
Thank you for your contribution @macos-fuse-t!  Please update the README with newer Homebrew instructions if you package this.

ggtakec:
@macos-fuse-t Thank you for the corresponding PR of `fuse-t`!
I will prepare and post a PR to restore macos testing to the CI process.
(However since an error is output, it will be necessary to investigate the cause)

macos-fuse-t:
@ggtakec  Sure, I will investigate the failures if any

ggtakec:
@macos-fuse-t Thanks for your kindness.
I created a draft PR for #2198.
It looks like some tests are still failing, so I'll try to debug these.

gaul:
@macos-fuse-t Does FUSE-T support the FUSE3 API?

macos-fuse-t:
No, fuse3 APIs are not supported at this moment. Future support is planned though.

PR #2194, opened 2023-06-24
Protect FdEntity::physical_fd with fdent_lock

gaul:
Found via ThreadSanitizer.


gaul:
I found a few more instances of this.

Issue #2179, opened 2023-05-31
Consider requiring C++11


gaul:
Currently s3fs only relies on C++03 which gives it compatibility with ancient Linux distros.  Upgrading to C++11 would allow `optional`, `span`, `unique_ptr`, `unordered_map`, and some modest syntactic sugar.  A bigger benefit is moving closer to compatibility with the AWS SDK (C++14) in #1068.  Recently brought up in #2177.

ggtakec:
If we don't have (or very few) environments that still have to rely on C++03, then I think it's fine to proceed with the migration to C++11.
_(Because I have not been able to grasp the affected users and environments that depend on C++03, I would like to know that.)_

Regarding the `AWS SDK(CPP)`, we should consider the license of s3fs, how to link it, etc.
_The only thing you want to incorporate into s3fs is around authentication, right?_
_(It has the same function as the library created at https://github.com/ggtakec/s3fs-fuse-awscred-lib, so it is easy to code and embed)_

ggtakec:
@gaul 
During the regular build last weekend, SUSE's cppcheck has an error because cppcheck has been upgraded to 2.10.

Most of the errors say we should use the std's algorithm.
For checking the condition that it must be C++03 currenty, I tried to test using C++11(lambda function) as a trial.
As a [result](https://github.com/ggtakec/s3fs-fuse/actions/runs/5324489766), only CentOS7 failed and other OSs(specified by GithubActions of the current s3fs) succeeded.

If we drop C++03 support(it means stop CentOS7 support), I can PR the current code, but if we still want to continue C++03 support, I create a PR after changing the code.

Given this situation, I think we can move to C++11.

gaul:
[RHEL7 includes GCC 4.8.x](https://access.redhat.com/solutions/19458) which has strong C++11 support so I think we can use most features: https://gcc.gnu.org/projects/cxx-status.html#cxx11.  I think the error you encountered might just be some configuration needed, e.g., specify `-std=c++11`.  Note that there are a bunch of other library changes that newer libstdc++ might accidentally expose in C++11 mode.

I prefer to use C++11 or newer since features like `unique_ptr` make it easier to write correct code but more importantly writing modern C++ makes it easier for contributors.  If there is some backwards compatibility issue I think we can drop support for that distro or they can use a non-standard compiler on that distro.

[...]

PR #2195, opened 2023-06-24
Replace some raw pointers with `std::unique_ptr`

gaul:
This simplifies code paths and makes memory leaks less likely.  It also makes memory ownership more explicit by requiring `std::move`. This commit requires C++11.  References #2179.

### DIFF
--- a/src/curl_multi.cpp
+++ b/src/curl_multi.cpp
@@ -60,19 +60,17 @@ bool S3fsMultiCurl::ClearEx(bool is_all)
 {
     s3fscurllist_t::iterator iter;
     for(iter = clist_req.begin(); iter != clist_req.end(); ++iter){
-        S3fsCurl* s3fscurl = *iter;
+        S3fsCurl* s3fscurl = iter->get();
         if(s3fscurl){
             s3fscurl->DestroyCurlHandle();
-            delete s3fscurl;  // with destroy curl handle.
         }
     }
     clist_req.clear();
 
     if(is_all){
         for(iter = clist_all.begin(); iter != clist_all.end(); ++iter){
-            S3fsCurl* s3fscurl = *iter;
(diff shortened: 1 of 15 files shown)


gaul:
CentOS 7 CI fails since it cannot find the `std::unique_ptr` header.  I wonder if there is some way to configure it to use something newer?

gaul:
It appears that RHEL 7 includes a mostly-C++11 compiler but not the C++11 ABI/library.

ggtakec:
@gaul 
On CentOS7, if adding `-std=c++11`CXXFLAG, this PR will be able to be built without errors.(see. #2197)
This makes `c++03` unsupported, but that's okay because we're moving in that direction.

src/curl_multi.cpp
```diff
@@ -271,35 +269,36 @@ int S3fsMultiCurl::MultiRead()
...
+                    S3fsCurl::ResetOffset(s3fscurl.get());
                 }
 
                 // For retry
+                S3fsCurl* retry_ptr = nullptr;
                 if(RetryCallback){
-                    retrycurl = RetryCallback(s3fscurl);
+                    retry_ptr = RetryCallback(s3fscurl.get());
+                    retrycurl.reset(retry_ptr);
                     if(NULL != retrycurl){
```
ggtakec, line 292: Shouldn't `if(NULL != retrycurl){` be changed to `if(retrycurl.get()){`?

ggtakec (changes requested):
The part I noticed was commented (only one point), other than that I think it's fine.
Now we can take advantage of C++11, which is a good thing. :-)

src/curl_multi.cpp
gaul: Changed to `retry_ptr` (same thing).

ggtakec (approved):

PR #2197, opened 2023-06-25
Supported cppcheck 2.10(and changed std from c++03 to c++11 for RHEL7)

ggtakec:
#2163
#2162 
#2179

### Details
Fixed errors detected by `cppcheck v2.10 (SUSE leap:15)`.

This reverts the setting that was temporarily disabled(`v2.10` and `useStlAlgorithm`) in #2163.
With this fix, cppcheck checks can be done on all OSs.

Support for `CentOS7`(which was discussed in #2195) is resolved by adding the `-std=c++11` CXXFLAGS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,7 +102,7 @@ jobs:
       - name: Cppcheck
         run: |
           # specify the version range to run cppcheck (cppcheck version number is x.y or x.y.z)
-          if cppcheck --version | sed -e 's/\./ /g' | awk '{if (($2 * 1000 + $3) <= 1086 || ($2 * 1000 + $3) >= 2010) { exit(1) } }'; then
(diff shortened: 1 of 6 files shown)


gaul:
I don't like these std::algorithm idioms which seem harder to read.  Can we just disable the cppcheck for them instead?

ggtakec:
@gaul It is no problem, please wait a while.
(I don't like clutter either. I was trying to follow the cppcheck instructions, but if we can agree, I'll remove the clutter.)

ggtakec:
@gaul Reverted parts using std::algorithm(and numeric).
_I left modified two lines where the TAB character was detected, because we will forget to fix those if I do not correct it when I find it._

gaul (approved):

ggtakec:
@gaul 
I'll reflect the merge of #2229.
But, on top of that, cppcheck has been changed to 2.11.x on macOS, and another warnings have appeared, so that cannot be resolved with this PR.
I will include them and correct them and request again.
I'm sorry for the double hassle.

PR #2198, opened 2023-06-25
Reverted the macos CI process(using macos-fuse-t)

ggtakec:
#2154
#2191

### Details
By using `macos-fuse-t`, CI process can support macos11 or later.
(Thanks to @macos-fuse-t.)

CI processing(until just before the test) has been confirmed on both macos11/12. 
But the macos CI processing is macos12 only.
_(If necessary, macos11 can also be added.)_

Part of the patch #2197 is applied to pass the cppcheck2.10 test.
_(I will rebase after merged #2197.)_

#### NOTE
This PR has not passed some parts of tests yet, so it is marked as a `Draft`.

### DIFF
--- a/test/test-utils.sh
+++ b/test/test-utils.sh
@@ -396,14 +396,41 @@ function make_random_string() {
     else
         local END_POS=8
     fi
-
-    "${BASE64_BIN}" --wrap=0 < /dev/urandom | tr -d /+ | head -c "${END_POS}"
+    if [ "$(uname)" = "Darwin" ]; then
+        local BASE64_OPT="--break=0"
+    else
+        local BASE64_OPT="--wrap=0"
+    fi
+    "${BASE64_BIN}" "${BASE64_OPT}" < /dev/urandom 2>/dev/null | tr -d /+ | head -c "${END_POS}"
 
     return 0
 }
 
 function s3fs_args() {
-    ps -o args -p "${S3FS_PID}" --no-headers
+    if [ "$(uname)" = "Darwin" ]; then
+        ps -o args -p "${S3FS_PID}" | tail -n +2
+    else
+        ps -o args -p "${S3FS_PID}" --no-headers
+    fi
+}
+
+#
+# $1:    sleep seconds
+# $2:    OS type(ex. 'Darwin', unset(means all os type))
+#
+# [NOTE] macos fuse-t
+# macos fuse-t mounts over NFS, and the mtime/ctime/atime attribute
+# values are in seconds(not m/u/n-sec).
+# Therefore, unlike tests on other OSs, we have to wait at least 1
+# second.
+# This function is called primarily for this purpose.
+#
+function wait_ostype() {
+    if [ -z "$2" ] || uname | grep -q "$2"; then
+        if [ -n "$1" ] && ! (echo "$1" | grep -q '[^0-9]'); then
+            sleep "$1"
+        fi
+    fi
 }
 
 #
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,18 +127,15 @@ jobs:
           make ALL_TESTS=1 check -C test || (test/filter-suite-log.sh test/test-suite.log; exit 1)
 
   # [NOTE]
-  # This Job does not work for macOS 11 and later because load_osxfuse returns exit code = 1.
-  # Apple states "You must be signed in as an administrator user to install new kernel
-  # extensions, and your Mac must be rebooted for the extensions to load.", so it needs
-  # to reboot OS.
-  # As of May 2023, GitHub Actions are no longer able to launch macos 10.15 as runner,
-  # so we can not run this Job.
-  # In the future, if it is found a solution, we will resume this Job execution.
+  # Using macos-fuse-t
+  #   This product(package) is a workaround for osxfuse which required an OS reboot(macos 11 and later).
+  #   see. https://github.com/macos-fuse-t/fuse-t
+  # About osxfuse
+  #   This job doesn't work with Github Actions using macOS 11+ because "load_osxfuse" returns
+  #   "exit code = 1".(requires OS reboot)
   #
-  macos10:
-    if: false
-
-    runs-on: macos-10.15
+  macos12:
+    runs-on: macos-12
 
     steps:
       - name: Checkout source code
@@ -149,14 +146,15 @@ jobs:
           TAPS="$(brew --repository)/Library/Taps";
           if [ -e "$TAPS/caskroom/homebrew-cask" ]; then rm -rf "$TAPS/caskroom/homebrew-cask"; fi;
           HOMEBREW_NO_AUTO_UPDATE=1 brew tap homebrew/homebrew-cask
+          HOMEBREW_NO_AUTO_UPDATE=1 brew tap macos-fuse-t/homebrew-cask
 
-      - name: Install osxfuse
+      - name: Install fuse-t
         run: |
-          HOMEBREW_NO_AUTO_UPDATE=1 brew install osxfuse
+          HOMEBREW_NO_AUTO_UPDATE=1 brew install fuse-t
 
       - name: Install brew other packages
         run: |
-          S3FS_BREW_PACKAGES='automake cppcheck python3 coreutils gnu-sed shellcheck';
+          S3FS_BREW_PACKAGES='automake cppcheck python3 coreutils gnu-sed shellcheck jq';
           for s3fs_brew_pkg in ${S3FS_BREW_PACKAGES}; do if brew list | grep -q ${s3fs_brew_pkg}; then if brew outdated | grep -q ${s3fs_brew_pkg}; then HOMEBREW_NO_AUTO_UPDATE=1 brew upgrade ${s3fs_brew_pkg}; fi; else HOMEBREW_NO_AUTO_UPDATE=1 brew install ${s3fs_brew_pkg}; fi; done;
 
       - name: Install awscli2
@@ -164,10 +162,6 @@ jobs:
           cd /tmp
           curl "https://awscli.amazonaws.com/AWSCLIV2.pkg" -o "AWSCLIV2.pkg"
           sudo installer -pkg AWSCLIV2.pkg -target /
-
-      - name: Check osxfuse permission
-        run: |
-          if [ -f /Library/Filesystems/osxfusefs.fs/Support/load_osxfusefs ]; then sudo chmod +s /Library/Filesystems/osxfusefs.fs/Support/load_osxfusefs; elif [ -f /Library/Filesystems/osxfuse.fs/Contents/Resources/load_osxfuse ]; then sudo chmod +s /Library/Filesystems/osxfuse.fs/Contents/Resources/load_osxfuse; else exit 1; fi
 
       - name: Build
         run: |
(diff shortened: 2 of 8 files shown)


macos-fuse-t:
Let me know if I need to look at something. One thing I definitely noticed that extended attributes are not working.

gaul:
@macos-fuse-t I guess this fails due to missing xattr support?

ggtakec:
@gaul The problem I'm detecting now is that mtime/ctime/atime are not consistent in CI testing.
Outside of CI testing, see #2205.

I've been detecting and working on invisible directories and files. (It is similar to the phenomenon reported by some users other than macOS. On macOS, an fts_read error occurs.)

ggtakec:
Fixed CI test error.

I had to fix some tests.
The fix is probably not the right way, but these are needed to fix for running fuse-t with Github Actions.

Its contents are as follows:

### Time lag
In attribute updates, there are cases where values cannot be obtained without waiting(sleep).
In particular, there seems to be a large time lag in updating attributes in parent directories.
For fuse-t, the time of `mtime`/`ctime`/`atime` is up to seconds(not nsec), and if we do not wait at least 1 second or more, the test will fail.
But I didn't understand why I needed more waiting(sleep).
Our current test code takes that into consideration and gives it enough waiting.
This additional sleep time is done with the `wait_ostype` helper function, so you can find it in your test script.

The following problems also contain content related to time lag.
Check the test code for details.

### test_hardlink
In fuse-t, error strings(display output) cannot be obtained, so I'm dealing with that.

### test_extended_attributes
This is very mysterious.
When I get and display xattr, it seems that the string is not null-terminated.
(xattr's value length does not include null charactor, so that it maybe a reason.)
Propabry, I suspect this is a fuse-t issue.

### test_update_time_chown
For some reason the s3fs's chown handler is not being called.
Therefore, any time attributes are not updated.
The test is bypassed by not failing by not updating.
In my current test I'm chowning with the same UID value, but chowning to a different UID gives the same result.
The reason why attribute information cannot be obtained is probably because the handler on the s3fs side has not been called.
I was investigating the cause of this on the s3fs side, but it seems that s3fs is not the cause.

@macos-fuse-t
Do you have any information about the time lag from writing the file to being reflected?

@gaul
Other than the lag time, there are some tests that fail to run, which we are working around now. (or bypassed)
The code also contains other PR code, but will be rebased once they are merged.

macos-fuse-t:
Regarding the time lag I suspect some kind of attribute caching either in fuse-t (can be disabled) or on the macos nfs or file system level. I'll check and let you know. I'll also check the extended attribute issue.

ggtakec:
@macos-fuse-t 
I suspected fuse-t or nfs cache, but I haven't been able to investigate those options.
Also, the current test is in a dead loop somewhere.

macos-fuse-t:
I ran the tests locally and here's what I found:

in test_external_creation:
Caching lag is caused by caching in the nfs server. There's `-o attrcache=false` mount option which disables the attribute cache in fuse-t but doesn't do so for the nfs mount. I will update the code to disable nfs cache when the option is given. That should fix the test.

in test_update_directory_time_chown:
macOS ignores the call when uid and gid are the same so you'll have to modify the test to change either uid or gid for it to work. Together with disabling caching the test should pass.

in test_extended_attributes:
This one I don't get. The test should pass but it always fails. I manually applied the commands and it worked. Also the value shouldn't be null terminated. You can check it by executing `xattr -px` command in macOS. Perhaps you should use "grep -e" instead of "-q" for the regular expression?

So if you're ok with testing with `-o attrcache=false` I will update the fuse-t code.

ggtakec:
@macos-fuse-t Thank you for your kindness.
I'll try to run test with `-o attrcache=false`.
I will try to remove the waiting(sleep) in the test accordingly.

macos-fuse-t:
@ggtakec The latest release supports disabling NFS cache so you may try to run the CI without sleeps
Thank you

ggtakec:
@macos-fuse-t Thanks for your kindness.

But could you tell me how to specify the attracache options?
(Maybe I'm doing something wrong)

### About the attrcache option
I tried to run s3fs with `-o attrcache=false` on CI, but it did not recognize this option.( `fuse: unknown option 'attrcache=false'` )
Is there something wrong with the specification method?

### test_update_directory_time_chown
When I was trying CI testing in a few cases, I also tried a case where I actually changed the UID, but it was unsuccessful.
This may also be affected by `attrcache`, so I'll try this with disabling the cache later.

### test_extended_attributes
This is the same for other Linux, when s3fs sets the value of xattr(eg. `key=val`), the value length does not include null characters.(It is `3` for `val`)
On other Linux, I can read `key` xattr's value which is exactly `val`(terminated null).
However, in the case of macos, it becomes a strange situation as follows:
```
xattr -w key1 value01 <file>
xattr -w key2 value2 <file>
xattr -p key1 <file>
-> value01
xattr -p key2 <file>
-> value21
```
Previous values were displayed intermingled, as in the example below.
(It seems that the value is not terminated with null character and the previous buffer seems to be read).
Then I changed the test code, the grep parameter was changed to '^value1' because the test could not pass by `grep '^value01$'`.

This may also be affected by `attrcache`, so I'll try this with disabling the cache later.

macos-fuse-t:
@ggtakec My mistake, It's actually `-o noattrcache`

ggtakec:
@macos-fuse-t Thanks for youe kindness.

I could test for updating stats using `noattrcache` option.
However, becasue the requests send/recieve via NFS(?), those time attributes are changed only in units of seconds, so I add a sleep command to some parts.

`test_update_directory_time_chown` can set a different UID, but due to a dead loop(probably interaction input for `sudo` execution) when running CI, `fuse-t` is now a test case that does not update stat information.
(I will consider fixing this in the future.)

ggtakec:
@macos-fuse-t About `extended attributes`
Here is my test function when I checked.
_Below is based on the current master test function with the fix I found for this issue._
```
function test_extended_attributes {
    describe "Testing extended attributes ..."

    rm -f "${TEST_TEXT_FILE}"
    touch "${TEST_TEXT_FILE}"

    # set value
    set_xattr key1 value1 "${TEST_TEXT_FILE}"
    get_xattr key1 "${TEST_TEXT_FILE}" | grep -q '^value1$'

    # append value
    set_xattr key2 next2 "${TEST_TEXT_FILE}"
    get_xattr key1 "${TEST_TEXT_FILE}" | grep -q '^value1$'
    get_xattr key2 "${TEST_TEXT_FILE}" | grep -q '^next2$'     **** ERROR

    # remove value
    del_xattr key1 "${TEST_TEXT_FILE}"
    get_xattr key1 "${TEST_TEXT_FILE}" && return 1
    get_xattr key2 "${TEST_TEXT_FILE}" | grep -q '^next2$'

    rm_test_file
}
```
In the ERROR line above, the read value was `next21`.

macos-fuse-t:
@ggtakec If apply a sleep after set_xattr the problem's gone. Maybe something's wrong with the bucket emulation ?

 # append value
sleep 1
set_xattr key2 next2 "${TEST_TEXT_FILE}"
sleep 1
get_xattr key1 "${TEST_TEXT_FILE}" | grep -q '^value1$'
get_xattr key2 "${TEST_TEXT_FILE}" | grep -q '^next2$'

ggtakec:
I still need to modify the code for this CI for macos a bit more, so I will be updating the code again.

ggtakec:
@macos-fuse-t _I'm sorry for that It took me a while to do some research._

I updated code in this PR to code rebased to latest master (and added #2283 code).

But, the `macos fuse-t` test still fails with `test_extended_attribute`.

The failing part is the (*) part below.
```
    # append value
    set_xattr key2 value2 "${TEST_TEXT_FILE}"
    get_xattr key1 "${TEST_TEXT_FILE}" | grep -q '^value1$'     ...(*)
    get_xattr key2 "${TEST_TEXT_FILE}" | grep -q '^value2$'
```
_The result of setting the xattr to `key1` and `key2` is fine._

When I changed this test code as follows, I was able to confirm that the value of `key1` was not read correctly.
```
     # append value
     set_xattr key2 value2 "${TEST_TEXT_FILE}"
     get_xattr key1 "${TEST_TEXT_FILE}"
     get_xattr key2 "${TEST_TEXT_FILE}"
```
The value of `key1` was `value2` which is the value set for `key2`.

The `s3fs_getxattr` function was called from FUSE(`fuse-t`) for reading `key1` and `key2` several times, but the buffer to returning the `key1` value was always NULL and which size was always 0.

If I'm right, the `getxattr` handler call is called without a buffer the first time (or several times) and returns the size of the value.
After that(if it works correctly), a buffer of the size of the value is given and the value can be returned.

In the example above, no call was made to allocate a buffer for `key1`.
Nevertheless, reading xttar succeeds and its value is displayed as the contents of `key2`.

This is the result of running with `-o noattrcache`.
(I feel that it seems that the cache for xattr is not disabled)

Do you have any idea what the cause is?

ggtakec:
#### Updated codebase to latest master and added one fix.

It reverts the fix added in #1498 (macos only, don't use stdbuf), MacOS and Linux both have the same stdbuf calls.
Because I've checked the status quo and confirmed that calling stdbuf doesn't cause any problems because of it.

alexfs:
@ggtakec I looked at it again and can finally explain the issue:
Apparently macOS still caches extended attributes even when told not to. This is why a sleep of 1 second after a second set xattr should actually work.  I'll try to see if it's possible to fix it more elegantly in the code.

ggtakec:
@alexfs Thank you for investigating.
It looks like the cache is still working, I added a 1 second sleep to `test_extended_attributes` and the error went away.
(Now updated code and running Github Actions)

ggtakec:
Unfortunately, `test_extended_attributes` seems to be able to pass the test, still even this latest code is not stable for CI testing. (It sometimes fails in other non-specific test case)

ggtakec:
> (It sometimes fails in other non-specific test case)

Sometimes the test fails due to insufficient free disk space (disk used for cache, etc.).
I will think about countermeasures.

ggtakec:
It seems that the insufficient disk space error(ENOSPC) can be avoided.

But I still haven't been able to solve the problem.
The `test_update_metadata_external_small_object` test does not complete.(This is only happening with Github Actions)
In the test, the attribute of the bucket root directory and the statvfs are read repeatedly, and an interval(ten or more seconds) occurs between repetitions.(There was a problem with statvfs, and even if I tried using the corrected code, the result was the same.)

@alexfs 
I'm still investigating, but if you notice anything please let me know.

ggtakec:
Rebased to master(546cdd0).
Still, testing the code in this PR results in an infinite loop. (only on macos12)

I have also made the following changes:
_(However, please note that these are code that have not yet been merged into separate PRs and are planned to be separated into separate PRs)_
- Fixed errors detected by cppcheck 2.12.0 on macos 12.
- There was a case where a CI error occurred in alpine, and it was fixed (#2260)
- Call to invalidate cache for stat command (#2198)

test/test-utils.sh
```diff
@@ -396,14 +396,41 @@ function make_random_string() {
...
+#
+# [NOTE] macos fuse-t
+# macos fuse-t mounts over NFS, and the mtime/ctime/atime attribute
+# values are in seconds(not m/u/n-sec).
+# Therefore, unlike tests on other OSs, we have to wait at least 1
+# second.
+# This function is called primarily for this purpose.
+#
+function wait_ostype() {
+    if [ -z "$2" ] || uname | grep -q "$2"; then
```
gaul, line 429: Should this be `[ "$(uname)" = "Darwin" ]` as above?

ggtakec:
I will share my progress along the way.

I'm currently continuing to modify the code for this PR.
While trying to support macos 12, I have detected potential bugs and are currently posting separate PRs.
The plan is to investigate and fix the remaining more two issues.

ggtakec:
This PR code was rebased the master branch.
It also incorporates the PR code of #2341, #2348 and #2353.
I think this allows support for macos12(`masoc fuse-t`).
Once these PRs are merged or modified, I will update/rebase the code again so it is no longer a Draft.

ggtakec:
@gaul Thank you for merging this PR.
However, there may still be problems with macos, still I was researching it.
Like other fixed bugs, it may be caused by the order of handlers called from FUSE.
It is a fix for the entire s3fs, so I will send a PR once the cause and fix are completed.

Issue #2203, opened 2023-07-01
StreamUpload is not working


VVoidV:
#### Version of s3fs being used (`s3fs --version`)
Amazon Simple Storage Service File System V1.92 (commit:unknown) with OpenSSL

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)
2.9.9

#### Kernel information (`uname -r`)
6.3.7-200.fc38.x86_64

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)
NAME="Fedora Linux"
VERSION="38 (Workstation Edition)"
ID=fedora
VERSION_ID=38
VERSION_CODENAME=""
PLATFORM_ID="platform:f38"
PRETTY_NAME="Fedora Linux 38 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:38"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f38/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=38
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=38
SUPPORT_END=2024-05-14
VARIANT="Workstation Edition"
VARIANT_ID=workstation

#### How to run s3fs, if applicable

[x] command line  
[] /etc/fstab  

```
s3fs -o streamupload,dbglevel=info,use_cache=/var/s3fs_cache,logfile=/tmp/s3fs.log,use_path_request_style,allow_other,del_cache,connect_timeout=300,readwrite_timeout=120,url=http://192.168.204.132:4566 bucket /exports/test/
```

#### s3fs syslog messages (`grep s3fs /var/log/syslog`, `journalctl | grep s3fs`, or `s3fs outputs`)

```
```

### Details about issue

streamupload is not working.
use **dd** to write a 100MB file, dd while hung forever.
```
[...]

Issue #2188, opened 2023-06-15
Unmounting segfault


arbrebinaire:
#### Version of s3fs being used (`s3fs --version`)

V1.92 (commit:unknown) with OpenSSL

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)

Version     : 2.9.7
Packager    : infrastructure@rockylinux.org
Vendor      : Rocky

#### Kernel information (`uname -r`)

4.18.0-425.19.2.el8_7.x86_64

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)

NAME="Rocky Linux"
VERSION="8.8 (Green Obsidian)"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.8"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Rocky Linux 8.8 (Green Obsidian)"
ANSI_COLOR="0;32"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:rocky:rocky:8:GA"
HOME_URL="https://rockylinux.org/"
BUG_REPORT_URL="https://bugs.rockylinux.org/"
SUPPORT_END="2029-05-31"
ROCKY_SUPPORT_PRODUCT="Rocky-Linux-8"
ROCKY_SUPPORT_PRODUCT_VERSION="8.8"
REDHAT_SUPPORT_PRODUCT="Rocky Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="8.8"

#### How to run s3fs, if applicable

```
umount  "/mnt/some_mountpoint"
```

#### s3fs syslog messages (`grep s3fs /var/log/syslog`, `journalctl | grep s3fs`, or `s3fs outputs`)

```
[Wed Jun 14 21:24:40 2023] s3fs[375044]: segfault at 7f3aa13c89d0 ip 00007f3aa4fd5550 sp 00007fff47292b90 error 4 in libpthread-2.28.so[7f3aa4fcc000+1b000]
[...]

PR #2204, opened 2023-07-01
fix streamuplod not working

VVoidV:

#2203  #2188 
### Details

fuse_main will do fork to daemonize. if ThreadPool is created before this, worker thread will not be placed in child process.

Move ThreadPool Init to s3fs_init and destory to s3fs_destory.

#2188  may be also related to this.  When umount s3fs, ThreadPool::Destroy() will dereference some stale handler.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -4241,6 +4241,11 @@ static void* s3fs_init(struct fuse_conn_info* conn)
          conn->want |= FUSE_CAP_BIG_WRITES;
     }
 
+    if(!ThreadPoolMan::Initialize(max_thread_count)){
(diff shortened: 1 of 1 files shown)


ggtakec (approved):
@VVoidV Thanks for your great help!

This was completely my mistake.
I should have known that fuse would fork, but I implemented it wrong.
With your modified code, I confirmed that the thread can be started correctly after forking.
(There are calling curl functions  such as bucket checks, but they are implemented without thread handlers, so they will work correctly with this modification.)

ggtakec:
As for #2188, I'll request if they can try this merge code.

PR #2206, opened 2023-07-04
Check FdEntity::Open() status correctly

eryugey:
FdEntity::Open() returns -errno on error, but FdManager::Open() only checks if its ret is -1. This may lead to use '-errno' as pseudo fd in next read or write, which would fail due to '-errno' is not in fent map.


gaul:
Agreed that this was incorrect.  Thank you for your contribution @eryugey!

The ThreadSanitizer error looks unrelated so merging this now.  I will try to look more closely at it this weekend.

PR #2212, opened 2023-07-12
Explicitly handle CommonPrefixes with nocompat_dir

gaul:
Previously the test missed listing implicit directories and another test was incorrect.  This fixes a regression from 1.91.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -21,6 +21,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <errno.h>
+#include <set>
 #include <unistd.h>
 #include <dirent.h>
 #include <sys/types.h>
@@ -3067,6 +3068,7 @@ struct multi_head_callback_param
 {
     void*           buf;
     fuse_fill_dir_t filler;
+    std::set<std::string> *filled;
 };
 
 static bool multi_head_callback(S3fsCurl* s3fscurl, void* param)
@@ -3096,6 +3098,7 @@ static bool multi_head_callback(S3fsCurl* s3fscurl, void* param)
             S3FS_PRN_INFO2("Could not find %s file in stat cache.", saved_path.c_str());
             pcbparam->filler(pcbparam->buf, bpath.c_str(), 0, 0);
         }
+        pcbparam->filled->insert(bpath);
     }else{
         S3FS_PRN_WARN("param(multi_head_callback_param*) is NULL, then can not call filler.");
     }
@@ -3170,6 +3173,7 @@ static int readdir_multi_head(const char* path, const S3ObjList& head, void* buf
     S3fsMultiCurl curlmulti(S3fsCurl::GetMaxMultiRequest());
     s3obj_list_t  headlist;
     int           result = 0;
+    std::set<std::string> filled;
 
     S3FS_PRN_INFO1("[path=%s][list=%zu]", path, headlist.size());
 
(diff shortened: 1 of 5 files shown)


gaul:
@ggtakec this fixes a serious regression from 1.91.  I think we can handle CommonPrefixes more elegantly/explicitly but this is a minimal fix.

ggtakec (changes requested):
@gaul Thanks for your fix!

Bit, it seems to be missing a place to set the path to `filled` yet.

With this current PR code, the `ls` command ends up listing the same file multiple times.

I've confirmed that adding the path to `filled` at every call to `filler` fixes this problem.

I made the code in this PR plus the following fixes and confirmed that fixed code works fine.
Could you check the following and fix it?


- Added `filled` to Multi-Head request callback parameter
```
3067:  struct multi_head_callback_param
3067:  {
3067:      void*           buf;
3067:      fuse_fill_dir_t filler;
     +     std::set<std::string>* pfilled;
3067:  };
```

- In Multi-Head request callback function
```
3091:      if(param){
3092:          struct multi_head_callback_param* pcbparam = reinterpret_cast<struct multi_head_callback_param*>(param);
3093:          struct stat st;
3094:          if(StatCache::getStatCacheData()->GetStat(saved_path, &st)){
3095:              pcbparam->filler(pcbparam->buf, bpath.c_str(), &st, 0);
3096:          }else{
3097:              S3FS_PRN_INFO2("Could not find %s file in stat cache.", saved_path.c_str());
3098:              pcbparam->filler(pcbparam->buf, bpath.c_str(), 0, 0);
3099:          }
     +         pcbparam->pfilled->insert(bpath);
3100:      }else{
3101:          S3FS_PRN_WARN("param(multi_head_callback_param*) is NULL, then can not call filler.");
3102:      }
```

- Set `filled` pointer to Multi-Head callback parameter
```
3186:      struct multi_head_callback_param success_param;
3187:      success_param.buf    = buf;
3188:      success_param.filler = filler;
     +     success_param.pfilled = &filled;
3189:      curlmulti.SetSuccessCallbackParam(reinterpret_cast<void*>(&success_param));
```

- The `filled` calling is leaking
```
3291:                  if(StatCache::getStatCacheData()->AddStat(dirpath, dummy_header, true)){    // set forcedir=true
3292:                      // Get stats from stats cache(for converting from meta), and fill
3293:                      std::string base_path = mybasename(dirpath);
3294:                      if(use_wtf8){
3295:                          base_path = s3fs_wtf8_decode(base_path);
3296:                      }
3297:  
3298:                      struct stat st;
3299:                      if(StatCache::getStatCacheData()->GetStat(dirpath, &st)){
3300:                          filler(buf, base_path.c_str(), &st, 0);
3301:                      }else{
3302:                          S3FS_PRN_INFO2("Could not find %s directory(no dir object) in stat cache.", dirpath.c_str());
3303:                          filler(buf, base_path.c_str(), 0, 0);
3304:                      }
     +                     filled.insert(base_path);
3305:                  }else{
3306:                      S3FS_PRN_ERR("failed adding stat cache [path=%s], but dontinue...", dirpath.c_str());
3307:                  }
```

Once this fix is done, I'll make a PR which is fixed possible the Thread data race soon.

gaul:
> Bit, it seems to be missing a place to set the path to `filled` yet.

Done.

ggtakec (approved):
Thnaks for fix.

gaul:
This symptom is gaul/s3proxy#303 which the next version will address.

ggtakec:
@gaul Thanks, wait for next version.

PR #2215, opened 2023-07-13
Guard filler calls with filled check

gaul:
Follows on to e650d8c55cc86d2bbaf2ff060761785e46e7a406.


ggtakec:
This fix will be reflected in #2214.

PR #2224, opened 2023-07-19
Specify C++03 when configuring

gaul:
This prevents a std::auto_ptr warning about not using std::unique_ptr. Remove similar setting from CI configuration.  References #2177.


gaul:
Abandon this and we can fix it with C++11 in the next version.

PR #2227, opened 2023-07-22
Experiment with minio

gaul:
This allows testing SSE which S3Proxy does not.

TODO:
* CentOS 7 CI fails since it does not support the openssl `-addext` flag
* `-o update_parent_dir_stat` fails due to `XMinioInvalidObjectName`
* `test_time_mountpoint fails`
* better way to free space from minio?


gaul:
Fails due to out of disk space.  minio has some kind of asynchronous GC that cannot keep up with writes and deletes.

gaul:
I'd like some way to test SSE but there doesn't seem like a path forward here due to minio's GC.  Maybe if there were some compressed FUSE filesystem underneath it since the tests write highly-compressible data?

PR #2236, opened 2023-07-27
Use C++11 std::map::erase return value

gaul:


ggtakec (approved):
C++11 makes it easier to see.:)

PR #2237, opened 2023-07-27
Address stray warnings

gaul:


src/curl.cpp
```diff
@@ -1949,10 +1949,13 @@ bool S3fsCurl::ResetHandle(AutoLock::Type locktype)
     if(CURLE_OK != curl_easy_setopt(hCurl, CURLOPT_NOPROGRESS, 0)){
         return false;
     }
-    // [NOTE]
-    // CURLOPT_PROGRESSFUNCTION should be changed to CURLOPT_XFERINFOFUNCTION,
-    // but CnetOS7's curl version is old, so it haven't been changed yet.
-    if(CURLE_OK != curl_easy_setopt(hCurl, CURLOPT_PROGRESSFUNCTION, S3fsCurl::CurlProgress)){
+    if(CURLE_OK != curl_easy_setopt(hCurl,
```
ggtakec, line 1952: I would like to avoid defining this ifdef statement inline.
```
#if LIBCURL_VERSION_NUM >= 0x073100
         CURLOPT_XFERINFOFUNCTION
#else
         CURLOPT_PROGRESSFUNCTION
#endif
```

How about defining another symbol at the beginning of the file as shown below and using that symbol?
```
#if LIBCURL_VERSION_NUM >= 0x073100
#define S3FS_CURLOPT_XFERINFOFUNCTION   CURLOPT_XFERINFOFUNCTION
#else
#define S3FS_CURLOPT_XFERINFOFUNCTION   CURLOPT_PROGRESSFUNCTION
#endif
```
I prefer this one.

test/junk_data.cc
```diff
@@ -28,12 +28,10 @@ int main(int argc, const char *argv[])
     if (argc != 2) {
         return 1;
     }
-    long long count = strtoull(argv[1], NULL, 10);
+    unsigned long long count = strtoull(argv[1], NULL, 10);
```
ggtakec, line 31: Should the NULL here be changed to nullptr?

ggtakec (commented):
I made two comments. Please confirm them.

src/curl.cpp
gaul: Done.

test/junk_data.cc
gaul: Done.  I'll fix the clang-tidy target in another PR.

ggtakec (approved):
Thanks!

PR #2242, opened 2023-07-30
Convert fixed-size allocations to C++11 std::array

gaul:
This is safer and more efficient.

### DIFF
--- a/src/common_auth.cpp
+++ b/src/common_auth.cpp
@@ -29,18 +29,16 @@
 //-------------------------------------------------------------------
 std::string s3fs_get_content_md5(int fd)
 {
-    unsigned char* md5;
+    md5_t md5;
     char* base64;
     std::string Signature;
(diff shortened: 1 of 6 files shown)


ggtakec (commented):
@gaul I would like to discuss a little.
For numbers such as 16/32, I think it would be better to prepare a symbol(define) or leave the meaning of the definition in a comment.
(ex. "16 matches MD5_DIGEST_LENGTH in OpenSSL")
It may be unfounded, but it seems that we will forget and misunderstand it later.(Though I don't think it's going to lead to any problems.)
Please tell me your opinion.

gaul:
How about a `typedef`?

ggtakec:
Thank you.
This typedef made it very clear(and no more worries about the future). :-)

ggtakec (approved):

PR #2247, opened 2023-08-04
Convert most `std::list` to `std::vector`

gaul:
This tends to be more efficient due to fewer allocations.  Also fix `std::sort` comparator which should be strictly less than.


ggtakec (approved):
Thanks, no problem with this fix.

PR #2254, opened 2023-08-06
Fixed string test for s3fs_base64

ggtakec:
#2248

### Details
Enabled to check with empty string instead of nullptr in string test.
(This causes an error on macos, so we have fixed it.)

### DIFF
--- a/src/test_string_util.cpp
+++ b/src/test_string_util.cpp
@@ -20,7 +20,6 @@
 
 #include <cstdlib>
 #include <limits>
-#include <memory>
 #include <stdint.h>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


src/test_string_util.cpp
```diff
@@ -56,11 +56,11 @@ void test_base64()
     unsigned char *buf;
     size_t len;
 
-    ASSERT_STREQUALS(s3fs_base64(nullptr, 0).c_str(), nullptr);
+    ASSERT_STREMPTY(s3fs_base64(nullptr, 0).c_str());
```
gaul, line 59: This is strange because CI should have failed the previous PR.  It fails locally for me.

But the test here is wrong and the diff should be:

```
-    ASSERT_STREQUALS(s3fs_base64(nullptr, 0).c_str(), nullptr);
-    buf = s3fs_decode64(nullptr, 0, &len);
-    ASSERT_BUFEQUALS(reinterpret_cast<const char *>(buf), len, nullptr, 0);
-
-    ASSERT_STREQUALS(s3fs_base64(reinterpret_cast<const unsigned char *>(""), 0).c_str(), nullptr);
+    ASSERT_EQUALS(s3fs_base64(reinterpret_cast<const unsigned char *>(""), 0), std::string(""));
```

The new code path intentionally crashes on `NULL` inputs.

ggtakec: Changed to not use `ASSERT_STREMPTY` and use `ASSERT_EQUALS` instead.
I also changed the contents of `ASSERT_BUFEQUALS` when checking to decode empty data.
(I wonder why these are not following the error in the code before this PR. However, they are errors in my local environment.)

I think there is no problem with this modified code, so please check it.

PR #2255, opened 2023-08-07
Replace more raw pointers with `std::unique_ptr`

gaul:


ggtakec (commented):
Is it unnecessary to recover from the success failure of std::map::emplace?
It's hard to imagine this actually failing if memory isn't an issue, but I'm concerned that there's no handling for when it fails.

gaul:
`emplace` can allocate, for example when resizing a `vector` or adding an entry to a `map`.  The contract with the move call does what we want where either the local variable or the container owns the pointer and so if the allocation fails then the local destructor should run and if the allocation succeeds then the local no longer owns the pointer.  So this change makes the code more safe.

Note that s3fs currently compiles with `-fno-exceptions` so when `operator new` fails then the `std::bad_alloc` exception will abort the program when unwinding the function.  If we continue to make the code exception-safe via `std::unique_ptr` and similar techniques, we could add a top-level exception handler for every FUSE operation.  This would allow returning `ENOMEM` to the caller which could have a better semantics for applications.  They could gracefully fail instead of the aborted s3fs creating the "endpoint not connected" error.

However I don't think this situation is worth addressing at this time.  Memory corruption has traditionally been more common than out of memory situations for s3fs.  Making fail fast for allocation failures is simple and safe and what some other programs do.

ggtakec (approved):

PR #2256, opened 2023-08-08
Use `std::string::compare` and `operator==` where possible

gaul:


src/cache.cpp
```diff
@@ -248,7 +248,7 @@ bool StatCache::GetStat(const std::string& key, struct stat* pst, headers_t* met
                     std::string tag = lower(hiter->first);
                     if(tag == "etag"){
                         stretag = hiter->second;
-                        if('\0' != petag[0] && 0 != strcmp(petag, stretag.c_str())){
+                        if('\0' != petag[0] && petag != stretag){
```
ggtakec, line 251: By the way, why don't you make the following changes?
`'\0' != petag[0]` -> `!petag.empty()`

ggtakec (commented):
Please check a comment.

src/cache.cpp
gaul: `petag` is a `char*`.

ggtakec: Sorry for my fault.

ggtakec (approved):

PR #2257, opened 2023-08-08
Delete copy constructors and assignment operators

gaul:
One of these was buggy and others had the wrong parameters and return types.


ggtakec (approved):

ggtakec:
@gaul Thanks!

PR #2260, opened 2023-08-12
Fixed test_not_existed_dir_obj and test_update_time_touch

ggtakec:
n/a

### Details
In `ALPINE`, I confirmed that `test_not_existed_dir_obj` and `test_update_time_touch` occasionally failed.
The reason for this is that it may be affected by cache (although it is not clear), so I added sleep to the relevant part and added 1 second (time until cache out).
Also, the indentation was wrong, so I'm fixing them.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -994,6 +994,7 @@ function test_update_time_touch() {
     local t0=1000000000  # 9 September 2001
     local OBJECT_NAME; OBJECT_NAME=$(basename "${PWD}")/"${TEST_TEXT_FILE}"
     echo data | aws_cli s3 cp --metadata="atime=${t0},ctime=${t0},mtime=${t0}" - "s3://${TEST_BUCKET_1}/${OBJECT_NAME}"
+    sleep 1
     local base_atime; base_atime=$(get_atime "${TEST_TEXT_FILE}")
     local base_ctime; base_ctime=$(get_ctime "${TEST_TEXT_FILE}")
     local base_mtime; base_mtime=$(get_mtime "${TEST_TEXT_FILE}")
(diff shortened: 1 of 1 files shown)


ggtakec:
This PR is still not enough, so this is going back to Draft.

ggtakec:
@gaul This pull request is change to ready for review, please review this.

test/integration-test-main.sh
```diff
@@ -2255,59 +2263,58 @@ function test_not_existed_dir_obj() {
     echo data2 | aws_cli s3 cp - "s3://${TEST_BUCKET_1}/${OBJECT_NAME_2}"
 
     # Top directory
-    # shellcheck disable=SC2010
-    if ! ls -1 | grep -q '^not_existed_dir_single$'; then
-    echo "Expect to find \"not_existed_dir_single\" directory, but it is not found"
-    return 1;
+    local _LISTUP_OBJNAMES; _LISTUP_OBJNAMES="$(ls -1)"
```
gaul, line 2266: You don't need `ls -1` here -- since the terminal is non-interactive `ls` will be fine.

test/integration-test-main.sh
```diff
@@ -2246,6 +2247,13 @@ function test_not_existed_dir_obj() {
 
     local DIR_NAME; DIR_NAME=$(basename "${PWD}")
 
+    #
+    # Remove directories before this test
+    #
+    rm -rf not_existed_dir_single
+    rm -rf not_existed_dir_parent
+    sleep 1
```
gaul, line 2255: I'm not sure that I understand this -- does a previous test run interfere with this function?

ggtakec: I've run a few tests and have seen occasional cases where this directory is detected. (Others are unknown, but detected only by ALPINE)
Overall, I'm not sure why it happens (rarely), but definitely deleting the directory and waiting 1 second (stat cache timeout) made this case disappear.

ggtakec: @gaul Please give me your opinion on this PR. I would like to merge if possible.
That is, this may be a temporary fix, but it avoids the recent frequent CI test failures with ALPINE.

gaul: I don't think I understand this -- should we use a different test prefix for every test instead?  For example, we could create a new bucket or subdirectory instead.  I don't think any test should clean up _before_ it runs.  Rather the test harness should remove any state _after_ each test.

Also I generally distrust most of these sleeping tests which hide some other problem.

test/integration-test-main.sh
```diff
@@ -2243,59 +2251,58 @@ function test_not_existed_dir_obj() {
     echo data2 | aws_cli s3 cp - "s3://${TEST_BUCKET_1}/${OBJECT_NAME_2}"
 
     # Top directory
-    # shellcheck disable=SC2010
-    if ! ls -1 | grep -q '^not_existed_dir_single$'; then
-    echo "Expect to find \"not_existed_dir_single\" directory, but it is not found"
-    return 1;
+    local _LISTUP_OBJNAMES; _LISTUP_OBJNAMES="$(ls)"
```
gaul, line 2254: Is this a different change?  `ls -1` is not meaningful in non-interactive scripts.

Would `find -mindepth 1` address this more cleanly?  Happy to accept this as a separate PR.

test/integration-test-main.sh
ggtakec: Let's find out a little more. (The same error did not occur in ALPINE today, so we will see how it goes.)
The sleep is inserted to avoid noobj_cache(1s), which is necessary due to the deletion of this directory.

ggtakec: It was not correct modified code to delete the directory here and run sleep 1.
So I will remove this process.

The error that should be fixed in this PR was due to a 404 error being returned when listing a directory for a file containing a directory path created by an aws command.
I will update the code with this fix.

Since the code of this part to be PR will be significantly changed (deleted), I will mark this as solved for now.

test/integration-test-main.sh
ggtakec: First, I will replace all occurrences of `ls -1` with `ls`.

And, previously I used the results of a one-time `ls` command to check for `not_existed_dir_single` and `not_existed_dir_parent` directories.
but I will fix this by calling an `ls` command for each directory.

Since the code of this part to be PR will be significantly changed (deleted), I will mark this as solved for now.

ggtakec:
@gaul Please re-review again.

After uploading files containing directories created by an external program (aws command), I was able to confirm the cause of the error when listing those directories.

I have confirmed that there are cases where a `404` error is received from the `S3 server` when listing a directory(file) created externally.

It has nothing to do with the s3fs logic, and avoiding it in the test script should not affect the testing of the s3fs logic.
To avoid this, when a `404` error(directory not found) occurs, it retries 3 times.(interval is 1 second)

This retry confirms that the listing for this directory goes from a `404` error to a `200`(found the directory).

test/integration-test-main.sh
```diff
@@ -2243,59 +2244,76 @@ function test_not_existed_dir_obj() {
     echo data2 | aws_cli s3 cp - "s3://${TEST_BUCKET_1}/${OBJECT_NAME_2}"
 
     # Top directory
+    #
+    # [NOTE]
+    # Immediately after uploading with the aws command, there is a case where
+    # the directory path cannot be found (404 error).
+    # In this case, it may be listed correctly after a few retries.
+    # This error is a response problem on the S3 server side, not a problem
+    # with the s3fs process, so we try again.
```
gaul, line 2253: I don't understand this comment -- does S3Proxy have a bug where directories are not immediately exposed?  If so I would like to understand this more so I can address that issue.

ggtakec: The cause of this is still unknown, but I determined that it was a response problem to the API request, and tried to solve it with this loop.

I collected the logs when this test fails (it does fail occasionally) and the logs recorded a 404 error.
And I've tried a number of workarounds, but if I end up receiving a 404 error, I've seen it become 200 if I repeat the request multiple times.
A successful re-request after this interval may be #2319. (I may have made a mistake in the analysis of the 404 error above)

So we need to check again once #2319 is merged.

ggtakec: This problem seems to persist regardless of the stat `--cached=never` option.

So I ran CI again with `alpine:3.17`, only running the `test_not_existed_dir_obj` test, and with `curldbg` option.
As a result of this execution, I got an error occurs in the `(*)` part below.
```

    #
    # Create files under not existed directory by aws command
    #
    local OBJECT_NAME_1; OBJECT_NAME_1="${DIR_NAME}/not_existed_dir_single/${TEST_TEXT_FILE}"
    local OBJECT_NAME_2; OBJECT_NAME_2="${DIR_NAME}/not_existed_dir_parent/not_existed_dir_child/${TEST_TEXT_FILE}"
    echo data1 | aws_cli s3 cp - "s3://${TEST_BUCKET_1}/${OBJECT_NAME_1}"
    echo data2 | aws_cli s3 cp - "s3://${TEST_BUCKET_1}/${OBJECT_NAME_2}"

    # Top directory
    # shellcheck disable=SC2010
    if ! ls -1 | grep -q '^not_existed_dir_single$'; then
        echo "Expect to find \"not_existed_dir_single\" directory, but it is not found"
        return 1;
    fi
    # shellcheck disable=SC2010
(*) if ! ls -1 | grep -q '^not_existed_dir_parent$'; then
        echo "Expect to find \"not_existed_dir_parent\" directory, but it is not found"
        return 1;
    fi
```
You can see the execution log at this time:
https://github.com/ggtakec/s3fs-fuse/actions/runs/6290707707

ggtakec:
This PR is being withdrawn as similar test failures are no longer occurring on Alpine.
Please note that only the indentation correction included in this PR has been submitted as #2368.

Issue #2261, opened 2023-08-13
Use explicit ownership for all resources


gaul:
C++11 enables s3fs to manage resources explicitly, e.g., via `std::unique_ptr` and `std::move`, instead of ad-hoc management via `delete`.  This can simplify error handling and we can eliminate most or maybe all calls to `delete`.

We should be able to write some wrappers for `FILE*` and file descriptors as well.

PR #2262, opened 2023-08-14
Simplify xattr_value with owned values

gaul:
References #2261.


ggtakec (approved):
There is no problem with the modified content.
I don't think this to be the cause of the xattr-related problem in #2198, but I'll check it after merging.

PR #2272, opened 2023-08-16
Return `std::unique_ptr` from `S3fsCurl` callbacks

gaul:
References #2261.


gaul:
This fails due to an unmatched suppression that older cppcheck versions do not support:

```
        <error id="unmatchedSuppression" severity="information" msg="Unmatched suppression: mismatchingContainers" verbose="Unmatched suppression: mismatchingContainers">
```

ggtakec (approved):
Mergers are fine.
Just tell me why you restricted to cppcheck 2.04 or higher.

ggtakec:
> I restricted this since older cppcheck do not allow the suppression above.

Thanks, understood.

PR #2274, opened 2023-08-16
s3fs: print fuse context in s3fs fuse operations

eryugey:
Print fuse context like pid in fuse operations, so we know which process is triggering this operation.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1033,7 +1033,7 @@ static int s3fs_getattr(const char* _path, struct stat* stbuf)
     WTF8_ENCODE(path)
     int result;
 
-    S3FS_PRN_INFO("[path=%s]", path);
+    FUSE_CTX_INFO("[path=%s]", path);
 
     // check parent directory attribute.
     if(0 != (result = check_parent_object_access(path, X_OK))){
@@ -1070,6 +1070,7 @@ static int s3fs_readlink(const char* _path, char* buf, size_t size)
     }
     WTF8_ENCODE(path)
     std::string strValue;
+    FUSE_CTX_INFO("[path=%s]", path);
 
     // check symbolic link cache
     if(!StatCache::getStatCacheData()->GetSymlink(path, strValue)){
@@ -1147,7 +1148,7 @@ static int s3fs_mknod(const char *_path, mode_t mode, dev_t rdev)
     int       result;
     struct fuse_context* pcxt;
 
-    S3FS_PRN_INFO("[path=%s][mode=%04o][dev=%llu]", path, mode, (unsigned long long)rdev);
+    FUSE_CTX_INFO("[path=%s][mode=%04o][dev=%llu]", path, mode, (unsigned long long)rdev);
 
     if(nullptr == (pcxt = fuse_get_context())){
         return -EIO;
@@ -1176,7 +1177,7 @@ static int s3fs_create(const char* _path, mode_t mode, struct fuse_file_info* fi
     int result;
     struct fuse_context* pcxt;
 
-    S3FS_PRN_INFO("[path=%s][mode=%04o][flags=0x%x]", path, mode, fi->flags);
+    FUSE_CTX_INFO("[path=%s][mode=%04o][flags=0x%x]", path, mode, fi->flags);
 
     if(nullptr == (pcxt = fuse_get_context())){
         return -EIO;
@@ -1274,7 +1275,7 @@ static int s3fs_mkdir(const char* _path, mode_t mode)
     int result;
     struct fuse_context* pcxt;
 
-    S3FS_PRN_INFO("[path=%s][mode=%04o]", path, mode);
+    FUSE_CTX_INFO("[path=%s][mode=%04o]", path, mode);
 
     if(nullptr == (pcxt = fuse_get_context())){
         return -EIO;
@@ -1321,7 +1322,7 @@ static int s3fs_unlink(const char* _path)
     WTF8_ENCODE(path)
     int result;
 
-    S3FS_PRN_INFO("[path=%s]", path);
+    FUSE_CTX_INFO("[path=%s]", path);
 
     if(0 != (result = check_parent_object_access(path, W_OK | X_OK))){
         return result;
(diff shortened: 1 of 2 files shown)


gaul (approved):

src/s3fs.cpp
```diff
@@ -59,6 +59,31 @@
 #define ENOATTR                   ENODATA
 #endif
 
+#define PRINT_FUSE_CTX(level, indent, fmt, ...) do {                \
+    struct fuse_context *ctx = fuse_get_context();                  \
```
ggtakec, line 63: Could you change this condition about `if(S3fsLog::IsS3fsLogInfo()){...}` instead of `do{...}while(0);`?
This avoids unnecessary calls to `fuse_get_context()` when not at `INFO` level.

src/s3fs.cpp
```diff
@@ -59,6 +59,31 @@
 #define ENOATTR                   ENODATA
 #endif
 
+#define PRINT_FUSE_CTX(level, indent, fmt, ...) do {                \
```
ggtakec, line 62: Could this macro be written in `s3fs_logger.h`?
I think it's easier to understand.

ggtakec (changes requested):
The purpose of this fix is fine, but could you please check the commented part?

src/s3fs.cpp
guaneryu: Sorry, I didn't quite get it, fuse_get_context() is always necessary, no matter what the log level is. Would you please give more details?

src/s3fs.cpp
guaneryu: I thought about it too, but I decided to put the macros in s3fs.cpp instead of s3fs_logger.h, as these macros are only used in s3fs.cpp. But I don't have strong preference, I can move them to s3fs_logger.h if you confirm.

ggtakec: I can understand your thoughts and judgments. (me too)
I think that since this is a macro for outputting messages according to the debug log level, it seems easier to understand if they are grouped in s3fs_logger.h.
It would be helpful if you could move the macro description.

src/s3fs.cpp
ggtakec: `S3FS_LOW_LOGPRN2` calling from `PRINT_FUSE_CTX` determines the log level in that macro (that is the `s3fs_low_logprn2` function) and works.
However, the `fuse_get_context` function that gets called is always executed every time `PRINT_FUSE_CTX` is called, regardless of log level.

If possible, I also wanted to avoid calling this function depending on the log level.
Therefore, I think it is better to judge the log level at the beginning of the `PRINT_FUSE_CTX` macro and not call the `fuse_get_context` function if it is not necessary.
It may not cost much, but I would like to avoid unnecessary calls (which end up calling `pthread` functions) if possible.

guaneryu: I get it now, when current log level is not debug, s3fs_low_logprn2() won't try to print anything, so the fuse_get_context() call is wasted. Will update.

src/s3fs_logger.h
```diff
@@ -230,6 +230,34 @@ void s3fs_low_logprn2(S3fsLog::s3fs_log_level level, int nest, const char* file,
 #define S3FS_PRN_CURL(fmt, ...)   S3FS_LOW_CURLDBG(fmt, ##__VA_ARGS__, "")
 #define S3FS_PRN_CACHE(fp, ...)   S3FS_LOW_CACHE(fp, ##__VA_ARGS__, "")
 
+// Macros to print log with fuse context
+#define PRINT_FUSE_CTX(level, indent, fmt, ...) do {                \
```
ggtakec, line 234: Could you please replace the `do{...}while(0)` or enclose all the processing including the `fuse_get_context()` call with the following conditional statement inside it?
```
if(S3fsLog::IsS3fsLogLevel(level)){...}
```
This prevents `fuse_get_context` from being called if the debug levels do not match.

ggtakec (changes requested):
Thank you for implementing the change request.
Is it possible to fix the only remaining one?

src/s3fs_logger.h
eryugey: I did a similar check in FUSE_CTX_DBG, does that work for you? If you prefer I could update PRINT_FUSE_CTX as you suggested.

ggtakec: @eryugey I'm sorry for my late reply.

I believe your code will work fine.
However, what I would like to fix is:

I would like `PRINT_FUSE_CTX` to determine the debug level instead of individual `FUSE_CTX_XXXX` macros.
I've also tried the code below and it seems to work fine.
```
#define PRINT_FUSE_CTX(level, indent, fmt, ...) do {                    \
    if(S3fsLog::IsS3fsLogLevel(level)){                                 \
        struct fuse_context *ctx = fuse_get_context();                  \
        if(ctx == NULL){                                                \
            S3FS_LOW_LOGPRN2(level, indent, fmt, ##__VA_ARGS__);        \
        }else{                                                          \
            S3FS_LOW_LOGPRN2(level, indent, fmt"[pid=%u,uid=%u,gid=%u]",\
                ##__VA_ARGS__,                                          \
                (unsigned int)(ctx->pid),                               \
                (unsigned int)(ctx->uid),                               \
                (unsigned int)(ctx->gid));                              \
        }                                                               \
    }                                                                   \
} while (0)

#define FUSE_CTX_INFO(fmt, ...) do {                            \
    PRINT_FUSE_CTX(S3fsLog::LEVEL_INFO, 0, fmt, ##__VA_ARGS__); \
} while (0)

#define FUSE_CTX_INFO1(fmt, ...) do {                           \
    PRINT_FUSE_CTX(S3fsLog::LEVEL_INFO, 1, fmt, ##__VA_ARGS__); \
} while (0)

#define FUSE_CTX_DBG(fmt, ...) do {                             \
    PRINT_FUSE_CTX(S3fsLog::LEVEL_DBG, 0, fmt, ##__VA_ARGS__);  \
} while (0)
```

eryugey: Sure, I'll update as you suggested.

ggtakec: @eryugey Thank you for your time. I confirmed the fix.

ggtakec (approved):

ggtakec:
@eryugey Thanks for your contribution!

PR #2276, opened 2023-08-17
Convert `BodyData` to `std::string`

gaul:
This is simpler and avoids some copies.


ggtakec (approved):
Thanks, as per this fix, it is correct to let std:string manage memory.

PR #2277, opened 2023-08-17
Test filenames longer than POSIX maximum

gaul:


test/integration-test-main.sh
```diff
@@ -21,6 +21,7 @@
 
 set -o errexit
 set -o pipefail
+set -o errtrace
```
ggtakec, line 24: Thanks, we forgot to set this.

ggtakec (approved):

ggtakec:
@gaul It looks like it conflicted with another PR in master. Please rebase and fix code if needs.

test/integration-test-main.sh
gaul: Sorry I removed this lets add it in a separate PR if necessary.

PR #2280, opened 2023-08-18
Replace `xattr_value` with `std::string`

gaul:


ggtakec (approved):
I was using the `xattr_value` structure for setting the xattr value, considering the value could be binary(ex. a value containing '\0').
But that was an overkill.:-(
The value is string(`const char*`) when the s3fs handler is called from FUSE, so this fix is correct and removes complexity.

Thanks, this will help with the `fuse-t` support I am currently investigating.

gaul:
`std::string` supports NULs since it has an explicit `length` method.

ggtakec:
@gaul I found a little problem with emplace handling, and fixed this issue and created PR #2283.

PR #2283, opened 2023-08-19
Added std::map erase before calling emplace in xattrs operations

ggtakec:
#2280

### Details
`std::map emplace` will not replace the value if that key exists.
Fixed xattrs handling to call `erase` if there is a target key.

I also added a test case for the missing xattrs inspection related to this.

Check where emplace is used in the current source code, and there is no problem except for xattrs.
(In the `stat cache` process, the keys are deleted in advance, so I don't think there will be any problems.)

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -3739,7 +3739,7 @@ static size_t parse_xattrs(const std::string& strxattrs, xattrs_t& xattrs)
             // something format error, so skip this.
             continue;
         }
-        xattrs.emplace(std::move(key), std::move(val));
+        xattrs[key] = val;
     }
     return xattrs.size();
 }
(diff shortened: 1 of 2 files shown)


src/s3fs.cpp
```diff
@@ -3739,6 +3739,9 @@ static size_t parse_xattrs(const std::string& strxattrs, xattrs_t& xattrs)
             // something format error, so skip this.
             continue;
         }
+        if(xattrs.end() != xattrs.find(key)){
+            xattrs.erase(key);
+        }
         xattrs.emplace(std::move(key), std::move(val));
```
gaul, line 3745: Oh, I didn't realize this.  Maybe `xattrs[std::move(key] = std::move(val)` is a better replacement?

ggtakec: Yes, I think you can overwrite it with array access.
But in that case the constructor may have been called. (unconfirmed)

gaul: Right.  I think this is fine for the case of `std::string` due to small string optimization but `insert_or_assign` is the best way to do this in C++17.  But in any case we should prefer readability over efficiency.

ggtakec: The cost efficiency of moving this std::string is trivial, so let's go with array access.
I will fix and update.

test/integration-test-main.sh
```diff
@@ -815,6 +815,10 @@ function test_extended_attributes {
     touch "${TEST_TEXT_FILE}"
 
     # set value
+    set_xattr key1 value0 "${TEST_TEXT_FILE}"
+    get_xattr key1 "${TEST_TEXT_FILE}" | grep -q '^value0$'
+
+    # over write value
```
gaul, line 821: Thanks for the test!

gaul (approved):

PR #2291, opened 2023-08-26
Remove temporary files in test harness

gaul:
Previously removing files in tests did not remove all of the litter, particularly during failed test runs.


gaul:
Some strange errors:

```
[s3proxy] E 08-26 01:45:11.530 S3Proxy-Jetty-55 o.j.b.config.LocalBlobStore:93 |::] An error occurred storing the new blob with name [.mpus-b85e5dbb-ab32-483b-80e5-3d6638139aa2-testrun-9493/test_skipwrite-10] to container [s3fs-integration-test].
```

gaul:
I think the S3Proxy in-memory bucket is filling up and causing GC/OOM issues.

PR #2292, opened 2023-08-26
Skip `is_uid_include_group` when GID available

gaul:
This can avoid an expensive computation which is 20% of test runtime.


ggtakec (approved):
Thanks!
The original logic processed every time and was inefficient.

PR #2293, opened 2023-08-26
Run tests with HTTP instead of HTTPS

gaul:
This reduces the s3fs CPU run-time dramatically.

### DIFF
--- a/test/integration-test-common.sh
+++ b/test/integration-test-common.sh
@@ -72,11 +72,7 @@ S3FS=../src/s3fs
 # [NOTE]
 # CHAOS HTTP PROXY does not support HTTPS.
 #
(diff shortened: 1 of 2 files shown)


gaul:
This doesn't seem to improve test run time so I don't think we should merge this.

ggtakec:
It is possible to test with HTTP, but I would like to prioritize testing with HTTPS if possible.
This is because the servers that provide the S3 API are likely to have few HTTP types.

Digression:
Separately from this matter, in the macos local test, there are cases where the self-signed certificate error cannot be resolved, and the NO_CHECK_CERT flag is prepared.
(I can't test without granting this in my local macos test)

gaul:
I think the speedup I observed was only with Valgrind which is single-threaded.  Usually there are many CPU cores so the SSL overhead is smaller.  I will close this since it is not generally useful but we may want to use HTTP for the Valgrind CI to reduce run-time.

Issue #2176, opened 2023-05-30
Transport endpoint is not connected


dbartenstein:
#### Version of s3fs being used (`s3fs --version`)
V1.89 (commit:unknown)

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)
2.9.9-5

#### Kernel information (`uname -r`)
5.10.0-23-cloud-amd64

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)
Debian GNU/Linux 11 (bullseye)

#### How to run s3fs, if applicable

[x] /etc/fstab  

```
s3fs#prod.media.zemtu   /var/www/zemtu/public/media     fuse    _netdev,allow_other,use_path_request_style,use_cache=/tmp,uid=1001,gid=1002,umask=0022,url=https://sos-ch-dk-2.exo.io/,passwd_file=/root/.passwd-s3fs,comment=cloudconfig  0       0
```

#### s3fs syslog messages (`grep s3fs /var/log/syslog`, `journalctl | grep s3fs`, or `s3fs outputs`)

```
syslog
[Errno 107] Transport endpoint is not connected

May 29 09:16:58 prod-main-zemtu s3fs[2017306]: s3fs: unable to access MOUNTPOINT /var/www/zemtu/public/media: Transport endpoint is not connected
```

### Details about issue
After two weeks without any issues, the mountpoint suddenly stopped working. We had to execute:
```
sudo fusermount -uz /var/www/zemtu/public/media
sudo mount -a
```

I already commented another similar issue: https://github.com/s3fs-fuse/s3fs-fuse/issues/725#issuecomment-1568047863
Hint: we’re using [Exoscale](https://www.exoscale.com/), not AWS.

An option I came across is `nodnscache (disable DNS cache)` - what exactly happens when the dnscache is enabled (default)? Will s3fs pick up DNS changes automatically, e.g. when the URL suddenly points to a CNAME instead of an A record?

Thanks a lot!

dbartenstein:
Update: we have just upgraded to s3fs 1.92 and will continue monitoring the situation.

But we still wonder what` nodnscache (disable DNS cache)` exactly does. For how long is DNS cached when the option is not used? Thanks!

ggtakec:
@dbartenstein Thanks for testing with the latest code(v1.92).
```
[...]

Issue #2092, opened 2023-01-11
URL Changes causing Transport Endpoint Disconnected Error.


DicklePick:
#### Version of s3fs being used (`s3fs --version`)
V1.89 (commit:unknown)

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)
2.9.9

#### Kernel information (`uname -r`)
5.10.0-20-amd64

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)
Debian GNU/Linux 11 (bullseye)

#### How to run s3fs, if applicable

[] command line  
[X] /etc/fstab  

s3fs#ci-live-mountedfiles /ci-live-mountedfiles/ fuse _netdev,allow_other,passwd_file=/etc/.passwd-s3fs,umask=000,uid=33,gid=33,use_cache=/root/cache,dbglevel=debug,logfile=/var/log/s3fs.log 0 0

#### s3fs syslog messages (`grep s3fs /var/log/syslog`, `journalctl | grep s3fs`, or `s3fs outputs`)

[s3fs.log](https://github.com/s3fs-fuse/s3fs-fuse/files/10389716/s3fs.log)

### Details about issue
On a weekly basis, sometimes twice a week my file system gets unmounted and i receive an error that the transport endpoint is not connected. a server reboot resolves the issue until it happens again. For the most part it all works well for a decent amount of time.
I notice this alot in the logfile:
curl_util.cpp:url_to_host(327): url is https://s3-eu-central-1.amazonaws.com
curl.cpp:RequestPerform(2329): HTTP response code 404 was returned, returning ENOENT
AND
URL is https://s3-eu-central-1.amazonaws.com/ci-live-mountedfiles/documents/e3516443-f73d-43e5-9c58-bb20f10cdcd6/applications/4b29bbbe-4379-4458-83cb-29b369d415c5/CV_2022-12-05-095108_pdf.pdf_%24folder%24
curl_util.cpp:prepare_url(283): URL changed is https://ci-live-mountedfiles.s3-eu-central-1.amazonaws.com/documents/e3516443-f73d-43e5-9c58-bb20f10cdcd6/applications/4b29bbbe-4379-4458-83cb-29b369d415c5/CV_2022-12-05-095108_pdf.pdf_%24folder%24

ggtakec:
@DicklePick 
Do you know what you are doing when this error occurs?
[...]

PR #2295, opened 2023-08-28
s3fs: print unmounting hint when the mount point is stale 

adamqqqplay:
### Details
When situations such as out of memory (OOM) occur, s3fs may receive a SIGKILL signal and exit before cleaning up the mountpoint. At this time, a stale mount point will be left, causing an error to be reported the next time s3fs is mounted: Transport endpoint is not connected.

This patch adds a feature: before s3fs mount, check if the stat information of the mountpoint returns ENOTCONN error, if so, it will execute the umount operation to unmount the stale mountpoint.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -146,6 +146,8 @@ static bool set_mountpoint_attribute(struct stat& mpst);
 static int set_bucket(const char* arg);
 static int my_fuse_opt_proc(void* data, const char* arg, int key, struct fuse_args* outargs);
 static fsblkcnt_t parse_bucket_size(char* value);
+static bool is_cmd_exists(const std::string& command);
+static int print_umount_message(const std::string& mp, bool force);
 
 //-------------------------------------------------------------------
 // fuse interface functions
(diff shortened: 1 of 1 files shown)


src/s3fs.cpp
```diff
@@ -4668,6 +4703,11 @@ static int my_fuse_opt_proc(void* data, const char* arg, int key, struct fuse_ar
             memset(&stbuf, 0, sizeof stbuf);
             set_mountpoint_attribute(stbuf);
 #else
+            // check stale mountpoint
+            if(stat(arg, &stbuf) == -1 && errno == ENOTCONN){
+                S3FS_PRN_WARN("MOUNTPOINT: directory %s is stale, try to unmount it.", mountpoint.c_str());
+                do_umount(mountpoint, true);
```
ggtakec, line 4709: I think it should check the return value of this function, and if it fails, call `S3FS_PRN_EXIT` and `return -1`.

gaul:
Do any other filesystems have this magic behavior?  If not, why should s3fs do something different?  Does this problem really need to be solved?

Instead of remounting the filesystem, could s3fs respond to memory pressure by reducing its cache size or using fewer worker threads?

ggtakec:
@gaul Thanks.
Indeed, if the mount point was already mounted, it is impossible to tell whether it was due to s3fs' end processing missing.
Then, on my second thought umounting it from the s3fs start processing is not a good thing to do.
For example, you can think of the case where it is mounted by another program other than s3fs.
I don't think this process should be done at s3fs startup (even if it's an issue with s3fs' finalization).
After all, this umount processing should be done on the calling side of s3fs if necessary.

src/s3fs.cpp
adamqqqplay: Thanks, I fixed it.

adamqqqplay:
> Thanks for the PR. I think this feature is necessary, so it is good change code.
> 
> There are some errors in Github Actions, so please fix them. Also, I have one comment, so please check it.

Thanks for your review. I fixed the Github Actions errors and your comments, please take a look.

ggtakec:
I confirmed your fix. (I think you've fixed all my requests for corrections, but the test is still failing.)

Then, With this added `do_umount` function call, I'd like to discuss whether or not this should be done.

Right now, s3fs doesn't go as far as executing the `umount` operation and exits.(If the mount information already exists, s3fs will fail to start.)
In addition to the current situation, I think that outputting a detailed message is sufficient.
By displaying the command message of `umount` or `fusermount` in the message at that time, we can show the user the procedure to resolve the error.
With this method, it seems that it is possible to respond not excessively.

adamqqqplay:
> I confirmed your fix. (I think you've fixed all my requests for corrections, but the test is still failing.)
> 
> Then, With this added `do_umount` function call, I'd like to discuss whether or not this should be done.
> 
> Right now, s3fs doesn't go as far as executing the `umount` operation and exits.(If the mount information already exists, s3fs will fail to start.) In addition to the current situation, I think that outputting a detailed message is sufficient. By displaying the command message of `umount` or `fusermount` in the message at that time, we can show the user the procedure to resolve the error. With this method, it seems that it is possible to respond not excessively.

Thanks for your suggestion. I think s3fs output a command hint is very good for users, especially now there are many users who don't know how to fix this issue (such as #2176 #2092).

I will push a new patch soon, thanks.

ggtakec (approved):
Thank you. There is no problem with this fix.

ggtakec:
@gaul I merged this PR. If you still have any problems (message content, etc.), please let me know.

PR #2296, opened 2023-08-28
Return errors from AutoFdEntity::Open

gaul:
Found via pjdfstest which creates a `PATH_MAX` path that should return `NAMETOOLONG`.


ggtakec (approved):
Thanks, now errno can be propagated.

PR #2297, opened 2023-08-28
Enable Valgrind in CI

gaul:
Using HTTP instead of HTTPS and
82107f4b6c4cf455c2f3dd63e0ade29f042c1b0f improve test run-time so that this is now feasible.


ggtakec (approved):
I confirmed that Valgrind's check was shortened to about 25 minutes.
I think it's good enough for a CI process.(we wish it was faster)

PR #2301, opened 2023-09-01
Explicitly `grep` for `ps` args

gaul:


gaul:
Older shellcheck appear to have a false positive for this idiom.

ggtakec (commented):
Thank you, this `-o args` option is useful.
By the way, since the current command outputs multiple lines, how about adding `--no-headers` as well?

gaul:
> By the way, since the current command outputs multiple lines, how about adding `--no-headers` as well?

Done.  Not sure if macOS supports this but I abstracted it to a function.

ggtakec (approved):
Thank s for updating.

`--no-headers` option seems that it is not supported by macos.
It looks like I'll be using PIPE instead and extracting with something like `| tail -n +2`.
But I will modify that this will be inspected and changed when the macos will be supported.

PR #2302, opened 2023-09-03
Initialize variable before use

gaul:
clang-analyzer found a path where this could be used without initialization.


ggtakec (approved):
Thanks, certainly there are places that were used without being initialized.

PR #2305, opened 2023-09-04
Fixed a bug with setting the statvfs value

ggtakec:
n/a

### Details
I noticed that the results of the `df` command are strange.
In macos, all block sizes and free block sizes were zero.
Also, the values seemed strange on Linux(using a 64-bit block size).
Revised the `statvfs` setting function(`s3fs_statvfs`) and fixed it so that if the `block_size` option is not specified, the maximum size is used.


gaul (changes requested):
Can we test this somehow via `df`?

gaul:
> Can we test this somehow via `df`?

Or maybe `stat --file-system` should be better.

ggtakec:
Added the following test case.
- Check the s3fs mount point information: The Total and Available sizes will be the same (not 0), and the Used size will always be zero.

I considered the `stat` command, but considering MacOS, I decided to use the `df` command.

PR #2306, opened 2023-09-04
Address some Shellcheck SC2012 warnings

gaul:


test/integration-test-main.sh
```diff
@@ -1984,8 +1984,7 @@ function test_cache_file_stat() {
     # get cache file inode number
     #
     local CACHE_FILE_INODE
-    # shellcheck disable=SC2012
-    CACHE_FILE_INODE=$(ls -i "${CACHE_DIR}/${TEST_BUCKET_1}/${CACHE_TESTRUN_DIR}/${BIG_FILE}" 2>/dev/null | awk '{print $1}')
+    CACHE_FILE_INODE=$(stat --format '%i' "${CACHE_DIR}/${TEST_BUCKET_1}/${CACHE_TESTRUN_DIR}/${BIG_FILE}")
```
ggtakec, line 1987: On macos, it should be `stat -f '%i'`.
It seems better to define the function in test_util.sh.

ggtakec (changes requested):
Although it has not been reverted macos yet, please add a case the i-node number command for macos as well.

test/integration-test-main.sh
gaul: Done.

ggtakec (approved):

PR #2311, opened 2023-09-11
Remove unnecessary uses of `ls`

gaul:
Other call sites need the call to `readdir`/`getdents64`.


ggtakec (approved):
Thank you.
Since `ls` in the modified part specifies the target (file, directory), it was possible to replace it with `stat`.

PR #2313, opened 2023-09-12
Add scope_guard for ad-hoc resource management

gaul:
References #2261.  Suggested by:
https://stackoverflow.com/questions/10270328/the-simplest-and-neatest-c11-scopeguard


ggtakec (approved):
Thank you.
I think it seems scope_guard can be used in other places as well.

ggtakec (changes requested):
Sorry, I didn't look closely at the CI results.
It appears that `test_mtime_file` and `test_update_time_cp_p` are failing on multiple OSes.

#### Additional Information
It doesn't seem to be a problem with this PR.
I'm not sure what is causing these tests to fail, but the error started occurring even before I applied this PR. (The cause has not been identified)
As a result of executing `cp -p`, the value of `a/c/m-time` is different from before.

ggtakec (approved):

PR #2319, opened 2023-09-16
Do not cache `stat` attributes

gaul:
This is a workaround for CI failures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,7 +117,9 @@ jobs:
 
       - name: Shellcheck
         run: |
-          make shellcheck
+          if shellcheck --version | awk -F '[ .]' '/version:/ && ($2 * 1000 + $3 <= 7) { exit(1) }'; then
+            make shellcheck
+          fi
 
       - name: Test suite
         run: |
@@ -182,7 +184,9 @@ jobs:
 
       - name: Shellcheck
         run: |
-          make shellcheck
+          if shellcheck --version | awk -F '[ .]' '/version:/ && ($2 * 1000 + $3 <= 7) { exit(1) }'; then
+            make shellcheck
+          fi
 
       - name: Test suite
         run: |
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -2663,7 +2663,10 @@ function add_all_tests {
     add_tests test_truncate_cache
     add_tests test_upload_sparsefile
     add_tests test_mix_upload_entities
-    add_tests test_not_existed_dir_obj
+    # TODO: investigate why only Alpine cannot see the implicit directory objects.
+    if grep -q -i -e 'ID="alpine"' /etc/os-release; then
+        add_tests test_not_existed_dir_obj
+    fi
     add_tests test_ut_ossfs
     add_tests test_cr_filename
     if ! s3fs_args | grep -q ensure_diskfree && ! uname | grep -q Darwin; then
--- a/test/test-utils.sh
+++ b/test/test-utils.sh
@@ -71,6 +71,22 @@ else
 fi
 export SED_BUFFER_FLAG="--unbuffered"
 
+# [NOTE]
+# Specifying cache disable option depending on stat(coreutils) version
+# TODO: investigate why this is necessary #2327
+#
+if grep -q -i -e 'ID=ubuntu' /etc/os-release && grep -q -i -e 'VERSION_ID="20.04"' /etc/os-release; then
+    STAT_BIN=(stat)
+elif grep -q -i -e 'ID=debian' /etc/os-release && grep -q -i -e 'VERSION_ID="10"' /etc/os-release; then
+    STAT_BIN=(stat)
+elif grep -q -i -e 'ID="centos"' /etc/os-release && grep -q -i -e 'VERSION_ID="7"' /etc/os-release; then
+    STAT_BIN=(stat)
+elif [ "$(uname)" = "Darwin" ]; then
+    STAT_BIN=(stat)
+else
+    STAT_BIN=(stat --cache=never)
+fi
+
 function get_xattr() {
     if [ "$(uname)" = "Darwin" ]; then
         xattr -p "$1" "$2"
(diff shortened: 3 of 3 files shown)


gaul:
This fails on some systems with older coreutils but let's ignore that for now.  I don't understand why this helps.  The GNU coreutils changelog says:

> stat(1) supports a new --cached= option, used on systems with statx(2)
> to control cache coherency of file system attributes,
> useful on network file systems.

s3fs is a network filesystem but since this is a new file name there shouldn't be any caching involved.

ggtakec:
@gaul Great thanks for noticing this difference in the stat command.
Then, I would like you to release the draft and promote it.
Maybe this also affects #2260.

#### Supplement
I was also looking into errors in CI for `test_mtime_file` (and `test_update_time_cp_p`).

In `test_mtime_file`, it is confirmed that `ctime`/`mtime`/`atime` is included in the `x-amz-meta-*time` header in the `PUT` for file creation (`cp -p`) and the `HEAD` request to obtain its stat.
However, in the results of the `stat` command, the `mtime` value had changed.
```
* Headers at PUT/HEAD requests
    x-amz-meta-atime: 1694874095.118007656
    x-amz-meta-ctime: 1694874095.155758760
    x-amz-meta-mtime: 1694874095.135758493

* The result(error)
    cp(-p) expected times: mtime( 1694874095.135758493 == 1694874095.157601785 ), ctime( 1694874095.135758493 != 1694874095.155758760 ), atime( 1694874095.118007656 == 1694874095.118007656 )
```
At this time, `mtime` has changed from the response value.
That value is not recorded anywhere, this is strange.

But I confirmed that the correct value was obtained by adding the option `--cached=never` given as same as this PR.

#### In order to be compatible with all OS for CI testing
The versions of coreutils (stat) installed on each OS are as follows.
```
Ubuntu 23.04       9.1
Ubuntu 22.04       8.32
Ubuntu 20.04       8.30     <--- Not have "--cache" option
Debian bookworm    9.1
Debian bullseye    8.32
Debian buster      8.30     <--- Not have "--cache" option

CentOS 7           8.22     <--- Not have "--cache" option
Rocky Linux 9      8.32
Rocky Linux 8      8.30
Fedora 38          9.1
Fedora 37          9.1

Suse/Leap:15       8.32
Alpine 3.17        9.1
```
This option is supported in versions `8.32` and above, with the exception of `Rocky Linux 8`, which appears to have this option even in `8.30`.
Therefore, we can avoid this by branching by OS and version and adding options.

Shall we like to add the following(example codes) to `test-util.sh` in order to include the `--cache` option on an unsupported OS?

```
# [NOTE]
# Specifying cache disable option depending on stat(coreutils) version
#
STAT_CACHE_DISABLE="--cache=never"
if grep -q -i -e 'ID=ubuntu' /etc/os-release && grep -q -i -e 'VERSION_ID="20.04"' /etc/os-release; then
    STAT_CACHE_DISABLE=""
elif grep -q -i -e 'ID=debian' /etc/os-release && grep -q -i -e 'VERSION_ID="10"' /etc/os-release; then
    STAT_CACHE_DISABLE=""
elif grep -q -i -e 'ID="centos"' /etc/os-release && grep -q -i -e 'VERSION_ID="7"' /etc/os-release; then
    STAT_CACHE_DISABLE=""
fi

...
...
...

function get_permissions() {
    if [ "$(uname)" = "Darwin" ]; then
        stat -f "%p" "$1"
    else
        # shellcheck disable=SC2086
        stat ${STAT_CACHE_DISABLE} -c "%a" "$1"
    fi
}
```

gaul:
I'm not sure this is the correct solution.  Why are these attributes being cached in the first place?  Why did the behavior in CI recently change?  We can merge an improved version of this PR as a short-term solution but I think there may be some deeper root cause.

ggtakec:
I re-checked the prerequisites.
First, from the logs of the last CI build with no errors and the CI build where errors first started occurring, I checked the differences in the versions of the entire OS (Runner) image and coreutils.
There were no differences in OS images and coreutils versions except for Fedora 37/38.
(I inferred from the SHA of the OS image and whether updates were made by the package manager)
Regarding Fedora, there were differences in the OS images. (But I think maybe coreutils is the same thing.)

Next, if it does not specify the cached option of the stat command (probabry  same as `default`), it depends on the file system, so I was looking into `statvfs`.
I recently made a change to `s3fs_statfs`(#2305), so I investigated its impact.
The change itself has no effect, but I tried several `f_flag` values. (I expected the `ST_SYNCHRONOUS` flag, but there was no change in the result.)
I tried some of this flags and didn't find any change in results.

I suspect that there may have been changes in s3fs or other areas that are out of our control.
(However, I couldn't find anything about this in Github's release notes or anything like that.)

Is there anything else I should investigate?

It will likely take some time to determine the cause of this change. (There is also the question of how long we should support old coreutils)
Even if this investigation continues, I feel that it is necessary to reflect this PR content in the test code to avoid errors.
Although it is possible to ignore errors and inspect and merge the accumulated PR, I think it is better to adapt this PR even temporarily.

gaul:
Agree that we need to do something to unblock other work.  I integrated your proposed changes and opened an issue to track this strange behavior.

gaul:
This PR now includes a workaround to skip the recent Alpine Linux failures.

ggtakec (approved):
Thank you for opening this PR.

Currently, I asked if there were any changes to Github Actions' CI Runner, but haven't heard back yet(I've been waiting for about a week). I think it depends on the option for Disk (mount) on the Runner's base host side, so I am checking that.

I don't think there will be any problems with this modification.
However, I will send a PR later, so please check the `STAT_BIN` judgment part. (Although that PR may not be necessary)

ggtakec:
@gaul I received a reply from Github Support regarding this matter.
```
The changes introduced are shared in this release:
https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20230911.1

Initially, we see that there was a Kernal Version update. The runner images are built upon Azure Marketplace offerings, which recently had their ubuntu22's kernel updated to 6.2.0-1011-azure.

For next steps, I recommend [creating an issue post directly on the actions/runner-images repository](https://github.com/actions/runner-images/issues) to share your use case. From there, the maintainers of the runner images can directly look into your use case and evaluate how the image update may have impacted it.
```
After all, it seems that there has been a change in Runner's base OS since the `--cache` option became necessary.

I will refrain from pursuing this further. (Probably because you will have to do very tedious package comparisons, etc.)

The fixes in this PR are consistent with the base OS that has been changed, and we think it is okay to continue as is.

The above is for reference.

PR #2320, opened 2023-09-16
Add stat helper for user and group

gaul:


test/integration-test-main.sh
```diff
@@ -651,6 +642,7 @@ function test_multipart_copy {
 function test_multipart_mix {
     describe "Testing multi-part mix ..."
 
+    # TODO: why is this necessary?
```
ggtakec, line 645: I think I decided to clear the cache file here to avoid reaching the disk capacity limit when executing Github Actions on macos.

ggtakec (approved):

PR #2326, opened 2023-09-19
Document support for temporary credentials

iainelder:
Documents support for the `AWS_SESSION_TOKEN` variable. The variable is one way to support temporary credentials.

Without the documentation, the only way to know about the support is to read [s3fs_cred.cpp](https://github.com/s3fs-fuse/s3fs-fuse/blob/a83f4a48d0286062fc6837e913a523c3286974fd/src/s3fs_cred.cpp#L956-L972). 

Extends https://github.com/s3fs-fuse/s3fs-fuse/pull/2281 .


iainelder:
Discovered while researching an answer for https://github.com/benkehoe/aws-sso-util/issues/111.

gaul:
Thank you for your contribution @iainelder!

Issue #2328, opened 2023-09-23
Rethink CI targets


gaul:
Our CI has a large combination of distros:

* ubuntu:23.04
* ubuntu:22.04 (remove?)
* ubuntu:20.04 (remove?)
* debian:bookworm
* debian:bullseye (remove?)
* debian:buster (remove?)
* rockylinux:9
* rockylinux:8
* centos:centos7
* fedora:38
* fedora:37 (remove?)
* opensuse/leap:15
* alpine:3.17 (remove?)
* macOS (disabled)

I don't think that s3fs is well-served by this coverage since most of them are the same.  We should instead only test the latest version or versions that packagers actively maintain.  For example, RHEL 9, 8, and 7 all have a good value to ensure that packagers can use older compilers.  But Ubuntu 23.04, 22.04, and 20.04 are not since they don't repackage newer s3fs.  Same with Debian.  Further alpine has some strange issue that doesn't seem worth investigating.

Further some of the older distros have old versions of cppcheck and shellcheck that we should avoid.  Instead I think we should have distinct CI tasks for cppcheck and shellcheck and stop running them as part of each CI task.  I guess we can be more aggressive about version checks which is almost the same thing.

ggtakec:
In my opinion, I would like to preserve the OS provided as LTS.
_I think it is okay to exclude OS versions that are outside of LTS._
I would like to support the Current OS version (before it becomes Stable or LTS). (For example, Ubuntu23.04)

Regarding Alpine, I have the opposite opinion and should continue to support this OS.
Considering their maintenance story, I think we can always support only latest released version. (We should also change to 3.18 as soon as possible)
This is because the number of users using Alpine is increasing in consideration of Container, and we expect this to continue to increase in the future.
Regarding Alpine, there are some problems passing the test(now), but the related packages required by s3fs are also well supported, so we don't expect any major problems to occur.
(For example, Alpine already supports OpenSSL3.0, which is attractive.)

Below is the state of the OS in my opinion.

- ubuntu:23.04  
Either one is still fine. It is possible to wait until LTS
- ubuntu:22.04  
Main support
- ubuntu:20.04  
I think it should be supported (optional if unnecessary)
- debian:bookworm  
Main support
- debian:bullseye  
I think it should be supported (optional if unnecessary)
- debian:buster  
I think it should be removed.
- rockylinux:9  
Main support
- rockylinux:8  
Main support
- centos:centos7  
I think this can no longer be supported.
[...]

PR #2332, opened 2023-09-25
Upgrade CI to Alpine 3.18

gaul:
References #2328.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
           - fedora:38
           - fedora:37
           - opensuse/leap:15
-          - alpine:3.17
+          - alpine:3.18
 
     container:
       image: ${{ matrix.container }}
--- a/.github/workflows/linux-ci-helper.sh
+++ b/.github/workflows/linux-ci-helper.sh
@@ -210,7 +210,7 @@ elif [ "${CONTAINER_FULLNAME}" = "opensuse/leap:15" ]; then
     INSTALL_CHECKER_PKGS="cppcheck ShellCheck"
     INSTALL_CHECKER_PKG_OPTIONS=""
 
-elif [ "${CONTAINER_FULLNAME}" = "alpine:3.17" ]; then
+elif [ "${CONTAINER_FULLNAME}" = "alpine:3.18" ]; then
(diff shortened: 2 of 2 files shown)


gaul:
This has an interesting failure in bullseye:

```
stat: cannot statx 'test-s3fs-24610.txt': No such file or directory
```

Maybe similar to the caching issues we have observed earlier?

Jessb6996 (approved):

ggtakec:
This seems similar to the issue being detected with alpine.
I think that s3fs could not detect the file uploaded from outside(by aws command).
(But I would know if I had the curldbg logs, but the info level logs don't allow me to be sure of this.)

ggtakec:
@gaul Just for quick reference.
I don't know if this will help you, but I cut out the `alpine 3.18` test (`test_mtime_file` and `test_update_time_cp_p` only) on Github Actions and tried that.
(The log setting is `curldbg` and `dbglevel=dbg`.)
https://github.com/ggtakec/s3fs-fuse/actions/runs/6604273488

This result is very strange.
These tests (`test_mtime_file` and `test_update_time_cp_p`) compare the stats of the underlying file object and its copied file object.
The (`PUT` and) `HEAD` requests to the underlying file are being called successfully.
However, the copy file operation is only being called with a `PUT` request.
(Originally, there would be a `HEAD` request after the `PUT`, which would retake the stat information.)

It may be a good idea to check whether there have been any changes to the cache related to the stat command since `alpine 3.18`.

ggtakec:
@gaul 
I have identified the cause of this issue, it has been fixed and posted PR #2361.
Once #2361 is merged, please make this Alpine PR from Draft to official PR.

gaul:
Alpine still fails after rebasing:

```
s3fs: [INF] s3fs.cpp:s3fs_getattr(1041): [path=/testrun-7150][pid=2653,uid=0,gid=0]
cp with -p option expected updated ctime: 1000000000.000000000 != 1000000000.000000000 and same mtime: 1000000000.000000000 == 1000000000.000000000, atime: 1000000000.000000000 == 1000000000.000000000

test_update_time_cp_p failed
```

ggtakec:
I think this remaining bug was fixed in #2364.

PR #2337, opened 2023-10-03
Add a helper script to compile all targets

gaul:
This is useful to compile different SSL libraries and 32-bit targets.


gaul:
This is just for local testing when I was doing the `std::unique_ptr` conversion.  I guess CI could run this but I don't think that has high value.

ggtakec:
Thanks for the explanation.

ggtakec (approved):

PR #2341, opened 2023-10-07
Force disk free space recovery in test(for only macos)

ggtakec:
#2320 

### Details
In the test code that was pointed out in #2320, there was still a problem with the `test_multipart_mix` test on macos12.

#### Background
For macos, I was forcing the test file size to zero at the beginning of the `test_multipart_mix`, because this was to synchronize the free space area.
But it was detected that the free space may not be recovered immediately after deleting a file.
(Free space recovery may not be reflected in the results of the `statvfs` or `df` commands.)
I still don't know why this delay occurs.

#### Purpose
The code trying to recover free space on the current disk wasn't good enough.
The `test_multipart_mix test` repeatedly creates files of several tens MB.
In rare cases, you will run out of disk space and the test will fail (on macos12).
To reliably reproduce this, you can run only the `test_multipart_mix` test function repeatedly(dozens of times) on macos12.

#### Changing test code
Forces the test files and its cache files size to zero, and deletes them upon completion of the `test_multipart_mix` test.
And call `sync` command for disk.

So far, this fix has not caused any shortage of disk space (problem where free space is not restored).

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -642,10 +642,6 @@ function test_multipart_copy {
 function test_multipart_mix {
     describe "Testing multi-part mix ..."
 
-    # TODO: why is this necessary?
-    if [ "$(uname)" = "Darwin" ]; then
-       cat /dev/null > "${BIG_FILE}"
-    fi
     ../../junk_data $((BIG_FILE_BLOCK_SIZE * BIG_FILE_COUNT)) > "${TEMP_DIR}/${BIG_FILE}"
     dd if="${TEMP_DIR}/${BIG_FILE}" of="${BIG_FILE}" bs="${BIG_FILE_BLOCK_SIZE}" count="${BIG_FILE_COUNT}"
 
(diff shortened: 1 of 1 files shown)


test/integration-test-main.sh
```diff
@@ -717,10 +713,29 @@ function test_multipart_mix {
...
+    # In the case of macos, even if you delete a file, there seems to be a
+    # delay in the free space being reflected.
+    # Testing the ensure_diskfree option shows that if this is not done, free
+    # disk space will be exhausted.
+    #
+    if [ "$(uname)" = "Darwin" ]; then
+        cat /dev/null > "${TEMP_DIR}/${BIG_FILE}"
+        cat /dev/null > "${TEMP_DIR}/${BIG_FILE}-mix"
+        cat /dev/null > "${BIG_FILE}"
+        cat /dev/null > "${BIG_FILE}-mix"
```
gaul, line 728: I wonder how this is happening.  Perhaps Time Machine/backup software is running?  I wonder if we could disable it for test directories?

ggtakec: I also tried various things.
_(I don't think it's Time Machine. (I think it is possible to check the macos process, but I have not confirmed it))_

Simply deleting test files, including cache files, it seems that is often not enough to recover disk space.

This problem is unlikely to become apparent during normal testing, but it rarely occurred on macos 12.
Then, I tried running `test_multipart_mix` 20-50 times in a row and it was able to confirm that this problem was manifested.
At that time, when I checked the free disk space every time I executed `test_multipart_mix`, I was able to see the increase and decrease.
When the test fails, free disk space has not been recovered.

If s3fs is started in a mode that does NOT use cache, it creates a temporary file(tmpfile), so I think a possible that the space in that temporary file may not be freed.
However, the test that confirmed this error was using a cache file.

I eventually knew that we could restore free disk space by reducing this file size to zero and deleting it and calling sync.
(It's possible that it depended on APFS, but I'm not familiar with that.)

gaul:
Merging to unblock this.  I suspect there is something else strange going on.  But we should address this and other disk space issues with a different approach, e.g., https://github.com/derbylock/zfus.  Because I would like to test much larger files, e.g., 5 GB + 1, and we will never have enough temporary space on the CI runners.

PR #2342, opened 2023-10-07
Convert some `const` to `constexpr`

gaul:
This guarantees that the function or value will resolve at compile-time.


src/s3fs_cred.cpp
```diff
@@ -113,21 +113,19 @@ bool UpdateS3fsCredential(char** ppaccess_key_id, char** ppserect_access_key, ch
...
-const int   S3fsCred::IAM_EXPIRE_MERGIN         = 20 * 60;              // update timing
-const char* S3fsCred::ECS_IAM_ENV_VAR           = "AWS_CONTAINER_CREDENTIALS_RELATIVE_URI";
-const char* S3fsCred::IAMCRED_ACCESSKEYID       = "AccessKeyId";
-const char* S3fsCred::IAMCRED_SECRETACCESSKEY   = "SecretAccessKey";
-const char* S3fsCred::IAMCRED_ROLEARN           = "RoleArn";
-
-const char* S3fsCred::IAMv2_token_url           = "http://169.254.169.254/latest/api/token";
-int         S3fsCred::IAMv2_token_ttl           = 21600;
-const char* S3fsCred::IAMv2_token_ttl_hdr       = "X-aws-ec2-metadata-token-ttl-seconds";
-const char* S3fsCred::IAMv2_token_hdr           = "X-aws-ec2-metadata-token";
```
gaul, line 130: Previously these were not truly `const` -- `IAMv2_token_hdr` could be reassigned to another `const char *`

src/s3fs_cred.cpp
```diff
@@ -113,21 +113,19 @@ bool UpdateS3fsCredential(char** ppaccess_key_id, char** ppserect_access_key, ch
...
+constexpr char S3fsCred::ALLBUCKET_FIELDS_TYPE[];
```
gaul, line 116: clang-tidy flags this which is strange -- `static` class fields have external linkage and need some definition.

ggtakec: Will the flag be set even if I write the following in the declaration part of `ALLBUCKET_FIELDS_TYPE`?
````
static constexpr char ALLBUCKET_FIELDS_TYPE[] = {'\0'};
````

File: src/s3fs_cred.cpp (hunk 1 above)
Comment (ggtakec) [reply src/s3fs_cred.cpp:130]: Please explain in more detail what you mean by this comment.

Comment (gaul) [reply src/s3fs_cred.cpp:130]: `IAMv2_token_hdr = "foo"` later is legal.  `const char *` only means that the data pointed to must be constant, not the variable itself.  `const char * const`  or `const char[]` prevent this.

Comment (ggtakec) [reply src/s3fs_cred.cpp:130]: Thanks

File: src/s3fs_cred.cpp (hunk 2 above)
Comment (ggtakec) [reply src/s3fs_cred.cpp:116]: It seems that the error can be avoided by surrounding the part where constexpr is defined in the `S3fsCred` and `S3fsLog` class members with the following `#if` statement.(in `s3fs_cred.cpp` and `s3fs_logger.cpp`)
```
#if !defined(__cpp_inline_variables)
constexpr char S3fsCred::ALLBUCKET_FIELDS_TYPE[];
constexpr char S3fsCred::KEYVAL_FIELDS_TYPE[];
constexpr char S3fsCred::AWS_ACCESSKEYID[];
constexpr char S3fsCred::AWS_SECRETKEY[];

constexpr char S3fsCred::ECS_IAM_ENV_VAR[];
constexpr char S3fsCred::IAMCRED_ACCESSKEYID[];
constexpr char S3fsCred::IAMCRED_SECRETACCESSKEY[];
constexpr char S3fsCred::IAMCRED_ROLEARN[];

constexpr char S3fsCred::IAMv2_token_url[];
constexpr char S3fsCred::IAMv2_token_ttl_hdr[];
constexpr char S3fsCred::IAMv2_token_hdr[];
#endif
```
_Since it is declared (implicitly?) as an inline element, it seems possible to avoid it by writing it like this._

gaul: I disabled `readability-redundant-declaration` since this behavior seems weird.

ggtakec (approved):

PR #2343, opened 2023-10-07
Fixed  ETag parsing at completing the Multipart upload part

ggtakec:
n/a

### Details
When completing a multipart upload(in the `UploadMultipartPostComplete` method) it was not possible to remove double quotes from the response ETag string.
But it was deleted in the similar `CopyMultipartPostComplete`, then mixed Etags(with/without double quotes) for Mix Multipart Upload.
So far, this doesn't cause uploads to fail, but I'd like to keep the specs the same.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -4218,6 +4218,7 @@ bool S3fsCurl::UploadMultipartPostComplete()
     if (it == responseHeaders.end()) {
         return false;
     }
+    std::string etag = peeloff(it->second);
 
     // check etag(md5);
     //
@@ -4227,11 +4228,11 @@ bool S3fsCurl::UploadMultipartPostComplete()
     // https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingKMSEncryption.html
     //
     if(S3fsCurl::is_content_md5 && sse_type_t::SSE_C != S3fsCurl::GetSseType() && sse_type_t::SSE_KMS != S3fsCurl::GetSseType()){
-        if(!etag_equals(it->second, partdata.etag)){
+        if(!etag_equals(etag, partdata.etag)){
             return false;
         }
     }
-    partdata.petag->etag = it->second;
+    partdata.petag->etag = etag;
     partdata.uploaded = true;
 
     return true;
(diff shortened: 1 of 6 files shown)


src/curl.cpp
```diff
@@ -4232,6 +4232,10 @@ bool S3fsCurl::UploadMultipartPostComplete()
         }
     }
     partdata.petag->etag = it->second;
+    if(partdata.petag->etag.size() >= 2 && *partdata.petag->etag.begin() == '"' && *partdata.petag->etag.rbegin() == '"'){
+        partdata.petag->etag.erase(partdata.petag->etag.size() - 1);
+        partdata.petag->etag.erase(0, 1);
+    }
```
gaul, line 4238: Could you abstract the etag erasing logic from `S3fsCurl::CopyMultipartPostComplete` and `etag_equals` into a helper function?

gaul (changes requested):

src/curl.cpp
ggtakec: Changed to use `trim` string function(defined in `string_util.h`).
Strictly speaking, it no longer only removes pairs of double quotes, but I used this on the assumption that the `ETag` response would not contain anything like that.
The same thing can be done by implementing it in the `ETag` key acquisition method, but it seems to be complicated, so this modification method is simpler.
I have rebased it, so please review it again.

src/curl.cpp
```diff
@@ -4218,6 +4218,7 @@ bool S3fsCurl::UploadMultipartPostComplete()
     if (it == responseHeaders.end()) {
         return false;
     }
+    std::string etag = trim(it->second, "\"");
```
gaul, line 4221: I'm a little confused what this is trying to do -- it seems to unconditionally removed quotes.  But shouldn't s3fs always send the same ETag to the server that it previously returned to us?

ggtakec: OK, let's peel off it from a string strictly enclosed in double quotes.
If the ETag is inappropriate, it will be detected as an error during the upload process anyway.
To do this, I will add one utility function for string manipulation.

src/curl_util.cpp
```diff
@@ -319,17 +319,9 @@ const char* getCurlDebugHead(curl_infotype type)
...
-    if(s1.length() > 1 && s1[0] == '\"' && *s1.rbegin() == '\"'){
-        s1.erase(s1.size() - 1);
-        s1.erase(0, 1);
-    }
-    if(s2.length() > 1 && s2[0] == '\"' && *s2.rbegin() == '\"'){
-        s2.erase(s2.size() - 1);
-        s2.erase(0, 1);
-    }
-    return 0 == strcasecmp(s1.c_str(), s2.c_str());
+    return 0 == strcasecmp(trim(s1, "\"").c_str(), trim(s2, "\"").c_str());
```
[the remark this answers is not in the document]
ggtakec: Here too, I will remove only strict pairs of double quotes and change it to compare.

src/string_util.cpp
```diff
@@ -127,6 +127,14 @@ std::string trim(const std::string &s, const char *t /* = SPACES */)
     return trim_left(trim_right(s, t), t);
 }
 
+std::string peeloff(const std::string &s)
```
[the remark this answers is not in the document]
ggtakec: I fixed it. And I pushed it.

PR #2346, opened 2023-10-09
s3fs_cred: print detailed error message when stat file fails

adamqqqplay:
n/a

### Details
When the password file does not exist, now only the error "s3fs: unexpected error from stat" will be reported, which is not convenient for locating the problem.

In order to fix this problem, this patch adds detailed error message printing.


gaul:
Thank you for your contribution @adamqqqplay!

PR #2348, opened 2023-10-15
Fixed a bug in the re-upload part of Streamupload

ggtakec:
n/a

#### Problem phenomenon
The CI test was failing in rare cases in the `test_multipart_mix` test of StreamUpload.
(It seemed to happen more often on `macos` than on `linux`)

#### StreamUpload logic
In StreamUpload, if the part range of each multipart is filled with updates, that part will be uploaded first without waiting for `flush`.
This StreamUpload logic allows for re-uploads of the same part if there are updates(written) to the same part range after the first upload for the same part.
In this case, when flushed the file, this part will be re-uploaded and the previous upload will be overwritten(cancelled).
This is the usual expected StreamUpload's logic.

#### Behavior in case of error
The error detected this time occurred in a case where Flush was called and a re-upload was started when the previously uploaded part had not yet finished uploading.
It means that uploads of the same part range were running at the same time.
Normally, the upload that started first will complete first, and the ETag value will be overwritten by the update caused by the re-upload.
However, on this error case, the re-upload process completed first, and the earlier upload process completed later, resulting in an invalid (cancelled) ETag value.

#### Fixed bug
In order to avoid this, we have modified it so that if it is detected that the part to be re-uploaded is still being uploaded, the re-upload will start after the upload including that part is completed.

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -1737,7 +1737,7 @@ int FdEntity::RowFlushMixMultipart(PseudoFdInfo* pseudo_obj, const char* tpath)
                 return result;
             }
             untreated_list.ClearParts(untreated_start, untreated_size);
-	    }
+        }
         // complete multipart uploading.
         if(0 != (result = NoCacheCompleteMultipartPost(pseudo_obj))){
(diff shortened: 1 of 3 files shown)


gaul:
I think this would be hard for us to integration test.  I guess we could improve the unit testing although this would require a lot of other work.

PR #2369, opened 2023-11-17
s3fs: make dir size not zero

eryugey:
Directory has size 0, which looks weired and may confuse users. So fake dir size as 4k.

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -804,7 +804,11 @@ bool convert_header_to_stat(const char* path, const headers_t& meta, struct stat
     }
 
     // size
-    pst->st_size = get_size(meta);
+    if(S_ISDIR(pst->st_mode)){
+        pst->st_size = 4096;
+    }else{
+        pst->st_size = get_size(meta);
+    }
 
     // uid/gid
     pst->st_uid = get_uid(meta);
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1059,6 +1059,9 @@ static int s3fs_getattr(const char* _path, struct stat* stbuf)
                 stbuf->st_size = tmpstbuf.st_size;
             }
         }
+        if(0 == strcmp(path, "/")){
+            stbuf->st_size = 4096;
+        }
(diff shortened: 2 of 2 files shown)


gaul:
Did you actually test this?

eryugey:
> Did you actually test this?

Yes, I compiled s3fs and mount a bucket, dir size becomes 4096 instead of 0

original
```bash
drwxr-xr-x 1 root     root            0 Feb  9  2023 testdir
```

patched
```bash
drwxr-xr-x 1 root     root  4096 Feb  9  2023 testdir
```

gaul:
So you ran `make -C test/ check` which succeeded?  Because all of the CI runnings are failing or timing out.

eryugey:
> So you ran `make && make -C test/ check` which succeeded? Because all of the CI runners are failing or timing out.

Test fails because of "Error: Invalid or corrupt jarfile s3proxy-2.0.0"

src/s3fs.cpp
```diff
@@ -1059,6 +1059,9 @@ static int s3fs_getattr(const char* _path, struct stat* stbuf)
                 stbuf->st_size = tmpstbuf.st_size;
             }
         }
+        if((1 == strlen(path)) && (0 == strncmp(path, "/", 1))){
```
gaul, line 1062: Just change this to `strcmp(path, "/")` instead of more complicated logic.

src/cache.cpp
```diff
@@ -805,6 +805,9 @@ bool convert_header_to_stat(const char* path, const headers_t& meta, struct stat
 
     // size
     pst->st_size = get_size(meta);
+    if(S_ISDIR(pst->st_mode)){
+        pst->st_size = 4096;
+    }
```
gaul, line 810: This should have `} else { pst->st_size = get_size(meta); }` to avoid setting the size twice.

gaul (changes requested):
You can remove s3proxy.jar to fix your corruption issue.  You must have interrupted the download or something.

gaul:
Thank you for your contribution @eryugey!

PR #2375, opened 2023-11-19
Call C++11 `get_time` and `put_time`

gaul:
This removes workarounds and fixed-length buffers.

### DIFF
--- a/src/metaheader.cpp
+++ b/src/metaheader.cpp
@@ -19,6 +19,8 @@
  */
 
 #include <ctime>
+#include <iomanip>
+#include <sstream>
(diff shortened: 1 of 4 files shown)


ggtakec (approved):
Thanks, The code is no problem and is approved.
(The macos 12 error is due to a different cause)

gaul:
This is strange since `std::get_time` should be part of C++11: https://en.cppreference.com/w/cpp/io/manip/get_time

ggtakec:
It seems to depend on the GCC version.
CentOS7 is GCC4.8, and GCC5 or higher is required to use these.

Issue #2372, opened 2023-11-18
list_bucket request during new file creation process(macos fuse-t)


ggtakec:
### Relevant PR(Issue)
#2198

### Details about issue
Due to support for `macos fuse-t`, Github Actions on macos12 may fail.
(We fixed some bugs in supporting `macos fuse-t`, but it seems that some bugs still remain.)

The problem is that s3fs no longer creates 0-byte files when first creating new files.
(This support is necessary to speed up s3fs, and the current status will continue to be maintained.)

When creating a new file, the file does not exist until s3fs uploads the file contents.
In `macos-fuse-t`, something like `list_bucket` may be called while this file is created but has not yet been uploaded(flushed).
If the `list_bucket` is called at this timing, the response to the `HEAD` request will not include the new created file, so the user will receive an error that the file does not exist.
`utimens`, `chmod`, etc. are fixed, but this `list_bucket` is not yet fixed.
(We have to include reading from the `dummy stat cache`, and we are currently considering modifying it.)

ggtakec:
This problem has been resolved, but there may still be problems. It continues with #2418.

PR #2376, opened 2023-11-19
Corrected list_bucket to search in stat cache during creating new file

ggtakec:
#2372 
#2198

### Details
#2198 supported `macos fuse-t`, but the bug in #2372 remained, so fixed it.

#### About bug
When creating a new file, s3fs does not create a `0-byte` file and delays the upload until the file is flushed.
During this delayed period, s3fs creates a temporary cache in the Stat cache to avoid `file not found` errors in case s3fs receives an instruction to get the file's stat information.

The problem in #2372 is similar to this, but the `readdir` was received during a delayed period.
Because this `readdir` (`list bucket`) is not supported, some `macos12` tests are failing as a bug.

#### Correspondence
When registering temporary cache information in the Stat cache for a new file, use the file's parent directory as a key and retain the cache information that maps the file name list.
This mapping information exists until the temporary cache is deleted.
By using this temporary cache, s3fs can support `readdir` (`list bucket`) and avoid problems.

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -338,21 +338,17 @@ bool StatCache::AddStat(const std::string& key, const headers_t& meta, bool forc
     }
     S3FS_PRN_INFO3("add stat cache entry[path=%s]", key.c_str());
 
-    bool found;
-    bool do_truncate;
-    {
-        AutoLock lock(&StatCache::stat_cache_lock);
-        found       = stat_cache.end() != stat_cache.find(key);
-        do_truncate = stat_cache.size() > CacheSize;
-    }
+    AutoLock lock(&StatCache::stat_cache_lock);
 
-    if(found){
-        DelStat(key.c_str());
+    if(stat_cache.end() != stat_cache.find(key)){
+        // found cache
+        DelStat(key.c_str(), AutoLock::ALREADY_LOCKED);
     }else{
-        if(do_truncate){
+        // check: need to truncate cache
+        if(stat_cache.size() > CacheSize){
             // cppcheck-suppress unmatchedSuppression
             // cppcheck-suppress knownConditionTrueFalse
-            if(!TruncateCache()){
+            if(!TruncateCache(AutoLock::ALREADY_LOCKED)){
                 return false;
             }
         }
@@ -386,9 +382,6 @@ bool StatCache::AddStat(const std::string& key, const headers_t& meta, bool forc
         }
     }
 
-    // add
-    AutoLock lock(&StatCache::stat_cache_lock);
-
     const auto& value = stat_cache[key] = std::move(ent);
 
     // check symbolic link cache
@@ -398,6 +391,13 @@ bool StatCache::AddStat(const std::string& key, const headers_t& meta, bool forc
             DelSymlink(key.c_str(), AutoLock::ALREADY_LOCKED);
         }
     }
+
+    // If no_truncate flag is set, set file name to notruncate_file_cache
+    //
+    if(no_truncate){
+        AddNotruncateCache(key);
+    }
+
     return true;
 }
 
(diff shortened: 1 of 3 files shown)


ggtakec:
There seems to be a case where an error was detected first, so I need to check.

src/cache.cpp
```diff
@@ -746,6 +767,101 @@ bool StatCache::DelSymlink(const char* key, AutoLock::Type locktype)
...
+
+// [Background]
+// When s3fs creates a new file, the file does not exist until the file contents
+// are uploaded.(because it doesn't create a 0 byte file)
+// From the time this file is created(opened) until it is uploaded(flush), it
+// will have a Stat cache with the No truncate flag added.
+// This avoids file not existing errors in operations such as chmod and utimens
+// that occur in the short period before file upload.
+// Besides this, we also need to support readdir(list_bucket), this method is
+// called to maintain the cache for readdir and return its value.
```
gaul, line 837: I'm confused about how this happens -- while the file is uploading, the stat object must continue to exist.  Why doesn't readdir find this entry?

src/cache.cpp
```diff
@@ -509,10 +516,18 @@ void StatCache::ChangeNoTruncateFlag(const std::string& key, bool no_truncate)
...
+            if(0L == ent->notruncate){
+                // need to add no truncate chache.
+                AddNotruncateCache(key);
+            }
             ++(ent->notruncate);
         }else{
             if(0L < ent->notruncate){
                 --(ent->notruncate);
+                if(0L == ent->notruncate){
+                    // need to delete from no truncate chache.
```
gaul, line 528: Typo: cache.  Repeated above.

src/cache.cpp
ggtakec: In `readdir` is processing , `list_bucket()` is called.
`list_bucket` always calls `ListBucketRequest` to list the objects that currently exist.
(This is necessary in order to detect objects manipulated from other than s3fs in the S3 bucket.)
After calling `list_bucket`, `readdir_multi_head` is called to get information about each object.
The Stat cache will be accessed for objects listed in this `readdir_multi_head`.

The `readdir` process is as described above, and while a new file is being created(from `create` to `flush`), the file does not exist in the results of `ListBucketRequest`.
So this PR, if there is a temporary Stat cache for a newly created file, the file name is added to the list of `ListBucketRequest`.
With this addition, newly created files can also be read using readdir.

src/cache.cpp
ggtakec: Thanks!

ggtakec:
@gaul 
I modified this PR source code and rebase it.

I found new problem that is in `StatCache` class because this `list_bucket` was called while creating a new file.
The `AddStat` and `DelStat` methods of the `StatCache` class were not atomic operations.
This problem does not occur with calls other than `macos macfuse-t`, but this is exposed in only `macos`.(This causes some tests to fail intermittently.)

The pushed new code involved changing these methods to atomic operations, which allowed the tests to pass.
Please re-review this fixed code.

gaul:
I'm concerned about adding a separate cache since it will be hard to coordinate.  Why can't the fdcache hold these not-yet uploaded files?

ggtakec:
@gaul I had the same concerns as you.
However, I believe that this PR is the best way to deal with the `macos fuse-t` support issue that is currently occurring for the following reasons.

As you should know, the root cause of this is that s3fs does not create 0-byte file that is supposed to be created initially(at the create system call) when creating a new file.
_I believe that not creating 0-byte files is the best practice and the right thing to do, considering performance, platform, etc._

In order to avoid creating this 0-byte file object, the meta information (stat information) is held pending in FdCache.
It is also held in StatCache as information that will not be cached out until the file is created(until it is flushed and uploaded).
Both of these caches have prevented the problem from occurring until now.

However, in the case of `macos fuse-t`, the `readdir` system call is called on the parent directory before the new file is created(flushed). _This is the cause of the problem for `macos fuse-t` support._

FdCache and StatCache are caches of Stat(meta) information for **file object**.
It is not a cache of list information for file objects contained in a directory.
It might be possible to generate a list of files for a particular directory from FdCache or StatCache, but it would probably be more complicated.
(Each readdir is called, the existence of a file cache for a specific parent directory need to be checked from all cache data.)

If s3fs had a list of files being created, s3fs could find creating files that were not detected by listbucket.
_In the previous case, it has been detected by listbucket as there is a 0 byte file._
And if s3fs has that, it will perform better.

In the flow of creating a new file, temporary information is registered and deleted in StatCache when it has not been uploaded yet.
This PR uses this existing logic to add a cache of file list being created using directories as keys to StatCache.

I think it's best to fix this PR because it doesn't require any major logic additions or changes, and the timing of cache additions and deletions can be synchronized.
(Even with the method using FdCache, it seems necessary to have a static list with directories as keys, and if we implement it, this PR is probably better.)

##### NOTE
There is still an error in Github Actions, so I will make a fix soon.

ggtakec:
> There is still an error in Github Actions, so I will make a fix soon.

Since the problem does not occur again, there is no operation log, so I cannot investigate.
If it happens again, I will investigate at that time.

ggtakec:
@gaul I merged this.

ggtakec:
Unlike when I PR, the `test_update_time_append` test is causing an error.
I've pretty much figured out the cause, so I'm looking into ways to fix it.
As soon as we have a fix, we will post a new PR.

PR #2377, opened 2023-11-20
Remove obsolete C++11 `#ifdef`

gaul:

PR #2379, opened 2023-11-25
Changed to use devtoolset for building on CentOS 7

ggtakec:
#2377
#2328

### Details
Migration to C++11 is being hampered by CentOS 7's lower version of GCC4(lack of support for std::get_time/put_time).
To avoid this, I changed to use `devtoolset-11`.
These template methods are implemented inline, so the `devtoolset-11` environment is not required at runtime.
For the time being, we can use this method to eliminate differences with other OS.

@gaul
I think there may be no problem with this method.
Even if there are users who build on CentOS 7, I don't think it will be difficult to build an environment with `devtoolset`.

### DIFF
--- a/.github/workflows/linux-ci-helper.sh
+++ b/.github/workflows/linux-ci-helper.sh
@@ -79,6 +79,7 @@ PACKAGE_ENABLE_REPO_OPTIONS=""
 PACKAGE_INSTALL_ADDITIONAL_OPTIONS=""
 SHELLCHECK_DIRECT_INSTALL=0
 AWSCLI_DIRECT_INSTALL=1
+SCL_INSTALL=0
 
 if [ "${CONTAINER_FULLNAME}" = "ubuntu:23.10" ]; then
     PACKAGE_MANAGER_BIN="apt-get"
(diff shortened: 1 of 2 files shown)


.github/workflows/linux-ci-helper.sh
```diff
@@ -183,6 +184,15 @@ elif [ "${CONTAINER_FULLNAME}" = "centos:centos7" ]; then
     INSTALL_CHECKER_PKGS="cppcheck jq"
     INSTALL_CHECKER_PKG_OPTIONS="--enablerepo=epel"
 
+    # [NOTE]
+    # To use C++11 std::get_time/put_time, you need GCC5 or higher.
+    # For building, we need to use devtoolset on CentOS 7.
+    # (devtoolset is not required in the s3fs execution.)
+    #
+    SCL_INSTALL=1
+    INSTALL_SCL_PACKAGES="centos-release-scl"
```
juliogonzalez, line 193: One note about this, it could be we can't update package for EPEL7 (RHEL7, CentOS7, etc) anymore, as I think the EPEL building system will not allow me to use non EL7/EPEL7 packages (`devtoolset-11` is in the Software Collections SIG repository you are installing with `centos-release-scl`).

That said, I can still update the SPEC and README in my repository to explain this, so people can build the RPMs on their own.

The only limitation is that 1.93 will be the latest version available at EPEL7.

gaul: I don't think we should break CentOS 7 in EPEL.  We can just revert the trivial C++11 change for now and un-revert it when CentOS 7 goes EOL in summer 2024.

ggtakec: @juliogonzalez
Thank you for pointing out the packaging. That helps a lot!
I didn't even pay attention to the packaging.

@gaul
I agree.
I don't want to make changes that affect the package, so could you submit a PR to revert #2377?
I will confirm that PR and close this PR.

ggtakec: @gaul I confirmed #2381 and merged it.

ggtakec:
I merged #2381 fore reverting  #2377, so close this PR.

PR #2383, opened 2023-11-26
Convert `FdEntity` to `std::unique_ptr`

gaul:


gaul:
This has trickier memory ownership than previous conversions.  Please look more carefully.

PR #2387, opened 2023-12-02
Demonstrate static checking with curl_handles_lock

gaul:
This can find missing locks at compile-time.

### DIFF
--- a/.github/workflows/linux-ci-helper.sh
+++ b/.github/workflows/linux-ci-helper.sh
@@ -66,6 +66,7 @@ AWSCLI_ZIP_FILE="awscliv2.zip"
 #-----------------------------------------------------------
 # Parameters for configure(set environments)
 #-----------------------------------------------------------
+CXX="g++"
 CXXFLAGS="-O -DS3FS_PTHREAD_ERRORCHECK=1"
 CONFIGURE_OPTIONS="--prefix=/usr --with-openssl"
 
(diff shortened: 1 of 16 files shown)


gaul:
@ggtakec I'd like to move towards static lock checking instead of relying on only dynamic ThreadSanitizer.  I think this can find bugs more reliably.  What do you think?  `FdEntity` will be a more complicated case but I can demonstrate this before moving any call sites.

gaul:
Also the `Mutex` wrappers should not be required but we would need to opt into clang libc++ instead of GNU libstdc++.

.github/workflows/linux-ci-helper.sh
```diff
@@ -290,8 +293,11 @@ fi
 #-----------------------------------------------------------
 echo "${PRGNAME} [INFO] Set environment for configure options"
 
-echo "CXXFLAGS=${CXXFLAGS}"                   >> "${GITHUB_ENV}"
-echo "CONFIGURE_OPTIONS=${CONFIGURE_OPTIONS}" >> "${GITHUB_ENV}"
+cat << EOF > "${GITHUB_ENV}"
```
ggtakec, line 296: The ${GITHUB_ENV} file has been overwritten.
Please change to addendum.
Is there a reason why you changed to Here Document?

src/mutex.h
```diff
@@ -0,0 +1,150 @@
+/*
```
ggtakec, line 1: Please make sure to indent 4 space characters.

ggtakec (commented):
I think I can promote this amendment.

I have a few questions, so I will summarize them here.
##### About license
Is there any external license involved in mutex.h?
I was curious about this because when I searched for `THREAD_ANNOTATION_ATTRIBUTE__`, I saw some examples of like code.

##### About AutoLock::Type
This `Mutex` class will replace all `AutoLock` classes, but can't it implement `AutoLock::Type`?
(Is it difficult to add attributes for `ThreadSanitizer` ?)
However, there is no particular problem even if it cannot be implemented.

gaul:
Sorry this should be broken into separate PRs:

* Remove `ALREADY_LOCKED` state
* Move to `std::mutex`
* Enable static lock checking

gaul:
Pursing this in a variety of separate PRs.

PR #2388, opened 2023-12-02
Replace miscellaneous pointers with `unique_ptr`

gaul:


ggtakec (approved):
Sorry for being late in reviewing to this PR.
I think this fix is fine(I did some testing).

What I was concerned about was the changes to `S3fsLog::logfile`.
The reason s3fs set this variable as a pointer was to take into account the influence of `SIOF`(Static Initialization Order Fiasco).
This is a static variable related to log output, and we were concerned about the possibility that this class variable could not be initialized before being manipulated.
After trying a few things, I couldn't confirm the problem at the moment, so I'm merging this PR.

Issue #2418, opened 2024-02-23
macos-fuse-t CI often fails


ggtakec:
### Relevant PR(Issue)
#1769 #2154 #2191 #2198 #2205 #2372

### Details about issue
#2372 After merging, I've been observing for a while, and errors are randomly occurring in macos-fuse-t's CI.

When investigating the logs, I noticed that some files (directories) could not be detected or some false positives were detected.
In the `test_truncate_cache` test, there was a case where the call to delete a subdirectory was missed(it was not called), and the deletion of the parent directory failed.

We need to look into it more closely.

ggtakec:
With the fix in #2724, the MacOS CI tests continue to run without failure.
We'll continue to monitor the results, but will close this issue for now.
We'll reopen it if there are any problems.

PR #2419, opened 2024-02-23
Update ChangeLog and configure.ac for 1.94

gaul:


ggtakec:
> test_update_directory_time_touch failed

Did it happen on macos?
I'm thinking of collecting them in #2418.
(There seems to be a reason why macos still requires a time lag)

PR #2426, opened 2024-03-05
Use `xmlFreeNsList` to deallocate `xmlNsPtr`

gaul:
Previously `xmlFree` only freed the top-level allocation and not the entire linked list.  Multiple namespaces should be rare but perhaps some S3 implementations have both a global and local namespace.  The `xmlFree(void*)` prototype hid the type mismatch.

### DIFF
--- a/src/s3fs.h
+++ b/src/s3fs.h
@@ -64,11 +64,6 @@
           xmlFreeDoc(doc); \
           S3FS_MALLOCTRIM(0); \
         }while(0)
-#define S3FS_XMLFREE(ptr) \
-        do{ \
-          xmlFree(ptr); \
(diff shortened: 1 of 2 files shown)


gaul:
I'm not too sure about the caller expectations: https://fossies.org/linux/libxml2/tree.c#l_5844

gaul:
Maybe a clue: https://stackoverflow.com/questions/64111724/libxml2-memory-leak-renaming-namespace-prefixes although a different function.

gaul:
Judging by https://fossies.org/linux/libxml2/tree.c#l_5885 `xmlFree(ret);` this looks correct already.

PR #2430, opened 2024-03-07
Convert `thpoolman_param` to value

gaul:
This simplifies memory management.

### DIFF
--- a/src/threadpoolman.cpp
+++ b/src/threadpoolman.cpp
@@ -53,13 +53,14 @@ void ThreadPoolMan::Destroy()
     }
 }
 
-bool ThreadPoolMan::Instruct(std::unique_ptr<thpoolman_param> pparam)
+bool ThreadPoolMan::Instruct(const thpoolman_param& param)
 {
     if(!ThreadPoolMan::singleton){
         S3FS_PRN_WARN("The singleton object is not initialized yet.");
         return false;
     }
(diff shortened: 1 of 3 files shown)


src/threadpoolman.cpp
```diff
@@ -84,30 +85,24 @@ void* ThreadPoolMan::Worker(void* arg)
...
             AutoLock auto_lock(&(psingleton->thread_list_lock));
 
-            if(!psingleton->instruction_list.empty()){
-                pparam = std::move(psingleton->instruction_list.front());
-                psingleton->instruction_list.pop_front();
-                if(!pparam){
-                    S3FS_PRN_WARN("Got a semaphore, but the instruction is empty.");
-                }
+            if(psingleton->instruction_list.empty()){
+                continue;
```
ggtakec, line 93: I think we need to add the following message from the code before modification here.
It would be nice to have this output when debugging multi-threads.

```
S3FS_PRN_WARN("Got a semaphore, but the instruction is empty.");
```
You can change the message output level from WARN to DBG.

ggtakec (commented):
Please check a comment.

src/threadpoolman.cpp
gaul: Done.

ggtakec (approved):

Issue #2424, opened 2024-02-27
[FR] Resolve IPv4/v6 Only Option


jyxjjj:
#### Version of s3fs being used (`s3fs --version`)
```
Amazon Simple Storage Service File System V1.93 (commit:unknown) with OpenSSL
Copyright (C) 2010 Randy Rizun <rrizun@gmail.com>
License GPL2: GNU GPL version 2 <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)
```
Name        : fuse
Version     : 2.9.9
Release     : 17.fc39
Architecture: x86_64
Install Date: Mon 26 Feb 2024 06:01:56 PM +08
Group       : Unspecified
Size        : 224031
License     : GPL+
Signature   : RSA/SHA256, Thu 20 Jul 2023 07:21:37 AM +08, Key ID 75cf5ac418b8e74c
Source RPM  : fuse-2.9.9-17.fc39.src.rpm
Build Date  : Thu 20 Jul 2023 04:34:41 AM +08
Build Host  : buildhw-x86-01.iad2.fedoraproject.org
Packager    : Fedora Project
Vendor      : Fedora Project
URL         : http://fuse.sf.net
Bug URL     : https://bugz.fedoraproject.org/fuse
Summary     : File System in Userspace (FUSE) v2 utilities
Description :
With FUSE it is possible to implement a fully functional filesystem in a
userspace program. This package contains the FUSE v2 userspace tools to
mount a FUSE filesystem.
```

#### Kernel information (`uname -r`)
`6.6.13-200.fc39.x86_64`

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)
```
NAME="Fedora Linux"
VERSION="39 (Cloud Edition)"
ID=fedora
VERSION_ID=39
VERSION_CODENAME=""
PLATFORM_ID="platform:f39"
PRETTY_NAME="Fedora Linux 39 (Cloud Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:39"
HOME_URL="https://fedoraproject.org/"
[...]

PR #2432, opened 2024-03-10
Added ipresolve option

ggtakec:
#2424 

### Details
Added new option `ipresolve`.
This is an option to fix the IP address version of resolve, and is the `CURLOPT_IPRESOLVE` provided by libcurl.

The default value of the ipresolve option is `whatever`(or `default`), which is the same as `CURL_IPRESOLVE_WHATEVER`.
If `IPv4` (or `v4`) is specified, resolve will be executed using IPv4 only. (`CURL_IPRESOLVE_V4`)
Also, if `IPv6` (or `v6`) is specified, resolve will be executed using IPv6 only. (`CURL_IPRESOLVE_V6`)


src/curl.cpp
```diff
@@ -1949,7 +1967,11 @@ bool S3fsCurl::ResetHandle(AutoLock::Type locktype)
     if(CURLE_OK != curl_easy_setopt(hCurl, S3FS_CURLOPT_KEEP_SENDING_ON_ERROR, 1) && !run_once){
         S3FS_PRN_WARN("The S3FS_CURLOPT_KEEP_SENDING_ON_ERROR option could not be set. For maximize performance you need to enable this option and you should use libcurl 7.51.0 or later.");
     }
-
+    if(CURL_IPRESOLVE_WHATEVER != S3fsCurl::ipresolve_type){    // CURL_IPRESOLVE_WHATEVER is default, so not need to set.
+        if(CURLE_OK != curl_easy_setopt(hCurl, CURLOPT_IPRESOLVE, S3fsCurl::ipresolve_type)){
```
gaul, line 1971: Should this log when using the non-default value?

src/curl.cpp
```diff
@@ -1174,6 +1175,23 @@ bool S3fsCurl::SetProxyUserPwd(const char* file)
...
+bool S3fsCurl::SetIPResolveType(const char* value)
+{
+    if(!value){
+        return false;
+    }
+    if(0 == strcasecmp(value, "ipv4") || 0 == strcasecmp(value, "v4")){
+        S3fsCurl::ipresolve_type = CURL_IPRESOLVE_V4;
+    }else if(0 == strcasecmp(value, "ipv6") || 0 == strcasecmp(value, "v6")){
+        S3fsCurl::ipresolve_type = CURL_IPRESOLVE_V6;
+    }else if(0 == strcasecmp(value, "whatever") || 0 == strcasecmp(value, "both") || 0 == strcasecmp(value, "default")){
```
gaul, line 1187: Remove unneeded "both" value?  Also do we need `ipv4` and `ipv6`?  This will read weirdly: `ipresolve=ipv4`.

gaul (changes requested):

src/curl.cpp
ggtakec: This was an option value that I added casually, but it seems inappropriate.
Make it only ipv4/ipv6/whatever.

ggtakec: Changed option values, please re-check it.

PR #2443, opened 2024-04-12
retry request on HTTP 429 error

blinkseb:
Hi,

Some non-Amazon S3 implementation have a stricter quota / rate-limiting than AWS (for example Scaleway). It means that it's quiet easy to be rate limited in case of large buckets for example. In this case, an HTTP error 429 is returned and the request fails.

This PR makes s3fs retry when an HTTP error 429 is returned, as it's already doing for 500 and 503 errors.

Thanks!

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -2582,6 +2582,7 @@ int S3fsCurl::RequestPerform(bool dontAddAuthHeaders /*=false*/)
                         result = -ENOTSUP;
                         break;
 
+                    case 429:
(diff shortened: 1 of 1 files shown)


gaul:
Thank you for your contribution @blinkseb!  If you have a contact at Scaleway please ask them to change 429 to 503 SlowDown to better match other clients: https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList

Issue #2421, opened 2024-02-25
Providing more details in docs


tuxpeople:
Hi

I'm stumbled upon this:
https://github.com/s3fs-fuse/s3fs-fuse/blob/70a30d6e26a5dfd07a00cf79ce1196079e5ab11a/COMPILATION.md?plain=1#L9-L23

Would it be possible to differentiate between build dependencies and runtime dependencies?

Thanks!

ggtakec:
@tuxpeople I'm sorry for my late replying.
Could you tell me specifically what you stumbled on?

gaul:
I believe he wants to separate build dependencies:

* automake
* gcc-c++
* make
* pkg-config (or your OS equivalent)

From runtime dependencies:

* fuse >= 2.8.4
* libcurl
* libxml2
* openssl
* mime.types (the package providing depends on the OS)
	* s3fs tries to detect `/etc/mime.types` as default regardless of the OS
	* Else s3fs tries to detect `/etc/apache2/mime.types` if OS is macOS
	* s3fs exits with an error if these files are not exist
	* Alternatively, you can set mime.types file path with `mime` option without detecting these default files

Agree that this would be a slight improvement.  It would also be good to clarify that OpenSSL is not required and that either GnuTLS or NSS can be used.

tuxpeople:
> @tuxpeople I'm sorry for my late replying. 

No worries, I've got the same problem with issues people open in my repos ;-)

> I believe he wants to separate build dependencies:

Yes. Unfortunately, I'm unable to remember why exactly I asked for this. I think I've tried to create a container image with it, and my goal was to have a small footprint using a multi-stage build.

Anyway, the PR looks good to me, thanks a lot!

ggtakec:
@tuxpeople Thank you for the reply.
I merged PR#2445.
If there are any parts that are still unclear, please let us know.
(This PR will be closed.)

PR #2445, opened 2024-04-15
Updated COMPILATION.md about compilation on linux

ggtakec:
#2421

### Details
Supplemented the explanation of the `Compilation on Linux` section of COMPILATION.md.

### DIFF
--- a/COMPILATION.md
+++ b/COMPILATION.md
@@ -6,7 +6,9 @@ If you want specific instructions for some distributions, check the [wiki](https
 
 Keep in mind using the pre-built packages when available.
 
-1. Ensure your system satisfies build and runtime dependencies for:
+## Compilation on Linux
+
+### Ensure your system satisfies build and runtime dependencies for:
 
 * fuse >= 2.8.4
 * automake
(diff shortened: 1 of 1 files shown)


COMPILATION.md
```diff
@@ -6,32 +6,59 @@ If you want specific instructions for some distributions, check the [wiki](https
 
 Keep in mind using the pre-built packages when available.
 
-1. Ensure your system satisfies build and runtime dependencies for:
+## Compilation on Linux
+
+### Ensure your system satisfies build and runtime dependencies for:
```
gaul, line 11: Should this be build time dependencies only?

ggtakec: Does that mean we should also write `cppcheck` / `shellcheck` etc in this file?
The purpose of this `COMPILATION.md` file was not to perform checks or tests, but only to generate an executable file, so I considered it to be the minimum necessary.
Would it be better to add more detailed dependencies?

ggtakec:
@gaul I merge this PR.
If additional changes are required to the explanatory text, I create another PR to update COMPILATION.md again.

PR #2460, opened 2024-05-27
Fixed error reports of failure in cppcheck 2.14.0

ggtakec:
n/a

### Details
In the MacOS CI test, cppcheck now uses 2.14.0 and an error was output, so this PR fixed it.


ggtakec:
The `sanitize_thread` check is failing for a reason unrelated to this PR, so please ignore it.

PR #2466, opened 2024-06-04
Remove uses of `AutoLock::ALREADY_LOCKED`

gaul:
Instead annotate the methods with `REQUIRES` so that the caller knows if they should lock.  For public interfaces, introduce `HasLock` wrappers.  This simplifies control flow, allows migration to `std::mutex`, and eventually will enable use of static lock checking.

### DIFF
--- a/src/autolock.h
+++ b/src/autolock.h
@@ -23,6 +23,9 @@
 
 #include <pthread.h>
 
+// empty annotation to indicate lock requirement
+#define REQUIRES(...)
+
 //-------------------------------------------------------------------
 // AutoLock Class
 //-------------------------------------------------------------------
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -231,7 +231,7 @@ bool StatCache::GetStat(const std::string& key, struct stat* pst, headers_t* met
             if(ent->noobjcache){
                 if(!IsCacheNoObject){
                     // need to delete this cache.
-                    DelStat(strpath, AutoLock::ALREADY_LOCKED);
+                    DelStatHasLock(strpath);
                 }else{
                     // noobjcache = true means no object.
                 }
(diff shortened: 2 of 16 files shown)


gaul:
References #2387.

ggtakec (commented):
I agree with eliminating ALREADY_LOCKED.
However, I have some concerns about this approach.

That is, the `pthread_mutex_t` variable has been changed to `public`.

I think it is dangerous to make the flag(`pthread_mutex_t`) for exclusive control of internal variables of each class public.

Currently, the reason for using `ALREADY_LOCKED` is that it is necessary when the member function(mainly public) is called by another member function of the same class(the mutex already has beed locked by this function).

I would prefer to separate the methods like `DelStatWithoutLock()` and `DelStat()` that you implemented in this PR.
And we can make the method that does not lock like `DelStatWithoutLock()` private, and the method that always locks like `DelStat()` public.
_I think most calls from within the same class will be `DelStatWithoutLock()`._

With this change, I think you don't need to use the `REQUIRES()` macro.

And I think we can clearly distinguish whether to lock or not by the method name.(I think you can achieve the same effect as this PR.)

gaul:
> That is, the `pthread_mutex_t` variable has been changed to public.

I am neutral on this -- I will try to use the suggested `FooWithoutLock` wrappers but I don't anticipate exposing the lock will cause abuse, e.g., some unrelated code calling `pthread_mutex_unlock`.

> With this change, I think you don't need to use the `REQUIRES()` macro.

The `REQUIRES` and related macros will be required for compile-time lock checking.  This PR is just simplifying the code to allow enable these checks in the future.

gaul:
I added `HasLock` wrappers for the required methods instead of exposing the lock.

ggtakec (approved):
Thanks for fix. I will merge this.

PR #2471, opened 2024-06-12
Replace `memset` with C++11 value initialization

gaul:
This generates the same code but is safer due to using an implicit length and allowing member initialization.


ggtakec (approved):

ggtakec:
@gaul Becase of merging #2470 caused a conflict with the changes to cache.h.(The reason are simple though.)
Could you please check the conflict in cache.h and rebase?

PR #2477, opened 2024-06-23
Opt in to all clang-tidy checks by default

gaul:
Opt out of the noisy checks.  Disable clang-tidy on Debian bullseye and buster and Ubuntu 20.04 due to segfaults.


gaul:
Ubuntu 20.04 crashes with a backtrace that does not identify which checker is to blame.

ggtakec (approved):

ggtakec:
@gaul Could you please resolve the conflicts?
Then I'll merge.

gaul:
I think we will have to tweak the checks or disable Ubuntu 20.04?

ggtakec:
@gaul 
It seems there is no way to fix this error on Ubuntu 20.04.
_(I have found several other posts about clang-tidy segmentation faults. But I coud not find same issue and hot to fix it.)_
Could you modify ci.yaml to bypass clang-tidy for ubuntu 20.04 or ignore the error?

ggtakec (approved):

PR #2481, opened 2024-06-30
Convert `pthread_t` to C++11 `std::thread`

gaul:
This has better cross-platform support and stronger type-safety.

### DIFF
--- a/src/curl_multi.cpp
+++ b/src/curl_multi.cpp
@@ -21,6 +21,8 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cerrno>
+#include <future>
+#include <thread>
 #include <vector>
 
 #include "s3fs.h"
@@ -112,14 +114,12 @@ bool S3fsMultiCurl::SetS3fsCurlObject(std::unique_ptr<S3fsCurl> s3fscurl)
 
 int S3fsMultiCurl::MultiPerform()
 {
-    std::vector<pthread_t>   threads;
+    std::map<std::thread::id, std::pair<std::thread, std::future<int>>> threads;
     bool                     success = true;
     bool                     isMultiHead = false;
     Semaphore                sem(GetMaxParallelism());
-    int                      rc;
 
     for(s3fscurllist_t::iterator iter = clist_req.begin(); iter != clist_req.end(); ++iter) {
-        pthread_t   thread;
         S3fsCurl*   s3fscurl = iter->get();
         if(!s3fscurl){
             continue;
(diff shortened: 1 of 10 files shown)


gaul:
@siketyan Are you still interested in Windows support?  I wonder if the C++11 modernizations might allow more native Windows instead of using MSYS2?

siketyan:
@gaul Unfortunately I am no longer a Windows user 😢 so I cannot help well about porting for Windows. Thank you for mentioning me anyway.

src/curl_multi.cpp
```diff
@@ -21,6 +21,8 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cerrno>
+#include <future>
+#include <thread>
```
ggtakec, line 25: These two header files are defined in `curl_multi.h`, so there is no need to include them here.

src/openssl_auth.cpp
```diff
@@ -26,7 +26,6 @@
 #include <cstdlib>
 #include <cerrno>
 #include <mutex>
-#include <pthread.h>
 #include <unistd.h>
```
ggtakec, line 29: If you want to make the corrections you commented in subsequent reviews, you will need to `#include <thread>` here.

src/openssl_auth.cpp
```diff
@@ -96,14 +95,6 @@ static void s3fs_crypt_mutex_lock(int mode, int pos, const char* file, int line)
...
 }
 
-static unsigned long s3fs_crypt_get_threadid() __attribute__ ((unused));
-static unsigned long s3fs_crypt_get_threadid()
-{
-    // For FreeBSD etc, some system's pthread_t is structure pointer.
-    // Then we use cast like C style(not C++) instead of ifdef.
-    return (unsigned long)(pthread_self());
-}
-
```
ggtakec, line 106: This function is necessary.
However, it requires changes to std::thread as follows:
```
static unsigned long s3fs_crypt_get_threadid() __attribute__ ((unused));
static unsigned long s3fs_crypt_get_threadid()
{
    std::hash<std::thread::id> hash;
    return (unsigned long)hash(std::this_thread::get_id());
}
```

src/openssl_auth.cpp
```diff
@@ -144,7 +135,6 @@ bool s3fs_init_crypt_mutex()
     s3fs_crypt_mutex = new std::mutex[CRYPTO_num_locks()];
     // static lock
     CRYPTO_set_locking_callback(s3fs_crypt_mutex_lock);
-    CRYPTO_set_id_callback(s3fs_crypt_get_threadid);
```
ggtakec, line 147: The call to this function must remain.

src/openssl_auth.cpp
```diff
@@ -162,7 +152,6 @@ bool s3fs_destroy_crypt_mutex()
     CRYPTO_set_dynlock_destroy_callback(nullptr);
     CRYPTO_set_dynlock_lock_callback(nullptr);
     CRYPTO_set_dynlock_create_callback(nullptr);
-    CRYPTO_set_id_callback(nullptr);
```
ggtakec, line 165: The call to this function must remain.

ggtakec (changes requested):
I would like you to consider fixing `s3fs_crypt_get_threadid`.
This call should be required for OpenSSL's multithread support.
(However, I checked and it does not affect the error that currently occurs in `sanitize_thread`. = The error continues to occur, so we are investigating it separately.)

src/openssl_auth.cpp
gaul: I'm not sure this is actually used in OpenSSL 1.0 and later: https://linux.die.net/man/3/crypto_set_id_callback but I added it as you suggested.

src/curl_multi.cpp
ggtakec: OK, that's fine.

src/openssl_auth.cpp
ggtakec: Thank you for adding it.
I'm also concerned, but there's the datarace issue that's also mentioned in other issues, so I'd like to keep it.

PR #2482, opened 2024-06-30
Call `std::gettime` instead of `strptime`

gaul:
This reduces the Windows-specific code.


gaul:
It looks like CentOS 7 C++ 11 support is incomplete:

```
string_util.cpp:56:14: error: 'get_time' is not a member of 'std'
```

Abandoning this for now.

ggtakec:
If you decide to stop supporting CentOS7 due to its EOL(June 30th), you can also reinstate this code.
(We need to discuss support for CentOS7 in the existing issue.)

Issue #2469, opened 2024-06-10
Drop RHEL/CentOS 7 support when EOL


gaul:
s3fs should consider dropping RHEL/CentOS 7 support when it [reaches end-of-life on 30 June 2024](https://www.redhat.com/en/technologies/linux-platforms/enterprise-linux/rhel-7-end-of-maintenance).  Some benefits:

* Removes a blocker for FUSE 3 support, #1159
* Newer GCC allows use of C++17, e.g., `std::optional`, `std::string_view`, etc.
* Possibly move to AWS SDK due to the above, #1068
* Reduce debugging of CI failures due to older toolchain, e.g., OpenJDK 11

Some downsides:

* CentOS 7 users will continue to use it and file bugs.  We will be unable to support them via EPEL.

@juliogonzalez any thoughts about this?  Will EPEL updates still work after EOL?

gaul:
> And about this, I would like to carefully consider the transition to AWS C++.

Right, C++14 support is necessary for using the SDK but using C++14 doesn't obligate us to use the SDK.  Let's discuss that issue when we have a proof of concept and not as part of the RHEL 7 EOL.

ggtakec:
Guthub Actions(CI) tests for CentOS 7 can no longer be cpmpleted(could not use mirror repository).
We need to prepare to remove CentOS 7 from GHA.

gaul:
It will be difficult going forward to maintain CentOS 7 compability without CI.  Do you think we should run a 1.95 release now so those users get a final release?  Will EPEL still accept packages for this older distro?  @juliogonzalez

juliogonzalez:
> It will be difficult going forward to maintain CentOS 7 compability without CI. Do you think we should run a 1.95 release now so those users get a final release? Will EPEL still accept packages for this older distro? @juliogonzalez

I can try, at least.

Worst, worst case, I can prepare the package outside EPEL and provide it via the openbuildservice for CentOS7.

It could be interesting, as some companies are going to extend CentOS7 lifecycle, such as SUSE (Liberty Linux) and CIQ (Bridge). Of course being those subscription distributions, there's no way to provide support going forward, but it would be nice to provide a final version.

juliogonzalez:
> Guthub Actions(CI) tests for CentOS 7 can no longer be cpmpleted(could not use mirror repository). We need to prepare to remove CentOS 7 from GHA.

[...]

PR #2485, opened 2024-07-01
Dropped Github Actions(CI) for CentOS 7 and Debian 10(Buster)

ggtakec:
#2469

### Details
On June 30th, CentOS 7 and Debian10(Buster) reached EOL.

In Github Actions(CI), the error `Could not resolve host: mirrorlist.centos.org; Unknown error` is displayed, so the mirror repository cannot be accessed, so CI execution fails.
Users will be able to use the yum repository up to the current version 1.94 (until it is discontinued), but CI execution will no longer be possible.
Debian10(Buster) is still able to run CI, but CI execution has been stopped with this PR.


juliogonzalez:
Careful, if you want to provide 1.95 for CentOS, as discussed at https://github.com/s3fs-fuse/s3fs-fuse/issues/2469#issuecomment-2200413467, make sure you don't merge this before 1.95 is released :-)

PR #2486, opened 2024-07-01
Simplify curl progress tracking

gaul:
Use a struct with named fields instead of a `pair` for clarity and use a single `map` to store the structs for efficiency.


ggtakec (approved):
Thanks.

PR #2487, opened 2024-07-04
Add missing locking and fix incorrect annotations

gaul:
Found via static lock checking.


ggtakec (changes requested):
Please modify `fdcache_fdinfo.[cpp.h]`.
The reason is that there is a loophole in the exclusive processing of the `upload_id` variable, and I think that this should be reviewed.
The contents of the file that needs to be modified have been uploaded to my Gist below, so please refer to it.
- [fdcache_fdinfo.cpp](https://gist.github.com/ggtakec/2a8a928e923e6bddf4c97182c5bda4ea)
- [fdcache_fdinfo.h](https://gist.github.com/ggtakec/729abd1122a2901c1f503e87dcb6c5d5)

PR #2488, opened 2024-07-04
Experiment with C++17

gaul:
Most of these features should work now that we dropped CentOS 7 support.


gaul:
It appears that all the CI targets support C++17.  So we should consider using these features after releasing 1.95.

src/s3fs_xml.cpp
```diff
@@ -229,12 +230,13 @@ static unique_ptr_xmlChar get_exp_value_xml(xmlDocPtr doc, xmlXPathContextPtr ct
     return exp_value;
 }
 
-bool get_incomp_mpu_list(xmlDocPtr doc, incomp_mpu_list_t& list)
+std::optional<incomp_mpu_list_t> get_incomp_mpu_list(xmlDocPtr doc)
```
gaul, line 233: I don't think C++17 has major advancements compared to C++11 but `std::optional`  will make error checking more obvious and safer.  These kinds of little enhancements must wait until after we release 1.95 in #2496.

gaul:
OpenSUSE Leap 15.6 supported until at least April 2026: https://en.opensuse.org/Lifetime

eran132:
Hi @gaul — is this still on track? OpenSUSE Leap 15.6 passed EOL in April and was removed from the CI matrix in 1484c60, so the C++17 floor should be clean across all current targets. Happy to rebase this if it'd help unstick it.

gaul:
This isn't a PR intended to be merged -- it just tested the compiler support for C++17.  I opened #2895 for the first step but this isn't really something we need contributor help with.  The benefits of this are modest so don't expect any user-facing changes.

PR #2490, opened 2024-07-05
Add `GUARDED_BY` to `FdEntity` and fix locking

gaul:


gaul:
@ggtakec I annotated the most difficult class and found a variety of unlocked field accesses.  At least some of the comments appear to be wrong.  But I think the static lock checker is doing good work!

src/fdcache_entity.cpp
```diff
@@ -2386,6 +2402,7 @@ int put_headers(const char* path, headers_t& meta, bool is_copy, bool use_st_siz
 
 int FdEntity::UploadPendingHasLock(int fd)
 {
+    const std::lock_guard<std::mutex> lock(fdent_data_lock);
```
ggtakec, line 2405: Reentrancy issues were found in FlushHasLock() and RowFlushHasLock(), so locking here is no longer possible.
More details are explained in the full comments of this PR.

src/fdcache_entity.cpp
```diff
@@ -2552,6 +2570,8 @@ bool FdEntity::AddUntreated(off_t start, off_t size)
 
 bool FdEntity::GetLastUpdateUntreatedPart(off_t& start, off_t& size) const
 {
+    const std::lock_guard<std::mutex> lock(fdent_lock);
```
ggtakec, line 2573: This method is called as a callback from PseudoFdInfo::UploadBoundaryLastUntreatedArea().
The FdEntity object passed to PseudoFdInfo::UploadBoundaryLastUntreatedArea() has already performed this locking, so it cannot be locked here.

src/fdcache_entity.cpp
```diff
@@ -2561,6 +2581,8 @@ bool FdEntity::GetLastUpdateUntreatedPart(off_t& start, off_t& size) const
 
 bool FdEntity::ReplaceLastUpdateUntreatedPart(off_t front_start, off_t front_size, off_t behind_start, off_t behind_size)
 {
+    const std::lock_guard<std::mutex> lock(fdent_lock);
```
ggtakec, line 2584: This method is called as a callback from PseudoFdInfo::UploadBoundaryLastUntreatedArea().
The FdEntity object passed to PseudoFdInfo::UploadBoundaryLastUntreatedArea() has already performed this locking, so it cannot be locked here.

src/fdcache_entity.h
```diff
@@ -135,9 +137,12 @@ class FdEntity
         }
         int GetOpenCountHasLock() const REQUIRES(FdEntity::fdent_lock);
         // TODO: should this require a lock?
-        const std::string& GetPath() const { return path; }
+        std::string GetPath() const {
+            const std::lock_guard<std::mutex> lock(fdent_lock);
```
ggtakec, line 141: This fix is ​​fine, but it causes a deadlock in rename_object() by locking fdent_lock here. (A fix to s3fs.cpp is also required.)

ggtakec (changes requested):
It turns out there is a problem with the FdEntity::RowFlushHasLock() method.
Locking fdent_data_lock here seems to lead to a deadlock.

This method may call UploadPendingHasLock(), which may ultimately call RowFlushHasLock() method again.(There is a possibility of reentrancy.)

So that, RowFlushHasLock() method does not perform any locking, so it is necessary to define a method that wraps this method and performs the locking.

@gaul 
Thanks for finding the locking bug.
I've noticed that some additional code causes deadlocks in other places.
I've submitted a PR #2491 that brings that together, but would you be willing to merge that into this one?
(Or, if you're happy with the content of that PR, I'm happy to merge it.)

gaul:
Sorry I hadd not run with `ALL_TESTS=1` so I missed some deadlocks.  The latest PR should address the previous issues.

PR #2491, opened 2024-07-07
[Improvement #2490] Add GUARDED_BY to FdEntity and fix locking

ggtakec:
#2490

### Details
There were too many requests for fixes in PR #2490, so I created a PR based on #2490.

The biggest difference is that the RowFlushHasLock() (and FlushHasLock()) methods are reentrant, so in order to maintain a strict locking state, it is necessary to separate the functions.
For other details, please see the comments in PR #2490.

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -691,6 +691,9 @@ bool FdEntity::LoadAll(int fd, headers_t* pmeta, off_t* size, bool force_load)
 //
 bool FdEntity::RenamePath(const std::string& newpath, std::string& fentmapkey)
 {
+    const std::lock_guard<std::mutex> lock(fdent_lock);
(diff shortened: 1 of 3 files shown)


gaul:
I don't understand this PR.  Let's merge or improve the original #2490.

ggtakec:
@gaul 
This fix has the same purpose as #2490, and is the code that implements the parts I commented on.
(It includes code that was not fully explained in the comments in #2490.)

In the PR for #2490, there is a part that deadlocks (see the comments), and this code is the appropriate fix.

The code in this PR is just for reference, so could you merge it into the code in #2490?
If possible, I think it should be applied to #2490 as is.

gaul:
Sorry I missed this which duplicates #2549.  Let's try harder not to duplicate work int he future.

PR #2492, opened 2024-07-09
Add infrastructure for clang static lock checking

gaul:


gaul:
@ggtakec Sorry there are so many outstanding and conflicting PRs and this one should merge first.  This PR demonstrates how to use static lock checking with clang libc++ with its `std::mutex` annotations instead of g++ which requires a wrapper class.  I can remove the CI changes from this PR and introduce a separate target.  But the other changes will make future annotation fixes easier to write and review.

gaul:
I broke out the CI changes into #2493 and removed the annotation fixes.

ggtakec (approved):

PR #2495, opened 2024-07-09
Wrap `CURL*` in a `std::unique_ptr`

gaul:
This is safer and clarifies the ownership of pointers.


ggtakec (approved):
Thanks.
_Please be aware that I am currently preparing a relatively large fix, which may render the changes in this PR useless. However, those fixes will take a little more time._

gaul:
> _Please be aware that I am currently preparing a relatively large fix, which may render the changes in this PR useless. However, those fixes will take a little more time._

This sounds good but lets make sure as many pointers are owned, e.g., `std::unique_ptr`, as possible for safety and clarity.

PR #2499, opened 2024-07-14
Simplify `has_mp_stat`

gaul:
This is unnecessary since we moved from `pthread_mutex_t` to `std::atomic`.


ggtakec (approved):
Thanks for simplifying it.

PR #2501, opened 2024-07-15
Checked and fixed the internal data lock of PseudoFdInfo class

ggtakec:
#2487

### Details
Improved locking about upload_id variable in PseudoFdInfo class(`fdcache_fdinfo.[cpp|h]`).

### DIFF
--- a/src/fdcache_fdinfo.h
+++ b/src/fdcache_fdinfo.h
@@ -81,10 +81,11 @@ class PseudoFdInfo
         bool ResetUploadInfo() REQUIRES(upload_list_lock);
         bool RowInitialUploadInfo(const std::string& id, bool is_cancel_mp);
         bool CompleteInstruction(int result) REQUIRES(upload_list_lock);
-        bool ParallelMultipartUpload(const char* path, const mp_part_list_t& mplist, bool is_copy) REQUIRES(upload_list_lock);
+        bool ParallelMultipartUpload(const char* path, const mp_part_list_t& mplist, bool is_copy);
         bool InsertUploadPart(off_t start, off_t size, int part_num, bool is_copy, etagpair** ppetag) REQUIRES(upload_list_lock);
         bool CancelAllThreads();
         bool ExtractUploadPartsFromUntreatedArea(const off_t& untreated_start, const off_t& untreated_size, mp_part_list_t& to_upload_list, filepart_list_t& cancel_upload_list, off_t max_mp_size);
+        bool IsUploadingHasLock() const REQUIRES(upload_list_lock);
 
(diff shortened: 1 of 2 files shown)


src/fdcache_fdinfo.h
```diff
@@ -81,10 +81,11 @@ class PseudoFdInfo
         bool ResetUploadInfo() REQUIRES(upload_list_lock);
         bool RowInitialUploadInfo(const std::string& id, bool is_cancel_mp);
         bool CompleteInstruction(int result) REQUIRES(upload_list_lock);
-        bool ParallelMultipartUpload(const char* path, const mp_part_list_t& mplist, bool is_copy) REQUIRES(upload_list_lock);
+        bool ParallelMultipartUpload(const char* path, const mp_part_list_t& mplist, bool is_copy);
         bool InsertUploadPart(off_t start, off_t size, int part_num, bool is_copy, etagpair** ppetag) REQUIRES(upload_list_lock);
         bool CancelAllThreads();
         bool ExtractUploadPartsFromUntreatedArea(const off_t& untreated_start, const off_t& untreated_size, mp_part_list_t& to_upload_list, filepart_list_t& cancel_upload_list, off_t max_mp_size);
+        bool IsUploadingWitoutLock() const REQUIRES(upload_list_lock);
```
gaul, line 88: This is misspelled but let's use a consistent naming convention with the other code.  The methods that lock should be named `Foo` and the methods that require the caller to have the lock should be named `FooHasLock`.  I am happy to debate the best naming for this in a separate PR but this PR should follow the existing practice.

gaul (changes requested):
I don't know why we both need to work on conflicting static locking PRs simultaneously.  If you would like to take over this area I am happy to abandon my work.  I recommend fixing the existing warnings so we can enable CI instead of working on new areas that could introduce regressions.

src/fdcache_fdinfo.h
ggtakec: We will follow the HasLock naming convention.(Also, this is a mispelling.)

ggtakec:
@gaul 
Since fdcache_fdinfo.cpp/h was deleted from the fix file for #2487, I mistakenly thought that this would be a separate PR.
Also, I thought it was urgent to move forward with Release 1.95.

Please abandon my PR and prioritize yours.
(I will change this PR to Draft and close it as soon as the work on your PR is merged.)
Sorry for the confusion.

src/fdcache_fdinfo.h
ggtakec: Fixed.

Issue #2503, opened 2024-07-24
Compiling on Windows is broken


MargaretaWhite23:
#### Version of s3fs being used (`s3fs --version`)
```
Commit:411e42384ee88acaf388c3d36070d109274bd027
Version 1.94
```

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)
```
# pkg-config --modversion fuse.pc
2.8.4
```

#### Kernel information (`uname -r`)
`3.5.3-d8b21b8c.x86_64`

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)
```
# cat /etc/os-release
NAME=MSYS2
ID=msys2
PRETTY_NAME="MSYS2"
ID_LIKE="cygwin arch"
HOME_URL="https://www.msys2.org"
BUG_REPORT_URL="https://github.com/msys2/MSYS2-packages/issues"

```
#### How to run s3fs, if applicable
`N/A`

#### s3fs syslog messages (`grep s3fs /var/log/syslog`, `journalctl | grep s3fs`, or `s3fs outputs`)
`N/A`

### Details about issue
Compilation of the Windows executable seems to be broken
(Following guide on: 
https://github.com/s3fs-fuse/s3fs-fuse/blob/master/COMPILATION.md
)

Error message is as follows:

```
Administrator@WIN-HRGF8TFIGPP MSYS ~/s3fs-fuse-master
# make
make  all-recursive
make[1]: Entering directory '/home/Administrator/s3fs-fuse-master'
Making all in src
make[2]: Entering directory '/home/Administrator/s3fs-fuse-master/src'
g++ -DHAVE_CONFIG_H -I. -I..  -I/home/Administrator/s3fs-fuse-master/WinFsp/inc/fuse -I/usr/include/libxml2  -DUSE_OPENSSL_30   -Wall -fno-exceptions -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=3 -std=c++11 -g -O2 -MT s3fs.o -MD -MP -MF .deps/s3fs.Tpo -c -o s3fs.o s3fs.cpp
In file included from /home/Administrator/s3fs-fuse-master/WinFsp/inc/fuse/fuse_common.h:30,
                 from /home/Administrator/s3fs-fuse-master/WinFsp/inc/fuse/fuse.h:30,
                 from s3fs.h:26,
                 from s3fs.cpp:35:
[...]

PR #2505, opened 2024-07-25
Fix minor issues when compiling with MSYS

gaul:
References #2503.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -149,7 +149,7 @@ static int set_bucket(const char* arg);
 static int my_fuse_opt_proc(void* data, const char* arg, int key, struct fuse_args* outargs);
 static fsblkcnt_t parse_bucket_size(char* value);
 static bool is_cmd_exists(const std::string& command);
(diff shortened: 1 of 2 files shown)


src/s3fs.cpp
```diff
@@ -149,7 +149,7 @@ static int set_bucket(const char* arg);
 static int my_fuse_opt_proc(void* data, const char* arg, int key, struct fuse_args* outargs);
 static fsblkcnt_t parse_bucket_size(char* value);
 static bool is_cmd_exists(const std::string& command);
-static int print_umount_message(const std::string& mp, bool force);
+static int print_umount_message(const std::string& mp, bool force) __attribute__ ((unused));
```
ggtakec, line 152: In the case of `__MSYS__` it is `unused`, but in other cases this is used.
For now it's fine as it is, but it may be better to use ifdef(etc) on `__MSYS__` in the future. (However, this will make the code harder to read.)

ggtakec (approved):

PR #2506, opened 2024-07-26
Fix Windows compilation

siketyan:
#2503 

### Details
This pull request updates the outdated compilation instructions for Windows + MSYS2 and fixes minor issues that cause compile errors.


ggtakec (approved):
Thanks for researching and PR.

PR #2510, opened 2024-08-08
s3 signv4 support pefixed s3 hosts

fenos:

When trying to connect to a compatible S3 server that is serving the S3 API under a URI prefix, ex: 

```
https://storage.example/s3
```

`s3fs` fails to compute the SignV4; specifically on capturing the URI of the s3host.

### Details

This PR will instead use the prefix (if any) when calculating the SignV4 uri; enabling support for custom S3 providers served on a URI.

If no prefix is given in the custom `url`, the current behaviour of using "/" is maintained.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -2900,6 +2900,37 @@ std::string S3fsCurl::CalcSignature(const std::string& method, const std::string
     return s3fs_hex_lower(md.get(), md_len);
 }
 
+std::string S3fsCurl::extractURI(const std::string& url) {
+    // If the URL is empty, return "/"
+    if (url.empty()) {
+        return "/";
+    }
+
+    // Find the position of "://"
+    std::size_t protocol_pos = url.find("://");
+    if (protocol_pos == std::string::npos) {
+        // If "://" is not found, return "/"
+        return "/";
+    }
+
+    // Find the position of the first "/" after "://"
+    std::size_t uri_pos = url.find('/', protocol_pos + 3);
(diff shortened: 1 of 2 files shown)


src/curl.cpp
```diff
@@ -2944,7 +2975,7 @@ void S3fsCurl::insertV4Headers(const std::string& access_key_id, const std::stri
     get_date_sigv3(strdate, date8601);
 
     std::string contentSHA256 = payload_hash.empty() ? EMPTY_PAYLOAD_HASH : payload_hash;
-    const std::string realpath = pathrequeststyle ? "/" + S3fsCred::GetBucket() + server_path : server_path;
+    const std::string realpath = pathrequeststyle ? extractURI(s3host) + S3fsCred::GetBucket() + server_path : server_path;
```
ggtakec, line 2978: As clang-tidy points out, make this method static.

ggtakec (changes requested):
@fenos
Thank you for reporting this bug and for the fix!

I have confirmed that the fix works fine.
Clang-tidy has found a warning, so could you change the extractURI method to static?

src/curl.cpp
fenos: @ggtakec the method is now static 👍

ggtakec (approved):
Thanks, I confirmed the fix.

ggtakec:
@fenos I've merged this.
Thanks for the fix and the collaboration.

PR #2524, opened 2024-09-26
Fixed hardlink test for macos

ggtakec:
#2418

### Details
Until recently, hardlink on macos-fuse-t was not implemented by s3fs and resulted in an error, but the error output was empty.
When I noticed currently, it was outputting `"Input/output error"`.
This caused `test_hardlink` execution(which had a special branch for macos-fuse-t) to fail, so I fixed it.

Note, this error output may change in the future, but I will check again if it does.
(s3fs is returning an ENOTSUP error, and "Input/output error" seems inappropriate as the message.)

Although the `test_hardlink` test was fixed, other tests still fail, so the macos tests do not pass overall.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -784,18 +784,11 @@ function test_hardlink {
     echo foo > "${TEST_TEXT_FILE}"
 
     (
-        if ! uname | grep -q Darwin; then
-            set +o pipefail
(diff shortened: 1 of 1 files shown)


gaul:
@macos-fuse-t How can we make sure FUSE-T returns `ENOTSUPP` like Linux?

PR #2528, opened 2024-09-26
Call Rename outside AutoFdEntity scope

gaul:
This avoids a use-after-free in the destructor.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1507,10 +1507,9 @@ static int rename_object(const char* from, const char* to, bool update_ctime)
         if(0 != (result = put_headers(to, meta, true, /* use_st_size= */ false))){
             return result;
         }
-
-        // rename
-        FdManager::get()->Rename(from, to);
     }
+
+    FdManager::get()->Rename(from, to);
 
     // Remove file
     result = s3fs_unlink(from);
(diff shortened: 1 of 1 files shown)


gaul:
@ggtakec I am not too sure about this fix but it addresses symptoms seen in #2527.  Could you help me understand what is going on?  Memory should only have one owner but it seems like the cache and the AutoFdEntity both try to destroy the same objects.

ggtakec:
@gaul Thanks for raising the issue.

I think the problem is that this funtion get an `FdEntity` object using `AutoFdEntity`, and then call `FdManager::Rename` before `AutoFdEntity` is destroyed.
`FdManager::Rename` internally operates on the map of `FdEntity` objects referenced by `AutoFdEntity`.
I think this is why a conflict is occurring.

To avoid this, I think it is sufficient to simply move `FdManager::Rename` outside the scope where `AutoFdEntity` exists.
Note that this correction is necessary not only for `rename_object()` but also for `rename_object_nocopy()`.

Below are the differences that need to be corrected:
```
diff --git a/src/s3fs.cpp b/src/s3fs.cpp
index 1503a46..40cb6fd 100644
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1507,10 +1507,9 @@ static int rename_object(const char* from, const char* to, bool update_ctime)
         if(0 != (result = put_headers(to, meta, true, /* use_st_size= */ false))){
             return result;
         }
-
-        // rename
-        FdManager::get()->Rename(from, to);
     }
+    // rename
+    FdManager::get()->Rename(from, to);

     // Remove file
     result = s3fs_unlink(from);
@@ -1562,8 +1561,9 @@ static int rename_object_nocopy(const char* from, const char* to, bool update_ct
             S3FS_PRN_ERR("could not upload file(%s): result=%d", to, result);
             return result;
         }
-        FdManager::get()->Rename(from, to);
     }
+    // rename
+    FdManager::get()->Rename(from, to);

     // Remove file
     result = s3fs_unlink(from);
```

ggtakec (approved):

PR #2530, opened 2024-09-28
Enable readability-implicit-bool-conversion

gaul:
This fixes one real error, one misreported `EPERM`, and some false positives.  References #2529.


ggtakec (approved):
Thanks!

Issue #2205, opened 2023-07-02
Summarize inspection with FUSE-T(macOS)


ggtakec:
Currently testing to use macos-fuse-t instead of osxfuse(macfuse) - #2198
I would like to summarize this in this issue.

ggtakec:
I'm currently testing(fuse-t: 1.0.20), but there are some things we need to check.
_(Because the investigation has not been completed yet, there is a high possibility that it is not a problem in combination with fuse-t, but a problem with s3fs alone, so please be careful.)_

#### # Owners of directories and files to be created
It seems that it will be always created by root user.

#### # On rare occasions, an error occurred in directory listings and some directories could not be displayed.  
At this time the following error is displayed:  
```
ls: fts_read: Input/output error
```  
It seems that this is a symptom similar to #2182 and #2190.  
I think it's likely an s3fs issue, not due to the combination with fuse-t.  

#### # timestamp(atime/ctime/mtime) seems to behave incorrectly  
I haven't been able to confirm it at all, but it may be because it is via NFS (Is there a possibility?)
We need to check logs.  
The following tests are failing:
- test_mtime_file
- test_update_time_touch_a
- test_update_directory_time_chown
- test_update_directory_time_set_xattr
- test_update_directory_time_touch_a
- test_update_parent_directory_time

alexfs:
test_mtime_file, test_update_time_touch_a and test_update_directory_time_touch_a are not going to work because the macos nfs server has the following bug: it's not possible to update access time separately from modified time, only together.
Other tests should work. I tested chown and xattr commands manually and it worked correctly.

PR #2531, opened 2024-09-28
Modified and bypassed some MacOS tests

ggtakec:
#2418 #2205 #2198

### Details
Currently, we still unable to pass the tests on MacOS using macos-fuse-t.

After some investigation, I knew again that (as previously pointed out) our test cases were failing to update(or get) attributes such as ctime/mtime/atime.
_We have disabled caching of attributes in macos-fuse-t and the NFS it uses (noattrcache), but we've heard before that this doesn't stop the caching completely._
I tried waiting a few second to try to cache out, but the test pass/fail is not consistent.

In light of this, I have temporarily disabled the following tests on macos.
- test_update_directory_time_set_xattr
- test_update_directory_time_subdir
- test_update_parent_directory_time

Also, since the `test_truncate_cache` test generates an error `ls: fts_read: Input/output error`, I have changed this test waits 1 second for a solution to that problem. (This change is also temporary.)

I consider the test fix in this PR to be temporary, not permanent.

@gaul
Why don't you merge this?

@macos-fuse-t
If you have any updated info on these (which I think are affected by caches) please let us know.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -2050,6 +2050,15 @@ function test_truncate_cache() {
         for file in $(seq 75); do
             touch "${dir}/${file}"
         done
+
+        # FIXME:
+        # In the case of macos-fuse-t, if you do not enter a wait here, the following error may occur:
+        #    "ls: fts_read: Input/output error"
+        # Currently, we have not yet been able to establish a solution to this problem.
+        # Please pay attention to future developments in macos-fuse-t.
+        #
+        wait_ostype 1 "Darwin"
+
         ls "${dir}"
     done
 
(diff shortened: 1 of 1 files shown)


gaul:
`test_multipart_mix` failed on macos12 runner but this error may be flaky?

ggtakec:
> test_multipart_mix failed on macos12 runner but this error may be flaky?

Since I don't know what the error is, I can't say for sure, but considering the content of the `test_multipart_mix` test, it's likely unrelated to the problem this PR is trying to avoid.
I will continue to carefully check any errors that occur in the macos tests.

ggtakec:
Sorry. I checked the log.

I found the following error in `test_multipart_mix` from the log:
```
cp: big-file-s3fs-32710.txt: could not copy extended attributes to big-file-s3fs-32710.txt-mix: Result too large
```
This is the first time I've seen this error, but it seems to be related to `extended attributes`.

Thus I denied it in my previous comment, but I think it could have something to do with the underlying issue this PR avoids.
We should check if it continues in the future.

ggtakec:
This error occurs during the `nocopyapi` option test.
The process immediately before the error was output was a `list`/`remove` operation to `xattr`.
The cause will require investigation, but as of now, as with this PR, no workaround has been found, so I will create a PR to bypass it.

#### [Reference] One of log lines near this error:
```
s3fs: [INF] s3fs.cpp:s3fs_listxattr(3988): [path=/testrun-12125/big-file-s3fs-14543.txt][list=0x7fda31008200][size=16384]
s3fs: [INF] s3fs.cpp:s3fs_listxattr(3988): [path=/testrun-12125/big-file-s3fs-14543.txt-mix][list=0x7fda31008200][size=16384]
s3fs: [INF] s3fs.cpp:s3fs_removexattr(4054): [path=/testrun-12125/big-file-s3fs-14543.txt-mix][name=com.apple.FinderInfo][pid=0,uid=501,gid=20]
s3fs: [INF] s3fs.cpp:s3fs_listxattr(3988): [path=/testrun-12125/big-file-s3fs-14543.txt][list=0x7fda28868200][size=16384]
s3fs: [INF] s3fs.cpp:s3fs_listxattr(3988): [path=/testrun-12125/big-file-s3fs-14543.txt][list=0x7fda29017400][size=16384]
cp: big-file-s3fs-14543.txt: could not copy extended attributes to big-file-s3fs-14543.txt-mix: Result too large
```

PR #2535, opened 2024-09-29
Fix issues discovered by Coverity

gaul:


src/fdcache.cpp
```diff
@@ -414,6 +414,7 @@ FILE* FdManager::MakeTempFile() {
     strncpy(cfn, fn.c_str(), sizeof(cfn) - 1);
     cfn[sizeof(cfn) - 1] = '\0';
 
+    umask(0600);
```
ggtakec, line 417: I think mkstemp creates a file with attributes of 0600, so why is the umask being used to enforce the attributes here?
Also, if you are setting a umask, I think you need to set it back to the original mask value before exiting this function.

test/cr_filename.cc
```diff
@@ -49,15 +49,16 @@ int main(int argc, const char *argv[])
         fprintf(stderr, "[ERROR] Could not open file(%s)\n", filepath);
         exit(EXIT_FAILURE);
     }
-    close(fd);
```
ggtakec, line 52: I think the reason for calling `close` here is to execute `flush`/`release` on the open file.
I think they wanted to execute `open`/`close(flush)` before executing stat.

test/write_multiblock.cc
```diff
@@ -73,6 +73,9 @@ static std::unique_ptr<unsigned char[]> create_random_data(off_t size)
             readcnt = 0;
         }
     }
+
+    close(fd);
```
ggtakec, line 77: Thanks, this is a bug.

src/fdcache.cpp
gaul: Reverted.  Coverity give bad advice here as described by: https://github.com/cockpit-project/cockpit/issues/760

test/cr_filename.cc
gaul: Reverted.

ggtakec (approved):

PR #2538, opened 2024-10-01
Fixed double free, changed fdent_map_t member to shared_ptr

ggtakec:
#2532 #2383

### Details
Fixed a bug that caused multiple releases of the same pointer (std::unique_ptr set in std::map).

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -587,7 +587,7 @@ FdEntity* FdManager::Open(int& fd, const char* path, const headers_t* pmeta, off
             return nullptr;
         }
         // make new obj
-        std::unique_ptr<FdEntity> ent(new FdEntity(path, cache_path.c_str()));
+        std::shared_ptr<FdEntity> ent(new FdEntity(path, cache_path.c_str()));
 
         // open
         if(0 > (fd = ent->Open(pmeta, size, ts_mctime, flags))){
@@ -700,7 +700,7 @@ void FdManager::Rename(const std::string &from, const std::string &to)
         // found
         S3FS_PRN_DBG("[from=%s][to=%s]", from.c_str(), to.c_str());
 
-        std::unique_ptr<FdEntity> ent(std::move(iter->second));
+        std::shared_ptr<FdEntity> ent(std::move(iter->second));
 
         // retrieve old fd entity from map
         fent.erase(iter);
@@ -750,10 +750,17 @@ bool FdManager::Close(FdEntity* ent, int fd)
     return false;
 }
 
-bool FdManager::ChangeEntityToTempPath(FdEntity* ent, const char* path)
+bool FdManager::ChangeEntityToTempPath(const FdEntity* ent, const char* path)
 {
     const std::lock_guard<std::mutex> lock(FdManager::except_entmap_lock);
-    except_fent[path] = ent;
+
+    // Search raw FdEntity pointer in map and copy std::shared_ptr.
+    for(fdent_map_t::const_iterator iter = fent.begin(); iter != fent.end(); ++iter){
+        if(iter->second.get() == ent){
+            except_fent[path] = iter->second;   // copy
+        }
+    }
+
     return true;
 }
 
@@ -764,12 +771,12 @@ bool FdManager::UpdateEntityToTempPath()
 {
     const std::lock_guard<std::mutex> lock(FdManager::except_entmap_lock);
 
-    for(fdent_direct_map_t::iterator except_iter = except_fent.begin(); except_iter != except_fent.end(); ){
+    for(fdent_map_t::iterator except_iter = except_fent.begin(); except_iter != except_fent.end(); ){
         std::string tmppath;
         FdManager::MakeRandomTempPath(except_iter->first.c_str(), tmppath);
 
         fdent_map_t::iterator iter = fent.find(except_iter->first);
-        if(fent.end() != iter && iter->second.get() == except_iter->second){
+        if(fent.end() != iter && iter->second.get() == except_iter->second.get()){
             // Move the entry to the new key
             fent[tmppath] = std::move(iter->second);
             iter          = fent.erase(iter);
@@ -785,7 +792,7 @@ bool FdManager::UpdateEntityToTempPath()
             S3FS_PRN_WARN("For some reason the FdEntity pointer(for %s) is not found in the fent map. Recovery procedures are being performed, but the cause needs to be identified.", except_iter->first.c_str());
 
             // Add the entry for recovery procedures
-            fent[tmppath] = std::unique_ptr<FdEntity>(except_iter->second);
+            fent[tmppath] = std::move(except_iter->second);
             except_iter   = except_fent.erase(except_iter);
         }
     }
--- a/src/fdcache.h
+++ b/src/fdcache.h
@@ -47,7 +47,7 @@ class FdManager
       static std::string     tmp_dir;
 
       fdent_map_t            fent;
-      fdent_direct_map_t     except_fent;           // A map of delayed deletion fdentity
+      fdent_map_t            except_fent;           // A map of delayed deletion fdentity
 
   private:
       static off_t GetFreeDiskSpaceHasLock(const char* path) REQUIRES(FdManager::reserved_diskspace_lock);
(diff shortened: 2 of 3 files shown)


src/fdcache.cpp
```diff
@@ -739,6 +739,16 @@ bool FdManager::Close(FdEntity* ent, int fd)
                 // check another key name for entity value to be on the safe side
                 for(; iter != fent.end(); ){
                     if(iter->second.get() == ent){
+                        // [NOTE]
+                        // Since "ent" has already been freed just before this loop,
+                        // it must call "release()" here without being freed.
+                        // The return value of release() can be ignored, but clang-tidy
+                        // will warn.
+                        //
+                        const FdEntity* enttmp = iter->second.release();
```
gaul, line 748: How do we have a situation where one `FdEntity*` is associated with different paths?  This seems like a scary situation since the owner of memory is the container.  Should this be a `shared_ptr` instead?

ggtakec: I can't remember exactly, but I think it had something to do with renaming.
For example, I think it was something like renaming from another process while uploading a file.
I'll review this area after the next release. (I'm currently working on another improvement, and it will conflict a little with that, so I don't want to review it now.)

VVoidV: https://github.com/s3fs-fuse/s3fs-fuse/pull/2532#issuecomment-2387518812 
This is one possible path

gaul: if s3fs has two `unique_ptr` that have the same raw pointer then there is a problem.  The crash is telling us that an invariant is broken.  Manual `release` calls are a workaround for a logic bug.  I think we should remove the first path and `std::move` the `unique_ptr` before inserting the second path.  If it is actually required to have two paths with the same raw pointer then we should use a `shared_ptr` instead.

VVoidV: I believe that when commit 1a50b9a04a82678b05e36927c323f42d94ca4a07 resolving the deadlock in ChangeEntityToTempPath, the scenario of using the tmp directory as cache was not taken into account. There shouldn't be two unique_ptrs pointing to the same raw pointer.

ggtakec: @VVoidV @gaul Thank you for your comment.
My understanding was not sufficient, so I changed `unique_ptr` to `shared_ptr` and reviewed the code.
Please review again with the revised code.

src/fdcache.cpp
```diff
@@ -754,7 +754,8 @@ bool FdManager::Close(FdEntity* ent, int fd)
 bool FdManager::ChangeEntityToTempPath(FdEntity* ent, const char* path)
 {
     const std::lock_guard<std::mutex> lock(FdManager::except_entmap_lock);
-    except_fent[path] = ent;
+    except_fent[path] = std::shared_ptr<FdEntity>(ent);
```
gaul, line 757: Do we have any kind of test that demonstrate the existing double delete?  I worry that there will be two separate `std::shared_ptr` each with reference count = 1 that will continue to double-delete the raw pointer.  For this to work we should expect a single `std::shared_ptr` with reference count = 2.  The crashes we are seeing are demonstrating that s3fs has a memory ownership misunderstanding -- where are we creating the second reference to the same raw pointer?  Ideally whatever allocates the `FdEntity` should wrap the raw pointer in an owned pointer (either `std::unique_ptr` or `std::shared_ptr` and all reference should go through that.

VVoidV: @ggtakec  
The current issue is that two shared_ptr are referencing the same raw pointer. I checked out your PR for verification, and the problem still persists.
When the deadlock occurs, the following content is printed in the logs:
> s3fs: [WAN] fdcache.cpp:UpdateEntityToTempPath(787): For some reason the FdEntity pointer(for /200M) is not found in the fent map. Recovery procedures are being performed, but the cause needs to be identified.

## To reporduce the deadlock, you could fellow these steps
1. modify small-integration-test.sh

```
diff --git a/test/small-integration-test.sh b/test/small-integration-test.sh
index 6a1029f..5f22388 100755
--- a/test/small-integration-test.sh
+++ b/test/small-integration-test.sh
@@ -48,18 +48,18 @@ export CACHE_DIR
 export ENSURE_DISKFREE_SIZE 
 if [ -n "${ALL_TESTS}" ]; then
     FLAGS=(
-        "use_cache=${CACHE_DIR} -o ensure_diskfree=${ENSURE_DISKFREE_SIZE} -o fake_diskfree=${FAKE_FREE_DISK_SIZE} -o use_xattr -o update_parent_dir_stat"
-        enable_content_md5
-        disable_noobj_cache
-        "max_stat_cache_size=100"
-        nocopyapi
-        nomultipart
-        sigv2
-        sigv4
-        "singlepart_copy_limit=10"  # limit size to exercise multipart code paths
-        #use_sse  # TODO: S3Proxy does not support SSE
-        #use_sse=custom:/tmp/ssekey  # TODO: S3Proxy does not support SSE
-        "use_cache=${CACHE_DIR} -o ensure_diskfree=${ENSURE_DISKFREE_SIZE} -o fake_diskfree=${FAKE_FREE_DISK_SIZE} -o streamupload"
+        "ensure_diskfree=${ENSURE_DISKFREE_SIZE} -o fake_diskfree=${FAKE_FREE_DISK_SIZE}"
+        # enable_content_md5
+        # disable_noobj_cache
+        # "max_stat_cache_size=100"
+        # nocopyapi
+        # nomultipart
+        # sigv2
+        # sigv4
+        # "singlepart_copy_limit=10"  # limit size to exercise multipart code paths
+        # #use_sse  # TODO: S3Proxy does not support SSE
+        # #use_sse=custom:/tmp/ssekey  # TODO: S3Proxy does not support SSE
+        # "use_cache=${CACHE_DIR} -o ensure_diskfree=${ENSURE_DISKFREE_SIZE} -o fake_diskfree=${FAKE_FREE_DISK_SIZE} -o streamupload"
     )
 else
     FLAGS=(
```

2.  start s3fs with following ENV
```
ALL_TESTS=1 INTERACT=1 DBGLEVEL=debug ./small-integration-test.sh
```

3. cp a 180MB file in s3fs dir

ggtakec: I’m sorry for this code still has problems, so I will update it after checking again.

@gaul
I will check if I can prepare test cases etc.
(Additinal comment: probably my previous comment(the reason is rename) was wrong.)
However, I think it is possible that the same FdEntity is assigned to multiple keys depending on the timing of processing.
As reported by @VVoidV, I think it occurs when the file cache is unused and disk space is insufficient.
(And I think the problem is that the FdEntity is deadlocked when it is closed. The fix in #2478 was insufficient.)
When the cache space is insufficient, the fdentity map's key is registered as a string that is not a file path, and then when the capacity runs out again, I think that there is a possibility that the same FdEntity will be registered to multiple keys with a later call to `UpdateEntityToTempPath()`.
In any case, the current this PR code still needs to be revised, so please review it again after the revisions are made.

@VVoidV
Thank you. I have confirmed the deadlock.
I'll update this PR code, please wait a little longer.

ggtakec: @gaul @VVoidV 
This source code has been slightly updated. Please review again with the new code.

ggtakec:
@gaul @VVoidV 
I've posted an additional PR #2539.
The PR #2539 makes debugging out-of-disk-space issues easier (but you still need to configure it with care).
I think getting it merged will help move this PR forward.
See the [comments ](https://github.com/s3fs-fuse/s3fs-fuse/pull/2539#issuecomment-2395350324) in #2539 for more details.

src/fdcache.cpp
```diff
@@ -750,10 +750,17 @@ bool FdManager::Close(FdEntity* ent, int fd)
...
 {
     const std::lock_guard<std::mutex> lock(FdManager::except_entmap_lock);
-    except_fent[path] = ent;
+
+    // Search raw FdEntity pointer in map and copy std::shared_ptr.
+    for(fdent_map_t::const_iterator iter = fent.begin(); iter != fent.end(); ++iter){
+        if(iter->second.get() == ent){
+            except_fent[path] = iter->second;   // copy
+        }
+    }
```
gaul, line 762: @ggtakec could you look at #2541?  I believe that using `shared_from_this` will remove the need for this lookup.

ggtakec:
@gaul 
I would like to close this PR and move on to #2541.

PR #2539, opened 2024-10-06
Fixed fake_diskfree option

ggtakec:
( #2538 is weakly related issues )

### Details
The value of the `fake_diskfree` option was affected by the values ​​of the `free_space_ratio` and `ensure_diskfree` options, and in some cases it could not be set.

Originally, `fake_diskfree` was an option for debugging s3fs, setting a pseudo amount of free disk space, so it should not have been affected by the results of the check of the amount of free space required when s3fs was started.

If the amount of free space was below the minimum required for multipart uploads, etc., it was unable to be set, but we have changed it so that it can be set without being affected by this condition.

When making this correction, we also found an error in the mutual exclusion control (mutex), so we have corrected that at the same time.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -237,9 +237,8 @@ bool FdManager::CheckCacheDirExist()
     return IsDir(cache_dir);
 }
 
-off_t FdManager::GetEnsureFreeDiskSpace()
+off_t FdManager::GetEnsureFreeDiskSpaceHasLock()
 {
-    const std::lock_guard<std::mutex> lock(FdManager::reserved_diskspace_lock);
     return FdManager::free_disk_space;
 }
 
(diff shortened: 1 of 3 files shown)


ggtakec:
### Additional Information for usage
To debug cases where there is insufficient free disk space for multipart uploads, you can specify the following options, for example:
```
multipart_size=20,fake_diskfree=40
```
This will start s3fs with 20MB of size for each multipart upload and 40MB of free disk space.
For example, if you specify this option and upload a file of about 100MB, FdEntity::NoCacheLoadAndPost() will be called, allowing you to debug the behavior when there is insufficient disk space.
Please note that you may need to change the settings of these options depending on the situation, such as when you specify a cache directory.

PR #2541, opened 2024-10-12
Use `std::shared_ptr` to refer to `FdEntity`

gaul:
`FdEntity` may have multiple references due to `ChangeEntityToTempPath`.  This relies on the `std::enable_shared_from_this` helper to create a `std::shared_ptr` from `this`.  Fixes #2532.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -587,7 +587,7 @@ FdEntity* FdManager::Open(int& fd, const char* path, const headers_t* pmeta, off
             return nullptr;
         }
         // make new obj
-        std::unique_ptr<FdEntity> ent(new FdEntity(path, cache_path.c_str()));
+        auto ent = std::make_shared<FdEntity>(path, cache_path.c_str());
 
         // open
         if(0 > (fd = ent->Open(pmeta, size, ts_mctime, flags))){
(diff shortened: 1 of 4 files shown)


gaul:
Only a draft since I still don't understand this problem well.  @VVoidV can you provide a test case?

VVoidV:
> Only a draft since I still don't understand this problem well. @VVoidV can you provide a test case?
Here is the case. 
https://github.com/s3fs-fuse/s3fs-fuse/pull/2532#issuecomment-2387518812

1. do not specify cache dir, s3fs will use /tmp, and the key in the fdmap will not be the real path.
2. trigger the not enough cache case.

ggtakec:
@gaul
I will close #2532 and move forward with this PR, so could you please remove the Draft?

> Only a draft since I still don't understand this problem well. @VVoidV can you provide a test case?

I think the problem is that the cache file path is changed to a temporary path due to insufficient cache disk space, which causes multiple releases of FdEntity managed by FdManager.

Testing for this issue requires running out of disk space.
Before #2539, it was difficult to artificially reduce the free space on the cache disk, but after merging it, you can simulate it just by specifying the `fake_diskfree` option.

For example, set the multipart size to 20M and the free disk space to 40M as follows.
```
multipart_size=20,fake_diskfree=40
```
In this state, if you run a multipart upload of an 80MB file (it should be fine if it's more than 40MB), the logic of the problem being fixed in this PR will be executed.
(FdEntity::NoCacheLoadAndPost() will be executed)

ggtakec (approved):

Issue #254, opened 2015-08-27
s3fs is leaking memory for me


monken:
I installed s3fs from master like this

``` bash
git clone https://github.com/s3fs-fuse/s3fs-fuse
cd s3fs-fuse/
git checkout cfdfecb4d17b52cdd88faaea53c64f098d0cceff
./autogen.sh
./configure --prefix=/usr --with-nss
make
make install
```

on a AWS Linux EC2 instance. libcurl is linked against nss on this machine.

I have mounted a bucket like this

```
s3fs -o url=https://s3-eu-west-1.amazonaws.com -o use_sse=1 -o default_acl=bucket-owner-full-control -o enable_content_md5 -o endpoint=eu-west-1 testbucket testbucket
```

The bucket has roughly 4000 files in it. Each `find testbucket/ | wc -l` grows the s3fs process by ~60mb.

Any advice would be highly appreciated!

andrewgaul:
@monken Can you provide `ps u` output before and after your `find` command?  I measured this when running find on 700 files in my bucket:

```
$ ps u `pidof s3fs`
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
gaul     31252  1.3  0.0 358172  7672 pts/28   Sl+  13:49   0:00 src/s3fs

$ ps u `pidof s3fs`
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
gaul     31252  1.1  0.1 1775412 9996 pts/28   Sl+  13:49   0:01 src/s3fs

$ ps u `pidof s3fs`
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
gaul     31252  0.8  0.1 1775412 9996 pts/28   Sl+  13:49   0:01 src/s3fs
```

Note that RSS only grew by 2 MB while VSZ grew by 140 MB.  The latter is not actually in use, just virtual address space reserved by the process.

monken:
Right, the virtual address space is not what I'm concerned about. The RSS grew by ~60mb to a point where the server started swapping and eventually killed the s3fs process.

andrewgaul:
@monken Sorry I cannot reproduce your symptoms.  Can you provide a test case, for example, something like:

```
for i in `seq 4000`; do touch /path/to/mountpoint/$i
ps u `pidof s3fs`
find /path/to/mountpoint
ps u `pidof s3fs`
```

[...]

PR #2545, opened 2024-10-13
Fixed exclusive control of upload_id in PseudoFdInfo class

ggtakec:
#2493 #2487 #254

### Details
This was modified to minimize exclusive control of the `PseudoFdInfo::upload_id` variable.
To achieve this, `IsUploadingHasLock()` is added, which does not hold the `IsUploading()` lock, to streamline the locking process on the caller side.
As a result of this efficiency, it is no longer necessary to lock before calling `ParallelMultipartUpload()`.
Please refer to the code for detailed differences.

### DIFF
--- a/src/fdcache_fdinfo.cpp
+++ b/src/fdcache_fdinfo.cpp
@@ -107,7 +107,7 @@ void* PseudoFdInfo::MultipartUploadThreadWorker(void* arg)
 //------------------------------------------------
 // PseudoFdInfo methods
 //------------------------------------------------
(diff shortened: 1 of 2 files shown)


gaul (changes requested):
Can you add `GUARDED_BY` to the member variables like #2549?  This will improve the quality and make the PR easier to review.

ggtakec:
I've applied `GUARDED_BY` to a member variable and also added some methods to make it easier to understand.

PR #2546, opened 2024-10-13
Ensure that test checks data length

gaul:
`wc` has an optimization that can use metadata when stdin is set to a file.  Also fix up logging.

### DIFF
--- a/test/test-utils.sh
+++ b/test/test-utils.sh
@@ -125,20 +125,21 @@ function check_file_size() {
     local FILE_NAME="$1"
     local EXPECTED_SIZE="$2"
 
-    # Verify file is zero length via metadata
+    # Verify file length via metadata
     local size
     size=$(get_size "${FILE_NAME}")
     if [ "${size}" -ne "${EXPECTED_SIZE}" ]
     then
-        echo "error: expected ${FILE_NAME} to be zero length"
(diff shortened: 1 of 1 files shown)


test/test-utils.sh
```diff
@@ -125,20 +125,21 @@ function check_file_size() {
...
-        echo "error: expected ${FILE_NAME} to be zero length"
+        echo "error: expected ${FILE_NAME} to be ${EXPECTED_SIZE} length but was ${size} via metadata"
         return 1
     fi
 
-    # Verify file is zero length via data
-    size=$(wc -c < "${FILE_NAME}")
+    # Verify file length via data
+    # shellcheck disable=SC2002
+    size=$(cat "${FILE_NAME}" | wc -c)
```
ggtakec, line 139: You can avoid SC2002 by writing the following. (We can also remove  `shellcheck disable=SC2002` comment.)
```
wc -c < "${FILE_NAME}"
```

ggtakec (changes requested):

test/test-utils.sh
gaul: The problem here is that `wc -c` will use lseek on the file instead of reading all the bytes in the pipe to determine the length.

ggtakec: Thank you. I understand.

ggtakec (approved):

PR #2559, opened 2024-10-21
Add OpenSSL suppression for ThreadSanitizer

gaul:


gaul:
This seemed to work locally but not in CI?

gaul:
Any idea why setting `TSAN_OPTIONS` makes configure fail?  This option should not be set then.

ggtakec:
> This seemed to work locally but not in CI?
> Any idea why setting TSAN_OPTIONS makes configure fail? This option should not be set then.

Were these two issues resolved?
Great. Now it's easier to understand the CI checks.

#### PS
I'm currently trying to resolve this `OPENSSL_sk_free` issue(it not only but also there are other purposes as well).
However, it's a fairly big change, so it won't be ready in time for v1.95 (we'll make a PR after 1.95).

ggtakec (approved):

PR #2228, opened 2023-07-22
Insert SSE headers when appropriate

gaul:
References #2218.  References #2227.

PR #2565, opened 2024-10-22
Remove some unused parameters

gaul:


gaul:
@ggtakec could you check the intent of these methods?  I am unsure if these are supposed to be used somehow.

ggtakec:
> could you check the intent of these methods? I am unsure if these are supposed to be used somehow.

`S3fsCurl::PreMultipartPostRequest()` is a pre-processing for multipart upload.
The `is_copy` flag was used up to `v1.93`, but when SSE-related fixes were made, the `x-amz-server-side-encryption-customer-key-md5` header was deemed unnecessary in this pre-processing and was removed.(But it seems that the fix was missed at that time.)
See #2228.

`FdEntity::LoadAll()` is used to load all the contents of an object.
Before the `PseudoFdInfo` class was introduced, `Load` was performed at the same time as `Open`, and the `headers_t` variable was required for `Open`.
However, although the `Open` process was separated in #1666, it seems that the parameter correction was missed.

#### PS
I'm currently try to modify these logic(although there will be no major changes), I will check to see if there are any similarities.

ggtakec (approved):

PR #2566, opened 2024-10-22
Separate serialization and deserialization code

gaul:
This is clearer than a bool parameter.


ggtakec (commented):
Just a suggestion, how about making the `original Serialize method` private, preparing separate methods for `another Serialize` and `Deserialize` which are wrapping the `original Serialize`?

gaul:
> Just a suggestion, how about making the `original Serialize method` private, preparing separate methods for `another Serialize` and `Deserialize` which are wrapping the `original Serialize`?

Done.

ggtakec (approved):

PR #2567, opened 2024-10-22
Separate clang-tidy into its own CI task

gaul:
clang-tidy takes 4 minutes on my laptop compared to ALL_TESTS=1 which takes 8 minutes.  Using a separate tasks avoids duplicating clang-tidy and unnecessarily slowing CI run-time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,6 @@ on:
 jobs:
   Linux:
     runs-on: ubuntu-latest
-
     #
     # build matrix for containers
     #
@@ -113,14 +112,6 @@ jobs:
           ./autogen.sh
           /bin/sh -c "./configure ${CONFIGURE_OPTIONS}"
           make --jobs=$(nproc)
-
-      - name: clang-tidy
-        run: |
-          # skip if clang-tidy does not exist, e.g., CentOS 7
-          if command -v clang-tidy; then
-            make -C src/ clang-tidy
-            make -C test/ clang-tidy
-          fi
 
       - name: Cppcheck
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -300,6 +291,31 @@ jobs:
...
+    container:
+      image: fedora:40
+
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v4
+
+      - name: Install packages
+        run: |
+          .github/workflows/linux-ci-helper.sh fedora:40
```
ggtakec, line 305: This line should be `.github/workflows/linux-ci-helper.sh ${{ matrix.container }}`

ggtakec (changes requested):

gaul:
I experimented with `run-clang-tidy` locally which can run in parallel (10x faster on my laptop): https://github.com/gaul/s3fs-fuse/tree/run-clang-tidy .  Unfortunately this requires a compilation database that I don't fully understand.

.github/workflows/ci.yml
gaul: I tried matrix.container and container.image but neither worked.

ggtakec: Ok, something's weird, but I'll merge this and check it later.

ggtakec (approved):

.github/workflows/ci.yml
ggtakec: @gaul I'm sorry about my comment, so blow is additional information.
Since this JOB does not use a matrix, I confirmed that the container name cannot be obtained as a GHA variable.
(I was able to find `FBR=f40`, `DISTTAG=f40container`, `FGC=f40` in the environment variables, but these should not be used.)
So there was no need to modify it as it was.

ggtakec:
@gaul 
Could it be the following?
- https://github.com/rizsotto/Bear
- https://github.com/Sarcasm/compdb
I don't have complete knowledge about these yet either.

However, on CI, clang-tidy is now run in a separate JOB due to your PR, so I think it will complete faster than other builds and tests running in parallel.

gaul:
Yeah I use `bear` in that branch since automake lacks native support.  But `run-clang-tidy` does not do quite the same thing of comping the headers as well as the source.  So I'll have to look at this more closely.  But if you are experimenting with different checks, that branch is a lot faster!

PR #2569, opened 2024-10-24
Make psemaphore similar to C++20 `std::counting_semaphore`

gaul:


src/sighandlers.cpp
```diff
@@ -113,9 +113,12 @@ void S3fsSignals::CheckCacheWorker(Semaphore* pSem)
         }
 
         // do not allow request queuing
+        // TODO: this appears wrong -- value is constant
+        /*
         for(int value = pSem->get_value(); 0 < value; value = pSem->get_value()){
-            pSem->wait();
+            pSem->acquire();
```
gaul, line 119: What is the intent here?  How does the loop make progress?

gaul:
The advantage of this PR is that we might be able to remove the `#ifdef __APPLE__` depending on what version of Xcode is installed on older macOS.  Linux would continue to use the POSIX semaphores.

src/sighandlers.cpp
ggtakec: s3fs tries not to block signals while processing them. (Because there is a possibility that even if it does block, it may not be completely effective.)
As a result, multiple signals may be received during this thread processing, and the Semaphore count may increase.
Then, this loop is the process to clear those queued signals.

**BUT** this code is incorrect.
It should loop until `Semaphore::try_wait()` returns `false`.
```
for(int value = pSem->get_value(); 0 < value; value = pSem->get_value()){
    if(!pSem->try_wait()){
        break;
    }
}
```

Could you fix that in this PR?
(If it should be a separate PR, I'll post a PR for that fix)

gaul: This also doesn't seem correct -- value never changes.  Can you test this and post a PR?

ggtakec: The problem was the usage of get_value() too.
Simply loop `Semaphore::try_wait()`.
I have promoted this fix #2573, so please check it out.

gaul: Let's delay merging this until after 1.95 release.

ggtakec: OK.

ggtakec (approved):

PR #2578, opened 2024-10-28
Use pass-by-value for peeloff

gaul:
This avoids copies when used with `std::move`.


ggtakec (approved):

ggtakec (commented):
It looks like `test_string_util` is failing.

ggtakec (approved):

PR #2579, opened 2024-10-28
Remove CentOS 7 from CI

gaul:
CentOS 7 is EOL and thus unsupported.  This reverts commit 44d5b5e1c97d034e4a862d62af44b74ef7b8604f.


ggtakec (approved):
End of CentOS 7 build support

PR #2582, opened 2024-10-30
Fixed a bug when changing fdentity to a temporary path

ggtakec:
#2532

### Details
If the path element does not exist in fent, it may be because a cache directory has not been specified, or FdEntity::NoCacheLoadAndPost has already been called.
In these cases, the path element(=ent) has already been registered as a TempPath element from fent, so there is no need to register ent in the except_fent map.
(Processing with UpdateEntityToTempPath should not be performed.)


gaul:
Is this still a draft?

ggtakec:
@gaul I was waiting for confirmation of test with #2532.
However, let's merge this first without waiting for confirmation.
I removed the draft for this.

gaul:
CI seems stuck.

PR #2587, opened 2024-11-03
Fixed readdir bug with objects receiving EPERM on HEAD request

ggtakec:
#2535

### Details
When listing a directory that contains objects encrypted with KMS, the listing itself fails, and the following message appears:
```
$ ls
ls: reading directory .: Operation not permitted
```
This is because if KMS is not specified as SSE or a different KMS ID is specified, the HEAD request for the target object returns EPERM.
This phenomenon occurs because the fix in #2535 changed it to return a multipart request error code.
The fix itself is not wrong, but the problem was realized because the code before the fix did not handle EPERM exceptions.

The problem has been fixed by this PR.
However, since CI testing of SSE (including KMS) is not currently possible, no test code has been added.
(This needs to be considered in the future, but it may be difficult at the moment. It may be possible to test by preparing an object with a pseudo encryption header, but this has not been verified.)

@gaul 
This issue occurs when you specify a different key (or do not specify one) for an encrypted object.
This issue may be considered a critical bug.

### DIFF
--- a/src/curl_multi.cpp
+++ b/src/curl_multi.cpp
@@ -165,7 +165,22 @@ int S3fsMultiCurl::MultiPerform()
         auto it = threads.find(thread_id);
         it->second.first.join();
         auto int_retval = it->second.second.get();
-        if (int_retval && !(int_retval == -ENOENT && isMultiHead)) {
+
+        // [NOTE]
+        // For multipart HEAD requests(ex. readdir), we need to consider the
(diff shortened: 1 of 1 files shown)


gaul:
This makes me sad that s3fs lacks encrypted CI testing currently.  I am working on some S3Proxy changes that might help with this but it is not an immediately solution.

ggtakec:
@gaul Thank you. I'm looking forward to s3proxy.

PR #2588, opened 2024-11-03
Enable clang-analyzer

gaul:
Also fix a few smaller issues.


gaul:
Note that this increases clang-tidy run time a lot but that's OK since we separated the task.

ggtakec (approved):

Issue #2552, opened 2024-10-16
macOS 12 runner image will be removed by 12/3/2024


gaul:
The macOS 12 runner image will be removed by December 3rd, 2024. To raise awareness of the upcoming removal, jobs using macOS 12 will temporarily fail during scheduled time periods defined below:

* November 4, 9:00 AM - 7:00 PM EST
* November 11, 9:00 AM - 7:00 PM EST
* November 18, 9:00 AM - 7:00 PM EST
* November 25, 9:00 AM - 7:00 PM EST

What you need to do

Jobs using the macos-12 YAML workflow label should be updated to macos-15, macos-14, macos-13, or macos-latest. You can always get up-to-date information on our tools by reading about the software in the [runner images](https://github.com/actions/runner-images?elqTrackId=52d12286b0ab49208138b8f06e3af7e9&elq=d9ca020bc2074e2c800b4ada53798ae8&elqaid=4233&elqat=1&elqCampaignId=4442&elqak=8AF539AA690C65F019049E0E1A8B640381522B665242355DEA6735AE30E1F2538DEE) repository. Please contact GitHub [Support](https://support.github.com/contact?form%5Bsubject%5D=Re%3A+GitHub+Actions&tags=dotcom-contact-params&elqTrackId=7a78e8710a0c4d378d4de10f9828752e&elq=d9ca020bc2074e2c800b4ada53798ae8&elqaid=4233&elqat=1&elqCampaignId=4442&elqak=8AF592F31F339E082BFF97F54F17F450F2B02B665242355DEA6735AE30E1F2538DEE) if you run into any problems or need help.

ggtakec:
I noticed this notice too.
After 1.95 is released, we should try 13->14->15.

ggtakec:
@gaul 
In the past few days (I think yesterday), there has been a change in the GHA Runner for macos-13, and an error has started to occur.

```
configure: error: Package requirements (fuse-t >= 1.0.20 libcurl >= 7.0 libxml-2.0 >= 2.6 libcrypto >= 0.9 ) were not met:

Package 'ldap', required by 'libcurl', not found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables DEPS_CFLAGS
and DEPS_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
```
It seems to depend on pkg-config for curl or ldap.

[...]

PR #2591, opened 2024-11-04
Changed the macOS Github Actions runner image to macos-13

ggtakec:
#2552 

### Details
This has already started, and Github Action executions for `macos-12` have begun to be automatically canceled.
_(November 4, 14:00 UTC - November 5, 00:00 UTC in [this page](https://github.com/actions/runner-images/issues/10721))_
With this PR, we can migrate to macos-13.

_I tried to migrate directly to `macos-14`, but received a build error, so I will stick with upgrading to `macos-13`._

### DIFF
--- a/test/test-utils.sh
+++ b/test/test-utils.sh
@@ -435,6 +435,39 @@ function wait_ostype() {
 }
 
 #
+# Avoid extended attribute errors when copying on macos(fuse-t)
+#
+# [NOTE][FIXME]
+# This avoids an error that occurs when copying (cp command) on macos
+# (fuse-t) stating that extended attributes cannot be copied and the
+# exit code becomes anything other than 0.
+#Even if this error occurs, the copy itself is successful.
+#
+# This issue is currently(2024/11/7) still in the process of being
+# fixed, so we will wait and see.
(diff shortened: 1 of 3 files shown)


ggtakec:
In some tests, the following error has started to occur:
```
could not copy extended attributes to <file>: Result too large
```
This occurs during the copy command, and although the copy succeeds, an error occurs regarding the extended attributes.
I have added and modified the code to avoid this error case.

- NOTE This issue is still being addressed in the following issue:  
https://github.com/macos-fuse-t/fuse-t/issues/71

test/test-utils.sh
```diff
@@ -434,6 +434,39 @@ function wait_ostype() {
...
+    if uname | grep -q Darwin; then
+        if ! cp "$@"; then
+            return $?
+        fi
+    else
+        local CP_RESULT="";
+        if ! CP_RESULT=$(cp "$@" 2>&1); then
+            local CP_EXITCODE=$?
+            if ! echo "${CP_RESULT}" | grep -q -i "Result too large"; then
+                return "${CP_EXITCODE}"
```
gaul, line 462: @macos-fuse-t any ideas why this fails on newer macOS?

PR #2596, opened 2024-11-06
Require C++14

gaul:
This only has some minor additions of `std::make_unique`, digits separators, and more flexible `constexpr`.  References #2469.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -54,7 +54,6 @@ Checks: '
     -misc-use-internal-linkage,
     -modernize-avoid-c-arrays,
     -modernize-loop-convert,
-    -modernize-make-unique,
     -modernize-use-nodiscard,
     -modernize-raw-string-literal,
     -modernize-return-braced-init-list,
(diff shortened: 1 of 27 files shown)


ggtakec:
@gaul Could you please make the following correction suggested in the review comment above?

`src/gnutls_auth.cpp` (L:114)
```
auto char[]> digest = std::make_unique<unsigned char[]>(SHA1_DIGEST_SIZE);
  |
  v
auto digest = std::make_unique<unsigned char[]>(SHA1_DIGEST_SIZE);
```

`src/gnutls_auth.cpp` (L:169)
```
auto digest = std::make_unique<unsigned char[](*digestlen + 1);
  |
  v
auto digest = std::make_unique<unsigned char[]>(*digestlen + 1);
```
Currently we haven't tested gnutls with Github Actions(CI), but when It was build locally, the above statement causes an error.
Thanks in advance.

gaul:
> @gaul Could you please make the following correction suggested in the review comment above?

Done.  I also ran `test/compile_all_targets.sh` to compile all the other supported configurations.

ggtakec (approved):

PR #2597, opened 2024-11-07
Make callbacks parameters type-safe

gaul:
Instead of storing and passing void pointers, lambdas can capture additional parameters.


gaul:
Obsolete.

Issue #2600, opened 2024-11-10
Refactoring about Thread / Curl Handle


ggtakec:
### Background
Currently, s3fs-fuse has complicated and cumbersome code related to threads and curl handles due to years of modifications and feature additions.
This is not the only cause, but a data race occurs in `Curl` + `OpenSSL`(#1997 #2211 #2209 #2220), and in order to fix this, the composition of threads and curl handles needs to be changed.
I believe that we can reduce the number of objects that are deployed in unnecessary dynamic memory, and reduce the complexity of managing them.

#### Thread management is complicated
When processing a request, there are three logic steps(included using and not using threads):
- Threads are not started, and requests are processed using Curl Handle directly in the thread where the hook is called from FUSE(all early s3fs were like this)
- Threads are dynamically started and stopped, and requests are processed in the thread (only some processes from when MultipartUpload was supported)
- Threads are started when s3fs is started, pooled, and requests are processed in the threads in this pool (used in StreamUpload)

These things could be unified.
Currently it's getting more complicated and affecting maintenance.
Also, s3fs options related to this should be unified too.

#### Curl handles and OpenSSL Data Rase(#1997 #2211 #2209 #2220)
I have identified the cause of this Data Race.
It is a problem with the Curl Share handle settings/using and is related to `CURL_LOCK_DATA_SSL_SESSION`.
This `CURL_LOCK_DATA_SSL_SESSION` prohibits the Curl Handle from being used in another thread, as shown below:
(It is not supported to share SSL sessions between multiple concurrent threads)[https://curl.se/libcurl/c/CURLSHOPT_SHARE.html]

s3fs pools Curl handles, but the threads that use these Curl handles are not pinned.
Therefore, the conditions for `CURL_LOCK_DATA_SSL_SESSION` are not met, so `CURL_LOCK_DATA_SSL_SESSION` cannot be enabled as it should be.
In the current code, it is enabled, which causes a data race.

### Composition modifications to resolve
To solve this problem, we should change all request processing to run in a sub thread.
This will consolidate request processing and improve maintainability.
Also, each request can be processed synchronously or asynchronously.

We should integrate thread management into the `ThreadPoolMan` class.
This will allow all request processing to be done in parallel using a thread pool.
And it will also integrate the parallel concurrency options of s3fs.

A thread worker should hold one S3fsCurl, and the S3fsCurl should hold one Curl Handle.
This ensures that only one thread uses the Curl Handle.
[...]

PR #2602, opened 2024-11-10
Refactored single type requests to use through ThreadPoolMan

ggtakec:
#2600

### Details
This is the split PR Phase 2 (2/8) for #2600.

Single requests(synchronous requests) are changed to process by a worker thread managed by `ThreadPoolMan`.

The implementation of each request process has been moved to the handler(Callback) function of the worker thread.
As a result, the `S3fsCurl` object held by each request process function has been moved to a worker thread managed by `ThreadPoolMan`.
(In a future PR, the construction location of the `S3fsCurl` object will be moved again.)
All worker thread processing is concentrated in new `s3fs_threadreqs.cpp`  file.
(From now on, I will continue to concentrate in this file.)

Asynchronous and multi-request processing is not yet modified in this PR.

This PR will be rebased and removed from draft status once the previous PR(#2601) is merged into master.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -43,6 +43,7 @@ s3fs_SOURCES = \
     string_util.cpp \
     s3fs_cred.cpp \
     s3fs_util.cpp \
+    s3fs_threadreqs.cpp \
     fdcache.cpp \
     fdcache_entity.cpp \
     fdcache_page.cpp \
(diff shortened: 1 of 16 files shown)


gaul (approved):
This looks good.  One thing I would like to understand is whether thread pools give any benefit since most of the s3fs time is waiting for the network.  But I agree to simplify the code by always using ThreadPoolMan.

PR #2603, opened 2024-11-10
Changed multiple Head requests from S3fsMultiCurl to ThreadPoolMan

ggtakec:
#2600 

### Details
This is the split PR Phase 3 (3/8) for #2600.

Multi-Head request processing has been changed to no longer use `S3fsMultiCurl`.
Instead, it has been moved to a worker thread managed by `ThreadPoolMan`.

This PR will be rebased and removed from draft status once the previous PR(#2602) is merged into master.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -126,8 +126,6 @@ static int check_object_access(const char* path, int mask, struct stat* pstbuf);
 static int check_object_owner(const char* path, struct stat* pstbuf);
 static int check_parent_object_access(const char* path, int mask);
 static int get_local_fent(AutoFdEntity& autoent, FdEntity **entity, const char* path, int flags = O_RDONLY, bool is_load = false);
-static bool multi_head_callback(S3fsCurl* s3fscurl, void* param);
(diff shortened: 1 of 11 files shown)


src/s3fs.cpp
```diff
@@ -3148,141 +3086,25 @@ static int s3fs_opendir(const char* _path, struct fuse_file_info* fi)
...
-    // Not found Callback function parameter
-    struct multi_head_notfound_callback_param notfound_param;
-    if(support_compat_dir){
-        curlmulti.SetNotFoundCallback(multi_head_notfound_callback);
-        curlmulti.SetNotFoundCallbackParam(reinterpret_cast<void*>(&notfound_param));
-    }
+    // common variables
+    Semaphore    multi_head_sem(0);
+    int          req_count  = 0;
+    int          req_result = 0;
```
gaul, line 3104: I think this is probably better expressed as a `std::future` but we can address this in a separate commit.

gaul (approved):

src/s3fs.cpp
ggtakec: Thanks.
I'll fix it after I merge the PRs.

PR #2605, opened 2024-11-10
Refactored parallel get object request

ggtakec:
#2600

### Details
This is the split PR Phase 5 (5/8) for #2600.

The Parallel Get Object process has been changed to no longer use `S3fsMultiCurl`.
Instead, it has been moved to a worker thread managed by `ThreadPoolMan`.

This PR will be rebased and removed from draft status once the previous PR(#2604) is merged into master.


ggtakec:
@gaul 
Maybe the build on macos13 will fail for a while, but this seems to be a Runner issue as per the comment below, so please ignore it.
https://github.com/s3fs-fuse/s3fs-fuse/issues/2552#issuecomment-2476041074

PR #2609, opened 2024-11-10
Organized multi-threading related options

ggtakec:
#2600

### Details
This is the split PR Phase 9 (Last) for #2600.

We've streamlined options related to the number of threads and parallel processing.
The `multireq_max`, `parallel_count`, and `parallel_upload` options have been removed and unified into the `max_thread_count` option.

This PR will be rebased and removed from draft status once the previous PR(#2608) is merged into master.


src/s3fs.cpp
```diff
@@ -4941,23 +4935,17 @@ static int my_fuse_opt_proc(void* data, const char* arg, int key, struct fuse_ar
...
-            int maxpara = static_cast<int>(cvt_strtoofft(strchr(arg, '=') + sizeof(char), /*base=*/ 10));
-            if(0 >= maxpara){
-                S3FS_PRN_EXIT("argument should be over 1: parallel_count");
-                return -1;
-            }
-            S3fsCurl::SetMaxParallelCount(maxpara);
-            return 0;
+        else if(is_prefix(arg, "multireq_max=") || is_prefix(arg, "parallel_count=") || is_prefix(arg, "parallel_upload=")){
+            S3FS_PRN_EXIT("this option is Obsoleted, the max_thread_count option can be used instead.");
+            return -1;
```
gaul, line 4940: I think some users may have s3fs configured in fstab and similar with older options and this can break their setups.  Can we just `S3FS_PRN_WARN` and ignore obsolete options and put a TODO: to remove it some day in the future?

gaul (changes requested):

src/s3fs.cpp
ggtakec: I was also worried about compatibility.
I agree with you, so I changed the message to a warning and changed it so that the specified value is treated the same as `max_thread_count`.
I haven't decided how long I'll keep it like this for, but I'll keep it like this for a while.

PR #2611, opened 2024-11-14
Filipzitny/pla 3139 refactor mounters authentication flow

Yggdrasill501:


gaul:
Could you share what you are trying to do?  Why not upstream your fixes instead of running a fork?

Yggdrasill501:
I needed to sync the latest release with our fork and not lose the updates that ware made there before, but opening PR here was by accident, I wanted to open in our fork.

Issue #2617, opened 2024-12-10
The performance decreases heavily for concurrent IOs.


FangQianan:
#### Version of s3fs being used (`s3fs --version`)
 V1.95 (commit:561ce1e20600bab42133b5e51511483800d3ef76)

#### How to run s3fs, if applicable

All parameters are the default ones.

### Details about issue
I started 10 processes to write 10 different large files via s3fs concurrently. The performance decreases heavily as I raise up the concurrency. I tried to analyze the problem and here is what I found:
1. Thread 1: The flush api holds fent lock, and uploads files during holding this lock;
<img width="725" alt="image" src="https://github.com/user-attachments/assets/a07c2bb3-f299-4c4a-8991-e2a2f07dae2f">

2. Thread 2: Writing another file. In s3fs_write() -> AutoFdEntity::GetExistFdEntity() -> FdManager::GetExistFdEntity(), s3fs tried to find the other file's pseudo_fd by searching all entities. Yet inside FindPseudoFd() every opened file's fent lock is required.
<img width="668" alt="image" src="https://github.com/user-attachments/assets/7d16596c-1572-4a1b-ad75-b0e4f77eabff">

<img width="621" alt="image" src="https://github.com/user-attachments/assets/b324a395-f55f-48f7-b430-cc8ca672e1cb">

This means that once one file is being uploaded with its fent lock being held, there is a chance that other threads trying to read/write other files are blocked.

Actually, not only does this happen when a file is being uploaded, but also when a file is being downloaded in Read() with fent lock being held. 

Given the information above, can you guys provide a solution to fix it? Or is there a reason why you guys designed the lock like this? Thanks!

FangQianan:
<img width="857" alt="image" src="https://github.com/user-attachments/assets/3624f90c-47b4-4f5f-9e10-5626d43f99a5" />
 Here is what I modified. And the performance is twice of that of the the original code.

PR #2623, opened 2024-12-18
change the way to get existing fdentity to optimize concurrent IO performance

FangQianan:
https://github.com/s3fs-fuse/s3fs-fuse/issues/2617

### Details

Change the way to get existing fdentity to better conccurent IO's performance.
Instead of traversing the fent map and only compare the pseudo fd, which needs to hold every fdentity 's lock, I look for the fdentity with path at first and then verify the pseudo fd, which only needs to hold the right fdentity's lock.

I started 20 processes to read/write 500 files of 1MB in total concurrently. For ver 1.91, the RW performance is twice of that of the old code. And for ver 1.95, the performance is 20% better after this fix.

P.S. v1.95 performs worse than v1.91 (just half) according to my test (without this fix). This is another topic though. I will try to figure out the reason too and maybe you can also have a look into it.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -503,27 +503,24 @@ FdEntity* FdManager::GetFdEntityHasLock(const char* path, int& existfd, bool new
 
     if(-1 != existfd){
         for(auto iter = fent.cbegin(); iter != fent.cend(); ++iter){
-            if(iter->second && iter->second->FindPseudoFd(existfd)){
+            if(iter->second && 
+              iter->second->GetROPath() == path &&
+              iter->second->FindPseudoFd(existfd)){
                 // found opened fd in map
-                if(iter->second->GetPath() == path){
-                    if(newfd){
-                        existfd = iter->second->Dup(existfd);
-                    }
-                    return iter->second.get();
+                if(newfd){
+                    existfd = iter->second->Dup(existfd);
                 }
-                // found fd, but it is used another file(file descriptor is recycled)
-                // so returns nullptr.
-                break;
+                return iter->second.get();
             }
         }
-    }
-
-    // If the cache directory is not specified, s3fs opens a temporary file
-    // when the file is opened.
-    if(!FdManager::IsCacheDir()){
-        for(auto iter = fent.cbegin(); iter != fent.cend(); ++iter){
-            if(iter->second && iter->second->IsOpen() && iter->second->GetPath() == path){
-                return iter->second.get();
+    } else {
+        // If the cache directory is not specified, s3fs opens a temporary file
+        // when the file is opened.
+        if(!FdManager::IsCacheDir()){
+            for(auto iter = fent.cbegin(); iter != fent.cend(); ++iter){
+                if(iter->second && iter->second->IsOpen() && iter->second->GetROPath() == path){
+                    return iter->second.get();
+                }
             }
         }
     }
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -641,10 +641,9 @@ FdEntity* FdManager::GetExistFdEntity(const char* path, int existfd)
       // no matter use_cache is enabled or not, search from all entities to 
       // find the entity with the same path. And then compare the pseudo fd.
       for(iter = fent.begin(); iter != fent.end(); ++iter) {
-        // path is protected by fd_manager_lock (RenamePath), therefore there 
-        // is no need to hold entity lock inside it (FindPseudoFd holds 
-        // lock inside).
-        if(iter->second && (iter->second->GetPathWithoutLock() == path)
+        // GetROPath() holds ro_path_lock rather than fdent_lock.
+        // Therefore GetExistFdEntity does not contends with FdEntity::Read() / Write().
+        if(iter->second && (iter->second->GetROPath() == path)
            && iter->second->FindPseudoFd(existfd)) {
           return iter->second.get();
         }
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -630,13 +630,28 @@ FdEntity* FdManager::GetExistFdEntity(const char* path, int existfd)
 
     UpdateEntityToTempPath();
 
-    // search from all entity.
-    for(auto iter = fent.cbegin(); iter != fent.cend(); ++iter){
-        if(iter->second && iter->second->FindPseudoFd(existfd)){
-            // found existfd in entity
-            return iter->second.get();
+    // If use_cache is disabled, or the disk space is insufficient when use_cache
+    // is enabled, the corresponding key of the entity in fent is not path. 
+    fdent_map_t::iterator iter = fent.find(std::string(path));
+    if(fent.end() != iter){
+      if(iter->second && iter->second->FindPseudoFd(existfd)){
+        return iter->second;
+      }
+    } else {
+      // no matter it's use_cache or not, we search from all entities to 
+      // find entity with the same path. Then we compare the pseudo fd.
+      for(iter = fent.cbegin(); iter != fent.cend(); ++iter){
+        // IsOpen() is protected by fd_manager_lock, and so is GetPath().
+        // Therefore there is no need to hold entity lock here (FindPseudoFd
+        // holds it inside).
+        if(iter->second && iter->second->IsOpen() && 
+            0 == strcmp(iter->second->GetPath(), path) &&
+            iter->second->FindPseudoFd(existfd)){
+          return iter->second.get();
         }
+      }
     }
+
     // not found entity
     return nullptr;
 }
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -632,21 +632,20 @@ FdEntity* FdManager::GetExistFdEntity(const char* path, int existfd)
 
     // If use_cache is disabled, or the disk space is insufficient when use_cache
     // is enabled, the corresponding key of the entity in fent is not path. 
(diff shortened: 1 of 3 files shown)


src/fdcache.cpp
```diff
@@ -630,13 +630,28 @@ FdEntity* FdManager::GetExistFdEntity(const char* path, int existfd)
...
-    for(auto iter = fent.cbegin(); iter != fent.cend(); ++iter){
-        if(iter->second && iter->second->FindPseudoFd(existfd)){
-            // found existfd in entity
-            return iter->second.get();
+    // If use_cache is disabled, or the disk space is insufficient when use_cache
+    // is enabled, the corresponding key of the entity in fent is not path. 
+    fdent_map_t::iterator iter = fent.find(std::string(path));
+    if(fent.end() != iter){
+      if(iter->second && iter->second->FindPseudoFd(existfd)){
+        return iter->second;
```
gaul, line 638: Should this be `iter->second.get()`?

src/fdcache.cpp
```diff
@@ -630,13 +630,28 @@ FdEntity* FdManager::GetExistFdEntity(const char* path, int existfd)
...
     UpdateEntityToTempPath();
 
-    // search from all entity.
-    for(auto iter = fent.cbegin(); iter != fent.cend(); ++iter){
-        if(iter->second && iter->second->FindPseudoFd(existfd)){
-            // found existfd in entity
-            return iter->second.get();
+    // If use_cache is disabled, or the disk space is insufficient when use_cache
+    // is enabled, the corresponding key of the entity in fent is not path. 
+    fdent_map_t::iterator iter = fent.find(std::string(path));
```
gaul, line 635: You can just use `auto` here.

src/fdcache.cpp
```diff
@@ -630,13 +630,28 @@ FdEntity* FdManager::GetExistFdEntity(const char* path, int existfd)
...
+      }
+    } else {
+      // no matter it's use_cache or not, we search from all entities to 
+      // find entity with the same path. Then we compare the pseudo fd.
+      for(iter = fent.cbegin(); iter != fent.cend(); ++iter){
+        // IsOpen() is protected by fd_manager_lock, and so is GetPath().
+        // Therefore there is no need to hold entity lock here (FindPseudoFd
+        // holds it inside).
+        if(iter->second && iter->second->IsOpen() && 
+            0 == strcmp(iter->second->GetPath(), path) &&
```
gaul, line 648: Should this be `iter->second->GetPath() == path`?

gaul (changes requested):
This does not compile -- how did you test this?

ggtakec (commented):
@FangQianan 
Thanks, although the build is failing (please fix what @gaul pointed out), I think the fix is ​​correct.

src/fdcache.cpp
FangQianan: Sure. Sorry that I got the version wrong... It turns out I was testing and compiling based on version 1.91. I'll fix the comments. Thank you!

src/fdcache.cpp
```diff
@@ -630,13 +630,27 @@ FdEntity* FdManager::GetExistFdEntity(const char* path, int existfd)
...
+        return iter->second.get();
+      }
+    } else {
+      // no matter use_cache is enabled or not, search from all entities to 
+      // find the entity with the same path. And then compare the pseudo fd.
+      for(iter = fent.begin(); iter != fent.end(); ++iter) {
+        // path is protected by fd_manager_lock (RenamePath), therefore there 
+        // is no need to hold entity lock inside it (FindPseudoFd holds 
+        // lock inside).
+        if(iter->second && (iter->second->GetPathWithoutLock() == path)
```
ggtakec, line 647: Make sure to call `GetPath()` instead of `GetPathWithoutLock()`.

src/fdcache_entity.h
```diff
@@ -154,6 +155,7 @@ class FdEntity : public std::enable_shared_from_this<FdEntity>
             const std::lock_guard<std::mutex> lock(fdent_lock);
             return path;
         }
+        std::string GetPathWithoutLock() const { return path; }
```
ggtakec, line 158: Since we don't want to expose the public method without locking, should remove this method.

ggtakec (changes requested):
In my previous comment, I said that the contents of the PR were fine, but I was wrong.
The `GetPathWithoutLock()` method has been added, but this method relies on logic outside the class for data protection.
Even if this method is fine now, it may cause problems in the future, so please use `GetPath()` with a lock.

Also, we have merged the enforcement of the `aws` command package used in testing into master, so please rebase the source code of this PR to master branch.
(If you don't rebase, all tests will be failed.)

src/fdcache.cpp
FangQianan: GetPath() holds the fdentity lock. If I call GetPath() here, this fix will make no sense. Because it is traversing every fdentity and trying to get their lock that make the cocurrent-IO performance this bad. GetPathWithoutLock here tries to get an fdentity's lock only if it has the target path.   
I can understand your concern that path might be changed without fd_manager_lock in the future. But I think the path for an FdEntity should a constant attribute unless RenamePath is called, which is protected by the fd_manager_lock.  I think this is also a basic rule.  
To avoid committing too many times, I'll mege master and recommit after you comment on this. Thank you!

FangQianan:
> In my previous comment, I said that the contents of the PR were fine, but I was wrong. The `GetPathWithoutLock()` method has been added, but this method relies on logic outside the class for data protection. Even if this method is fine now, it may cause problems in the future, so please use `GetPath()` with a lock.
> 
> Also, we have merged the enforcement of the `aws` command package used in testing into master, so please rebase the source code of this PR to master branch. (If you don't rebase, all tests will be failed.)

`GetPath()` holds the fdentity lock. If I call `GetPath()` here, this fix will make no sense. Because it is traversing every fdentity and trying to get their lock that make the cocurrent-IO performance this bad. `GetPathWithoutLock` here tries to get an fdentity's lock only if it has the target path.   
I can understand your concern that path might be changed without fd_manager_lock in the future. But I think the path for an FdEntity should a constant attribute unless `RenamePath` is called, which is protected by the fd_manager_lock.  I think this is also a basic rule.  
To avoid committing too many times, I'll mege master and recommit after you comment on this. Thank you!

ggtakec:
@FangQianan
I submitted PR #2637 to enable exclusive control of variables while minimizing the impact on performance.

Could you take a look at the code for #2637?
The added `FdEntity::GetROPath()` method should function as an alternative to `FdEntity::GetPathWithoutLock()`.
I believe this PR should enable exclusive control of variables while minimizing the impact of the lock.

@gaul
Please take a look at PR #2637, and review it.

FangQianan:
> #2637



> @FangQianan I submitted PR #2637 to enable exclusive control of variables while minimizing the impact on performance.
> 
> Could you take a look at the code for #2637? The added `FdEntity::GetROPath()` method should function as an alternative to `FdEntity::GetPathWithoutLock()`. I believe this PR should enable exclusive control of variables while minimizing the impact of the lock.
> 
> @gaul Please take a look at PR #2637, and review it.

FangQianan:
> @FangQianan I submitted PR #2637 to enable exclusive control of variables while minimizing the impact on performance.
> 
> Could you take a look at the code for #2637? The added `FdEntity::GetROPath()` method should function as an alternative to `FdEntity::GetPathWithoutLock()`. I believe this PR should enable exclusive control of variables while minimizing the impact of the lock.
> 
> @gaul Please take a look at PR #2637, and review it.

@ggtakec I've looked through PR #2637 . This 確かに is a solution to minimizing the impact of the FdEntity lock while ensuring multi-thread safety for path. I'll adopt GetROPath after #2637 is merged.

ggtakec:
@FangQianan Sorry for the delay.
I've merged #2637, could you rebase your fix?

FangQianan:
> @FangQianan Sorry for the delay. I've merged #2637, could you rebase your fix?

@ggtakec I've merged the latest code from the master branch. And I fixed a bug in #2637 (the `FdEntity::ro_path` is not initialized in the constructor). Please have a check.

FangQianan:
> > @FangQianan Sorry for the delay. I've merged #2637, could you rebase your fix?
> 
> @ggtakec I've merged the latest code from the master branch. And there are also some other changes:
> 
> 1. I fixed a bug in [Added backup variable for fdcache entity path #2637](https://github.com/s3fs-fuse/s3fs-fuse/pull/2637) (the `FdEntity::ro_path` is not initialized in the constructor).
> 2. I found that `FdEntity::GetFdEntityHasLock()` called by `AutoFdEntity::Attach()` (which is called by `s3fs_release()`) also had the same `FdEntity::fdent_lock` contention, and I changed here too. Testing results indicates that this change can improve the read performance (500 files sized of 1MB, concurrency of 20) for about 25% compared with my last commit.
> 
> Please have a check.

I believe the second change is safe. The path and exist_fd passed by`s3fs_release()` are valid, which is made sure by libfuse. Even if the path is changed while this file is opened, the path can keep consistency with the pseudo_fd, since the path is protected by `FdManager::fd_manager_lock` and `FdEntity::ro_path_lock()`.
As for the case exist_fd == -1 (called by `FdManager::HasOpenEntityFd()`), the logic is still the same.

ggtakec (approved):
I think the second fix you added is fine too.
Sorry for the late my replies and it took me a while, but I think your suggestion has made the code more secure.
Thanks for your kindness.

PR #2625, opened 2024-12-26
Expand clang-tidy CI target to all static-checks

gaul:
Relocate cppcheck and Shellcheck into a single CI target instead of running them as part of all distributions.  While this modestly reduces run-time by about 20 seconds, more importantly it avoids workarounds for older checker versions and simplifies the code.


gaul:
This should allow us to remove a lot of `cppcheck-suppress` comments.

ggtakec (approved):

ggtakec:
> This should allow us to remove a lot of `cppcheck-suppress` comments.

Thanks
Not only that, it will also save you time on your CI builds.:-)

PR #2631, opened 2025-01-17
Experiment with ubuntu-24.04-arm

gaul:
aarch64 GitHub runners should be faster than x86_64.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,14 @@ permissions:
 #
 jobs:
   Linux:
-    runs-on: ubuntu-latest
+    #
+    # [NOTE]
+    # The runner refuses to run JavaScript actions like actions/checkout in an
+    # Alpine container on anything but x86-64 because it only ships a musl
+    # node for that architecture, see actions/runner#801.  Run the Alpine
+    # container on x86-64 and everything else on the faster aarch64 runners.
+    #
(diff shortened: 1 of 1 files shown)


gaul:
This fails with an unexpected error:

```
2025-01-17T00:58:18.4719830Z upload failed: - to s3://s3fs-integration-test/testrun-22766/directory/test-s3fs-30462.txt An error occurred (XAmzContentSHA256Mismatch) when calling the PutObject operation: The provided 'x-amz-content-sha256' header does not match what was computed.
```

Maybe related to an AWS CLI upgrade?

gaul:
AWS CLI 2.23.0 release notes:

* feature:``s3``: The S3 client attempts to validate response checksums for all S3 API operations that support checksums. However, if the SDK has not implemented the specified checksum algorithm then this validation is skipped. Checksum validation behavior can be configured using the ``when_supported`` and ``when_required`` options - in the shared AWS config file using ``response_checksum_validation``, and as an env variable using ``AWS_RESPONSE_CHECKSUM_VALIDATION``.

gaul:
> Error: JavaScript Actions in Alpine containers are only supported on x64 Linux runners. Detected Linux Arm64

gaul:
This does not show a big difference in average run times: 9-10 minutes for ARM runners compared to 10-12 for x86-64 runners.  But Valgrind improved to 20 minutes on ARM compared to 24 minutes for x86-64.

gaul:
Upstream issue alpine: actions/runner#801.  Maybe just easier to disable this in the short-term?

PR #2633, opened 2025-01-17
Pin AWS CLI to work around S3Proxy limitation

gaul:
Also specify architecture to allow future ARM64 CI to work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,7 +139,8 @@ jobs:
       - name: Install awscli2
         run: |
           cd /tmp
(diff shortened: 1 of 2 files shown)


ggtakec (approved):
Thanks
I have confirmed that specifying this version resolves the following error.
```
upload failed: - to s3://s3fs-integration-test/... An error occurred (XAmzContentSHA256Mismatch) when calling the PutObject operation: The provided 'x-amz-content-sha256' header does not match what was computed.
```

Could this be reverted by the s3proxy handles (https://github.com/gaul/s3proxy/issues/760)?

PR #2637, opened 2025-01-21
Added backup variable for fdcache entity path

ggtakec:
#2623

### Details
This PR adds variables and methods for #2623.

To solve the problem of point out by #2623, the `FdEntity::path` variable needs to be available without locking for performance.
The `FdEntity::path` variable is protected by a lock(`FdEntity::fdent_lock`), but this lock is wide-ranging.
#2623 code should avoid this lock, but I propose this PR method instead of not locking.

I prepared a copy(backup for read only) variable of the `FdEntity::path` variable and prepared a lock(`FdEntity::ro_path_lock`) for only the `FdEntity::ro_path` variable.
This lock variable provides exclusive control only when the `FdEntity::path` variable is updated, so it provides the same functionality as not locking `FdEntity::fdent_lock`.
And I think the variable can be safely held.

@gaul
This process seems wasteful, but I think it is the best way to deal with #2623 at the moment.
Please give me your opinion.

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -163,6 +163,10 @@ void FdEntity::Clear()
     pagelist.Init(0, false, false);
     path      = "";
     cachepath = "";
+
+    // set backup(read only) variable
+    const std::lock_guard<std::mutex> ro_lock(ro_path_lock);
+    ro_path   = path;
 }
 
 // [NOTE]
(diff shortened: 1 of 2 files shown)


ggtakec:
There are multiple calls to `FdEntity::GetPath()` in the FdManager class.
I haven't estimated the performance impact of these calls, but if this PR gets merged, we can change them to `FdEntity::GetROPath()`.

ggtakec:
@gaul This PR is blocking #2623, so I would like to move forward with it. If you have any issues, please comment.

PR #2642, opened 2025-02-08
Remove unneeded uses of `std::map::operator[]`

gaul:
These unintentionally mutate the map.  Script suggested by @danmar.

### DIFF
--- a/src/s3fs_cred.cpp
+++ b/src/s3fs_cred.cpp
@@ -514,19 +514,32 @@ bool S3fsCred::SetIAMCredentials(const char* response)
         return false;
     }
 
-    AWSAccessToken = keyval[IAM_token_field];
+    auto aws_access_token = keyval.find(IAM_token_field);
+    if(aws_access_token == keyval.end()){
+        return false;
+    }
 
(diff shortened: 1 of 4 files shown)


src/s3fs_cred.cpp
```diff
@@ -514,19 +514,32 @@ bool S3fsCred::SetIAMCredentials(const char* response)
...
+        if(access_key_id == keyval.end() || secret_access_key == keyval.end() || access_token_expire == keyval.end()){
+            return false;
+        }
+
+        AWSAccessKeyId = access_key_id->second;
+        AWSSecretAccessKey = secret_access_key->second;
+        AWSAccessTokenExpire = cvtIAMExpireStringToTime(access_token_expire->second.c_str());
     }
+
+    AWSAccessToken = aws_access_token->second;
```
gaul, line 542: I reordered this since it publishes all the values after ensuring all the expected ones are found.

ggtakec: Thanks. Checking for essential keys will make it stronger.

ggtakec (approved):

PR #2652, opened 2025-04-01
Ask for the provider on support requests

juliogonzalez:
None

### Details
Since more and more users are using s3fs-fuse with backends other than AWS, it's a good idea to ask them what they are using, as it's not always clear.

We can't provide a whole list of providers (even if we could, tomorrow there could be new one), so I picked some of those I am aware of. Happy to change the list of examples to something else :-)


ggtakec (approved):
Thank you!
This is one of the information we need to help.

PR #2656, opened 2025-04-03
Upgrade to S3Proxy 2.6.0

gaul:
Release notes: https://github.com/gaul/s3proxy/releases/tag/s3proxy-2.6.0


ggtakec:
@gaul The test is failing because the values ​​of `a/c/mtime` of parent directories like `test_update_parent_directory_time` and `test_time_mountpoint` seem to be different from the expected values.

gaul:
I looked at this more closely and it fails because:

```
[s3proxy] D 08-21 10:45:30.664 S3Proxy-Jetty-44 o.gaul.s3proxy.S3ProxyHandler:305 |::] request: Request(PUT https://127.0.0.1:8080/s3fs-integration-test//)@337e99d2
...
[s3proxy] D 08-21 10:45:30.667 S3Proxy-Jetty-44 o.g.s.S3ProxyHandlerJetty:127 |::] IllegalArgumentException:
java.lang.IllegalArgumentException: Invalid key name: path traversal attempt detected: /
        at org.gaul.s3proxy.nio2blob.AbstractNio2BlobStore.checkValidPath(AbstractNio2BlobStore.java:1143)
        at org.gaul.s3proxy.nio2blob.AbstractNio2BlobStore.putBlob(AbstractNio2BlobStore.java:543)
```

This is due to the use of `//` as a path.  Real S3 supports this although S3Proxy is attempting to prevent a parent directory traversal.  This is a bit tricky since these are not valid filesystem paths but we can fake it somehow.

ggtakec (approved):

PR #2662, opened 2025-04-19
Reduce use of awk in tests

gaul:


ggtakec (approved):
This has reduced unnecessary calls. Thank you.

PR #2663, opened 2025-04-19
Use latest JDK for Rocky

gaul:
This matches Fedora


gaul:
Maybe these are only part of EPEL: https://dl.fedoraproject.org/pub/epel/9/Everything/x86_64/Packages/j/

PR #2673, opened 2025-05-31
Fixed a logic in the multi head request retry count check

ggtakec:
n/a

### Details
Found an issue with checking the retry limit for multi head requests and multi put head requests.

When s3fs executes commands(such as ls) on a directory that contains multiple files and directories, head requests are processed in parallel as the multi head request.

Currently, the retry limit is set for all multiple head requests.
This limit prevents new head requests(after problem request) from being sent if the retry limit is reached while processing multiple head requests.

The problem is that if an object that cannot be read exists in the directory where user is trying to execute `ls` and the retry limit is reached, user will not be able to read some files or directories after reaching limit.

For example, if user execute `ls` on a directory with 100 files, if there is a file that has a reason to reach the retry limit, the result of the `ls` may be 80(example) files instead of 99.

To avoid this issue, in the case of a head request, it is necessary to allow one head request to be executed even if the retry limit is reached.
This can be avoided by deleting the first retry upper limit check in the `multi_head_req_threadworker()` and `multipart_put_head_req_threadworker()` functions.
(Previously, when we added the retry upper limit check for multi requests, but we did not take into consideration head requests like this case.)

A similar retry upper limit check is also performed in `parallel_get_object_req_threadworker()`, but since this is parallel execution of the upload process for one file, there is no problem with the current code.

### DIFF
--- a/src/s3fs_threadreqs.cpp
+++ b/src/s3fs_threadreqs.cpp
@@ -61,18 +61,6 @@ void* multi_head_req_threadworker(S3fsCurl& s3fscurl, void* arg)
         return reinterpret_cast<void*>(-EIO);
     }
 
-    // Check retry max count and print debug message
-    {
-        const std::lock_guard<std::mutex> lock(*(pthparam->pthparam_lock));
-
-        S3FS_PRN_INFO3("Multi Head Request [filler=%p][thparam_lock=%p][retrycount=%d][notfound_list=%p][wtf8=%s][path=%s]", pthparam->psyncfiller, pthparam->pthparam_lock, *(pthparam->pretrycount), pthparam->pnotfound_list, pthparam->use_wtf8 ? "true" : "false", pthparam->path.c_str());
-
-        if(S3fsCurl::GetRetries() < *(pthparam->pretrycount)){
-            S3FS_PRN_ERR("Head request(%s) reached the maximum number of retry count(%d).", pthparam->path.c_str(), *(pthparam->pretrycount));
-            return reinterpret_cast<void*>(-EIO);
-        }
-    }
-
     s3fscurl.SetUseAhbe(false);
 
     // loop for head request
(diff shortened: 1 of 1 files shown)


gaul:
> The problem is that if an object that cannot be read exists in the directory where user is trying to execute `ls` and the retry limit is reached, user will not be able to read some files or directories after reaching limit.

Could you share an example of a problem object?  Should we add a test for this?

But I am not sure about this approach.  What does `ls` do for a local file system if `readdir` succeeds buts one of the `stat` calls fails?  I guess in this case `getdents` is called instead so `ls` cannot see this error?  I think NFS should use READDIRPLUS so it also has one operation not N operations.

ggtakec:
> Could you share an example of a problem object? Should we add a test for this?

The error is occurred when you access an object encrypted with SSE-C without specifying any encryption-related settings(i.e. without SSE mode), and the object returns a 400 error.
(Reference: x-amz-server-side​-encryption​-customer-* header)

If there is such an object in the directory and you access this object(Head request) the maximum number of retries, further accesses for other objects will not be possible, which causes a problem.

I noticed this problem by chance because I was testing a bucket that had such an object.

It seems difficult to test this case.
You need to support SSE or other methods in s3proxy, or prepare another object that returns a 400 error(case of retry).

> READDIRPLUS 
It seems that READDIRPLUS is supported from libfuse 3.10 onwards.
(Until then, readdir will continue to be provided.)

Note that in my failed example, some objects were unable to receive the Head response due to a 400 error.
In this case, the object is not displayed as a result of the ls command.
This is because the object's Stat information (Head response) cannot be read, so the object's existence is unknown and it cannot be displayed. (This is due to the file type, permissions, etc. being unknown.)

Even if READDIRPLUS is supported, s3fs-fuse will still send a Head request to get the stats for each object.

##### NOTE
The final goal of the current refactoring of StatCache is to cache the list_bucket information (not the direct data but the processed data).
This should be useful when we support READDIRPLUS.

ggtakec:
The fix for this PR was included in #2681 and merged, so it is now closed.

PR #2679, opened 2025-06-02
Upgrade CI to Alpine 3.22

gaul:


gaul:
This keeps failing due to:
> test-utils.sh: line 378: /usr/local/bin/aws: cannot execute: required file not found

Running `hash -r` doesn't seem to help.

ggtakec:
We need to use the packaged version of `aws-cli` rather than installing it manually.
(I don't know why this error occurs when installing it manually.)
It seems that you can run a test by reverting the parts you modified in `linux-ci-helper.sh`.
I wonder if this was deleted to make it the same as other OSes?
(There may not be official packages for other OSes, but ALPINE provides them as official packages, so we should use those.)

ggtakec:
#### Additional comments
You installed `aws-cli` manually and got a `required file not found` error, which seems to be because the dependency on glibc cannot be resolved.
As a test, I tried manually installing [alpine glibc](https://github.com/sgerrand/alpine-pkg-glibc), but it still couldn't resolve some symbols.
(Alpine Linux defaults to `musl libc`, so it seems that in many cases the glibc binary will get `"symbol not found"` depending on the version.)
In fact, it seems that the symbols are missing, as shown below.
```
$ ldd /usr/local/bin/aws
        /lib64/ld-linux-x86-64.so.2 (0x756c5153b000)
        libdl.so.2 => /lib64/ld-linux-x86-64.so.2 (0x756c5153b000)
        libz.so.1 => /usr/lib/libz.so.1 (0x756c51520000)
        libpthread.so.0 => /lib64/ld-linux-x86-64.so.2 (0x756c5153b000)
        libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x756c5153b000)
Error relocating /usr/local/bin/aws: __strcat_chk: symbol not found
Error relocating /usr/local/bin/aws: __snprintf_chk: symbol not found
Error relocating /usr/local/bin/aws: __vfprintf_chk: symbol not found
Error relocating /usr/local/bin/aws: __realpath_chk: symbol not found
Error relocating /usr/local/bin/aws: __strdup: symbol not found
Error relocating /usr/local/bin/aws: __memcpy_chk: symbol not found
Error relocating /usr/local/bin/aws: __vsnprintf_chk: symbol not found
Error relocating /usr/local/bin/aws: __strcpy_chk: symbol not found
Error relocating /usr/local/bin/aws: __fread_chk: symbol not found
Error relocating /usr/local/bin/aws: __fprintf_chk: symbol not found
```
I think it's better to use `aws-cli` in the official ALPINE package, as before.

gaul:
S3Proxy 2.6.0 lacks https://github.com/gaul/s3proxy/issues/760 so I don't believe I can do this right now.  Let me see if I can fix the actual S3Proxy bug.

gaul:
#2686 will unblock this PR.

PR #2689, opened 2025-06-16
Use curl instead of AWS CLI

gaul:
The latter starts up significantly slower which impedes integration test times.  curl has some limitations, e.g., no SSE support.


gaul:
curl is too old on Rocky 8 and Ubuntu 22.04 which lack `--aws-auth`.  curl is less old on Rocky 9 but still needs a workaround for `amz-content-sha256`.

gaul:
This appears to reduce run-time for fedora:42 from 14-15 to 10-11 minutes.

gaul:
For some reason Rocky Linux skips the integration tests with the newer curl?

gaul:
This is ready for review but seems like it triggers #2709 more often.

gaul:
Let's merge #2711 before this due to conflicts.

ggtakec (approved):

PR #2690, opened 2025-06-23
Create random string with date without /dev/urandom

ggtakec:
#2687

### Details
In the function for adding suffixes to test file names, I changed it so that it does not use `/dev/urandom`, but uses a random string using seconds and microseconds from the `date` command.

#### Why was this necessary?
I did not understand the root cause of this.
However, it seems that the phenomenon is related to `/dev/urandom` used in the `make_random_string` test utility function for MacOS.

In order to proceed with the problem phenomenon #2687, I was investigating the test results of `test_update_metadata_external_small(large)_object`.
And I found an abnormal slowdown in processing speed that sometimes occurred on MacOS.
This was caused by processing being blocked in the following part.
```
TEST_FILE_EXT=$(make_random_string)
```
I tried many things such as changing the way the `make_random_string` function is called, but as long as `/dev/urandom` is used, this problem continues to occur.
In particular, when running Github Actions (CI) with `dbglevel=dbg`, the blocking is noticeable.
_The block mentioned here means that the s3fs-fuse test script is blocked from running._

While this is blocked, s3fs-fuse periodically references the `/` and `/test-XXXX` directories.
_(For example, `s3fs_getattrs` called by `macos-fuse-t`.)_

#### About the solution
The fundamental solution to this problem is still unknown.
This is a very strange and difficult problem to understand.
But we need to avoid blocking work that was not essential to our tests, so I change it to not access `/dev/urandom` as a simple solution.

If possible, it would be a good idea to try only the above test cases at dbg level on MacOS.
I would also like some advice on this matter.

##### Note
This problem is blocking #2687.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -463,7 +463,7 @@ function test_update_metadata_external_small_object() {
     # [NOTE]
     # Use the only filename in the test to avoid being affected by noobjcache.
     #
-    local TEST_FILE_EXT; TEST_FILE_EXT=$(make_random_string)
+    local TEST_FILE_EXT; TEST_FILE_EXT=$(make_dec_string_9)
     local TEST_CHMOD_FILE="${TEST_TEXT_FILE}_chmod.${TEST_FILE_EXT}"
     local TEST_CHOWN_FILE="${TEST_TEXT_FILE}_chown.${TEST_FILE_EXT}"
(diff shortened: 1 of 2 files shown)


ggtakec:
#### Additional information
The symptoms of this problem are that access to `/dev/urandom` is simply blocked.
The s3fs-fuse calls during that block appear to just be accesses by the OS periodically (at intervals of less than a minute) checking directories and mount points.

ggtakec:
#2691 is a better fix so I'm closing this PR.

PR #2691, opened 2025-06-23
Simplify temporary file creation via `mktemp`

gaul:
Previously this used `/dev/urandom` does not guarantee uniqueness and sometimes blocked on macOS.  References #2690.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -463,7 +463,7 @@ function test_update_metadata_external_small_object() {
     # [NOTE]
     # Use the only filename in the test to avoid being affected by noobjcache.
     #
-    local TEST_FILE_EXT; TEST_FILE_EXT=$(make_random_string)
+    local TEST_FILE_EXT; TEST_FILE_EXT=$(mktemp "XXXXXXXXXX")
     local TEST_CHMOD_FILE="${TEST_TEXT_FILE}_chmod.${TEST_FILE_EXT}"
     local TEST_CHOWN_FILE="${TEST_TEXT_FILE}_chown.${TEST_FILE_EXT}"
     local TEST_UTIMENS_FILE="${TEST_TEXT_FILE}_utimens.${TEST_FILE_EXT}"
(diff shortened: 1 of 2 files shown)


ggtakec:
Thank you, this is a better fix than #2690.
I will close #2690 and move forward with this one.
Can you please provide additional fixes, referring to the comments?

gaul:
I am confused by these errors and it may take me more time to understand.  To me it seems like the file gets corrupted with all NUL bytes.  At least part of the symptom seems to be timing related since using the former fails more often than the latter:
```
-    echo "${TEST_INPUT}" | aws_cli s3 cp - "s3://${TEST_BUCKET_1}/${OBJECT_NAME}"
+    aws_cli s3api put-object --bucket "${TEST_BUCKET_1}" --key "${OBJECT_NAME}" --body "${TEST_INPUT_FILE}"
```
(`aws s3 cp -` seems to issue more operations)

ggtakec:
@gaul 
I believe the reason this test is failing has to do with what is my review comment.

The original code calls `make_random_string` to determine the suffix of each test file(`chmod`/`chown`/`utimens`...).
The reason this suffix is ​​unique is because AWS CLI creates files that s3fs-fuse is not aware of.

The purpose of this `test_update_metadata_external_small(large)_object` test is to test each command when there is no intermediate path(directory) object, there is a file object under it, and it is not in the s3fs cache(stat).

The `make_random_string` function simply creates a random string as a shell variable.
The same is true for #2690(has already been closed), but in #2690 it also tried not to touch the `/dev/urandom` device. (I think the `date` command actually touches `IO`, but not the disk device.)

In this PR, `mktemp` is used, so **a file is created in the current directory**.
This file is created in a directory managed by s3fs used in this test.
And the file path is the **same** as the file actually operated in this test.

In other words, a file identical to the object uploaded with aws cli is created in the process to determine this name before uploading.
Thus, this file created via s3fs is registered in the stat cache of s3fs, and when each command is tested, it still remains in the stat cache.

As a test for this theory, I inserted `sleep 2`(stat cache timeout is 1 second) between aws cli and each command, and the test did not fail. (except for macos)
Therefore, I don't think the theory of the above assumption is wrong.

I think it would be better to change this part of the correction code as follows. (ex. in `test_update_metadata_external_large_object`)
```
(*)  local TEST_FILE_EXT; TEST_FILE_EXT=$(mktemp "XXXXXXXXXX")
(*)  rm -f "${TEST_FILE_EXT}" >/dev/null 2>&1
     local TEST_CHMOD_FILE="${TEST_TEXT_FILE}_chmod.${TEST_FILE_EXT}"
     local TEST_CHOWN_FILE="${TEST_TEXT_FILE}_chown.${TEST_FILE_EXT}"
     local TEST_UTIMENS_FILE="${TEST_TEXT_FILE}_utimens.${TEST_FILE_EXT}"
     local TEST_SETXATTR_FILE="${TEST_TEXT_FILE}_xattr.${TEST_FILE_EXT}"
     local TEST_RMXATTR_FILE="${TEST_TEXT_FILE}_xattr.${TEST_FILE_EXT}"
```
_The only changes from master are (*)._
I have confirmed that the tests pass with this code.
I hope this problem is resolved.

gaul:
I now explicitly use `mktemp -p /tmp` which uses the correct directly and also this makes sure that the filename is not reused.  But I still see a symptom with an empty file.

gaul:
I think I misunderstood what this test is trying to do.  I figured out that the local temp file is not actually important just creating a unique name.

ggtakec (approved):

PR #2699, opened 2025-07-29
Remove S3FS_MALLOC_TRIM

gaul:
42b74c9d2ed67da4cf1f6586d179b1d3c34fcb10 introduced this flag but it is not clear that this behavior is required with recent libxml2 versions.


ggtakec (approved):
This flag was used for debugging purposes to detect memory leaks at the time.
Now, we have different detection methods and more sophisticated tests, so this flag is obsolete.
So, I think it's okay to remove it. (I was wondering about this too, so thanks to you.)

PR #2704, opened 2025-08-20
Remove unneeded x-prefix in comparisons

gaul:
Found via shellcheck.  Reference:

https://www.shellcheck.net/wiki/SC2268


gaul:
Merged to unblock CI.

ggtakec:
@gaul Thank you.
We attached these shell scripts just for utility exmaple, but I'll try to come up with some kind of official plan to cache out the file cache. (However, it's not a high priority yet.)

Issue #2709, opened 2025-08-22
first line(cache file stat) is different: "49425:0" != "49425:26214400"


gaul:
CI shows this symptom sometimes and I reproduced it locally in Valgrind:

```
s3fs: [INF]       curl.cpp:MultipartUploadContentPartSetup(3646): [tpath=/testrun-15985/big-file-s3fs-32049.txt][start=0][size=10485760][part=1]
s3fs: [INF]       curl_util.cpp:url_to_host(268): url is http://127.0.0.1:8081
s3fs: [INF]       curl.cpp:RequestPerform(1985): HTTP response code 200
s3fs: [INF]       curl.cpp:CreateCurlHandle(1511): already has handle, so destroyed it or restored it to pool.
s3fs: [INF]       s3fs_threadreqs.cpp:multipart_upload_part_req_threadworker(332): Multipart Upload Part Worker [path=/testrun-15985/big-file-s3fs-32049.txt][upload_id=d8018f10-3689-4929-b912-214ed80d880f][upload_fd=13][start=10485760][size=10485760][is_copy=false][part_num=2]
s3fs: [INF]       curl.cpp:MultipartUploadPartRequest(3846): Multipart Upload Part [tpath=/testrun-15985/big-file-s3fs-32049.txt][upload_fd=13][start=10485760][size=10485760][part_num=2][upload_id=d8018f10-3689-4929-b912-214ed80d880f][is_copy=false]
s3fs: [INF]       curl.cpp:MultipartUploadPartSetup(3400): Upload Part [path=/testrun-15985/big-file-s3fs-32049.txt][start=10485760][size=10485760][part=2]
s3fs: [INF]       curl.cpp:MultipartUploadContentPartSetup(3646): [tpath=/testrun-15985/big-file-s3fs-32049.txt][start=10485760][size=10485760][part=2]
s3fs: [INF]       curl_util.cpp:url_to_host(268): url is http://127.0.0.1:8081
s3fs: [INF] s3fs.cpp:s3fs_flush(3060): [path=/testrun-15985/big-file-s3fs-32049.txt][pseudo_fd=2][pid=1657532,uid=1000,gid=1000]
s3fs: [INF]       fdcache_entity.cpp:RowFlushHasLock(1371): [tpath=][path=/testrun-15985/big-file-s3fs-32049.txt][pseudo_fd=2][physical_fd=19]
[...]

PR #2710, opened 2025-08-24
Fixed CacheFileStat to force a sync when a file is changed

ggtakec:
#2709

### Details
Stats information for FileCache files is now always synced when updated.
For the cause and solution, see #2709.

### DIFF
--- a/src/fdcache_page.cpp
+++ b/src/fdcache_page.cpp
@@ -845,7 +845,10 @@ bool PageList::Serialize(CacheFileStat& file, ino_t inode)
         S3FS_PRN_ERR("failed to write stats(%d)", errno);
         return false;
     }
-
+    if(0 != fsync(file.GetFd())){
+        S3FS_PRN_ERR("failed to sync stats(%d), but continue...", errno);
+        return false;
+    }
     return true;
 }
 


src/fdcache_page.cpp
```diff
@@ -845,7 +845,10 @@ bool PageList::Serialize(CacheFileStat& file, ino_t inode)
         S3FS_PRN_ERR("failed to write stats(%d)", errno);
         return false;
     }
-
+    if(0 != fsync(file.GetFd())){
```
gaul, line 848: I don't understand how this helps.  `fsync` should take the data from the kernel buffers and write it to disk.  Without `fsync` another process should still be able to write from the kernel buffers.  If instead the data were in the user buffers, e.g., `fwrite`, and the code did not call `fflush`, then another process could not read the data.  But since this calls `pwrite` it should not be a problem.

ggtakec: You're right, fsync doesn't make sense.
I'll continue investigating other causes, and close this PR eventually (without merging).

ggtakec:
I close this PR, and #2714 posted instead f this.

PR #2714, opened 2025-08-26
Changed to serialize CacheFileStat after flushing a file

ggtakec:
#2709 (#2710)

### Details
The CacheFileStatus file(cache file stat information file) used to be updated when the cache file was closed, but we've changed it so that it also updates when the file is flushed.

The cause of bug #2709 appears to be as follows, so this can be worked around.

First, when updating a file, closing it, and reading the file's stat information, s3fs expects flush->relase to be called from FUSE, followed by getattr.

However, as previously(old) discussed, FUSE starts the next process without waiting for release to complete.
Presumably, this is why, in the case of bug #2709, the commands were called in the order flush->getattr->relase.

As a result, we've changed the CacheFileStatus file update, which previously only occurred on release, to now also occur on flush.

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -1419,6 +1419,24 @@ int FdEntity::RowFlushHasLock(int fd, const char* tpath, bool force_sync)
         FdManager::DeleteCacheFile(tpath);
     }
 
+    // [NOTE]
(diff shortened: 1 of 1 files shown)


gaul:
This makes sense!

PR #2717, opened 2025-08-27
Only install binaries in static-checks task

gaul:
Also remove outdated version checks.


.github/workflows/linux-ci-helper.sh
```diff
@@ -369,6 +362,7 @@ elif [ "${CONTAINER_FULLNAME}" = "opensuse/leap:15" ]; then
         openssl
         openssl-devel
         procps
+        python3
```
gaul, line 365: `clang-tools` on OpenSuSe includes some Python scripts and thus previously implicitly required `python3`.

PR #2718, opened 2025-08-27
Deduplicate CentOS 9 and 10

gaul:
EPEL is no longer required.


ggtakec:
@gaul Could you resolve the conflict?

PR #2720, opened 2025-08-28
Fixed test_update_time_chown in test

ggtakec:
#2689

### Details
I missed this in the review of #2689.
This header parameter misconfiguration often causes `test_update_time_chown` to fail. (m/c/atime were not set when uploading the test file.)


ggtakec:
@gaul I've merged this because I want to move forward with other PRs.

PR #2721, opened 2025-08-28
Changed to use rename when serializing to FileCacheStat

ggtakec:
#2709

### Details
Changed `PageList::Serialize()` to use `flock` to exclusively handle `ftruncate`->`pwrite` operations.
This fixes a bug reported in https://github.com/s3fs-fuse/s3fs-fuse/issues/2709#issuecomment-3226548346.

### DIFF
--- a/src/fdcache_page.cpp
+++ b/src/fdcache_page.cpp
@@ -820,32 +820,21 @@ bool PageList::ClearAllModified()
     return Compress();
 }
 
-bool PageList::Serialize(CacheFileStat& file, ino_t inode)
+bool PageList::Serialize(const CacheFileStat& file, ino_t inode) const
 {
-    if(!file.Open()){
-        return false;
-    }
-
-    //
-    // put to file
-    //
+    // make contents
     std::ostringstream ssall;
     ssall << inode << ":" << Size();
 
     for(auto iter = pages.cbegin(); iter != pages.cend(); ++iter){
         ssall << "\n" << iter->offset << ":" << iter->bytes << ":" << (iter->loaded ? "1" : "0") << ":" << (iter->modified ? "1" : "0");
     }
-
-    if(-1 == ftruncate(file.GetFd(), 0)){
-        S3FS_PRN_ERR("failed to truncate file(to 0) for stats(%d)", errno);
-        return false;
-    }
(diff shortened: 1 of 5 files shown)


src/fdcache_page.cpp
```diff
@@ -835,17 +836,26 @@ bool PageList::Serialize(CacheFileStat& file, ino_t inode)
     for(auto iter = pages.cbegin(); iter != pages.cend(); ++iter){
         ssall << "\n" << iter->offset << ":" << iter->bytes << ":" << (iter->loaded ? "1" : "0") << ":" << (iter->modified ? "1" : "0");
     }
+    std::string strall = ssall.str();
 
+    if(0 != flock(file.GetFd(), LOCK_EX)){
```
gaul, line 841: Let's not do file locking.  First of all this is advisory and it would require the shell script to use `flock(1)`.  But second this is slower.  Instead could s3fs write to a temporary file and rename on top of the old one?

ggtakec: Changed codes for using rename function instead of flock.

src/fdcache_stat.cpp
```diff
@@ -193,6 +193,47 @@ bool CacheFileStat::SetPath(const char* tpath, bool is_open)
...
+    if(0 != CacheFileStat::MakeCacheFileStatPath(path.c_str(), sfile_path, true)){
+        S3FS_PRN_ERR("failed to create cache stat file path(%s)", path.c_str());
+        return false;
+    }
+    std::string strTmpFile = mydirname(sfile_path) + "/.tmpstat.XXXXXX";
+    char *pTempFileName    = strdup(strTmpFile.c_str());
+
+    // open temporary file
+    int tmpfd;
+    if(-1 == (tmpfd = mkstemp(pTempFileName))){     // mode: 0600
```
gaul, line 212: Can we simplify this by calling `mkstemp(strTmpFile.data())` (actually this requires two `std::string`)?  This will requires to C++17 but we plan to do that anyway.  My objection is using manual memory management like `free`.

ggtakec: No problem. I fixed it by using `&str[]` to make it compatible with C++14.

gaul (approved):

PR #2723, opened 2025-08-29
Check integrity of downloaded binaries

gaul:
Usually package managers perform this check but we need to do it ourselves for custom binaries.


gaul:
Let's merge #2656 first.

ggtakec (approved):

PR #2724, opened 2025-08-30
Fixed test_external_modification test for MacOS

ggtakec:
n/a

### Details
In a recent CI(GHA) run, the `test_external_modification` test on macOS often failed.
I investigated the cause and have now found it, so I posted this PR.

The cause was a difference in the way `getattr` calls are made from FUSE on macOS and other Linuxs.
In the `test_external_modification` test, after uploading a file, it waits one second(while the file contents are modified by external process) and then reads it.
On other Linuxs, there is no `getattr` call after uploading, but on macOS, the `getattr` call is made.

The `stat_cache_interval_expire=1` option was set for s3fs when the test was run, so if the cache is accessed via a `getattr` call within the cache expiration period, the expiration period is extended.
For this reason, the `getattr` call is made on macOS, and waiting 1 second will still be within the cache expiration period.

To avoid this, added an additional 1-second sleep on macOS.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -403,7 +403,14 @@ function test_external_modification {
     # cache will be read out.
     # Therefore, we need to wait over 1 second here.
     #
+    # In particular, on MacOS, getattrs may be called after a file is
+    # uploaded(released).
+    # This extends the expiration date of the target file(1 sec by
+    # stat_cache_interval_expire option).
+    # Therefore, on MacOS, you need to add an additional 1 sec.
+    #
     sleep 1
+    wait_ostype 1 "Darwin"
 
     local OBJECT_NAME; OBJECT_NAME=$(basename "${PWD}")/"${TEST_TEXT_FILE}"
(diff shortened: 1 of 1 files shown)


ggtakec:
For reference, the following is the time extracted from the log when `test_external_modification` was executed on macOS.

#### (1) 2025-08-30T02:50:15.8581580Z
This is the time registered in the stat cache when the file upload was completed.
After this, `sleep 1` is executed.

#### (2) 2025-08-30T02:50:17.0314140Z
After the above, several `getattr` calls are made, and the time when the cache expiration date was extended by the last `getattr` is the time.
This date updated the registration time in the stat cache for the target file.

#### (3) 2025-08-30T02:50:17.0335750Z
And this is the time when the file was `read`.

On other Linux systems, the stat cache times out when read because there is no access(2), but on macOS it is still within the expiration period.

ggtakec:
@gaul If there are no issues with this PR, we need to fix this test, please let merge it first.
(This macOS CI test error occurs frequently, and it is occurring in checks of other PRs.)

PR #2729, opened 2025-09-07
The Truncate method of the StaCache class has been consolidated

ggtakec:
### Relevant Issue (if applicable)
#2728

### Details
Consolidated the Truncate method of the StaCache class.

PR #2737, opened 2025-10-09
Upgrade CI to Ubuntu 25.10

gaul:


gaul:
Merged due to triviality and I updated the GitHub settings.

ggtakec:
@gaul 
When running GHA on Ubuntu 25.10, a specific error appears to occur occasionally.
(This occured when running GHA on #2729 and #2738.)

This error appears to occur in the XML parser or a similar part. (Details have not yet been confirmed.)
```
 + stdbuf -oL -eL ../src/s3fs s3fs-integration-test s3fs-integration-test -o use_path_request_style -o url=https://127.0.0.1:8080 -o region=us-east-1 -o use_xattr=1 -o enable_unsigned_payload -o passwd_file=passwd-s3fs -o stat_cache_expire=1 -o stat_cache_interval_expire=1 -o dbglevel=info -o no_time_stamp_msg -o retries=3 -f -o use_cache=/tmp/s3fs-cache -o ensure_diskfree=10 -o fake_diskfree=200 -o use_xattr -o update_parent_dir_stat
 Trying: grep -q s3fs-integration-test /proc/mounts
 :1: parser error : Document is empty
 
 ^
 :1: parser error : Document is empty
```
The above `parser error` is output periodically, and I think an error is occurring in the XML parser processing of the ListBucket response.

The following test is failing, which I believe is due to this error message:
```
FAIL: test_update_parent_directory_time
FAIL: test_pjdfstest_mkdir
FAIL: test_pjdfstest_mkfifo
FAIL: test_pjdfstest_rename
FAIL: test_pjdfstest_rmdir
FAIL: test_pjdfstest_symlink
```
It seems that Ci does succeed in some cases, so it doesn't always fail.
So, I try to check this error.

gaul:
I've seen this with other CI runners too.  I suspect that either s3fs or S3Proxy is not handling some error codes properly.

ggtakec:
I think the `:1: parser error : Document is empty` error message has been fixed in #2739.
The CI test failure is caused by a different issue, so I'll continue to investigate.

ggtakec:
This test error issue has been moved to #2740 for further investigation.

PR #2757, opened 2025-11-24
Use JDK 25 on Ubuntu 25.10

gaul:


gaul:
Not-yet-released Alpine 3.23 supports JDK 25 and OpenSUSE 16 doesn't have a headless variant.

Issue #2760, opened 2025-11-30
Obsolete nonempty check from fuse2 left, while fuse3 doesn't support nonempty


pschichtel:
#### Version of s3fs being used (`s3fs --version`)

V1.96(commit:6235d0b +untracked files) with OpenSSL

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)
`libfuse3-4:amd64          3.17.2-3`

#### Provider (`AWS`, `OVH`, `Hetzner`, `iDrive E2`, ...)

Ceph RGW

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)

PRETTY_NAME="Debian GNU/Linux 13 (trixie)"
NAME="Debian GNU/Linux"
VERSION_ID="13"
VERSION="13 (trixie)"
VERSION_CODENAME=trixie
DEBIAN_VERSION_FULL=13.2
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

#### How to run s3fs, if applicable

[X] command line  
[] /etc/fstab  

```
mkdir -p /mnt/some-dir
touch /mnt/some-dir/test
s3fs -f some-bucket /mnt/some-dir -o ro,nonempty,listobjectsv2,instance_name=some-bucket,passwd_file=/tmp/tmp.YfZokQwAKK,use_path_request_style,url=https://some-host
```

or

```
mkdir -p /mnt/some-dir
touch /mnt/some-dir/test
s3fs -f some-bucket /mnt/some-dir -o ro,listobjectsv2,instance_name=some-bucket,passwd_file=/tmp/tmp.YfZokQwAKK,use_path_request_style,url=https://some-host
```

#### s3fs syslog messages (`grep s3fs /var/log/syslog`, `journalctl | grep s3fs`, or `s3fs outputs`)

```
fuse: unknown option(s): `-o nonempty'
```

or

```
s3fs: MOUNTPOINT directory /mnt/some-dir is not empty. if you are sure this is safe, can use the 'nonempty' mount option.
```

### Details about issue

[...]

PR #2761, opened 2025-11-30
Drop code related to the nonempty option

pschichtel:
The option was removed with fuse3 as it is the default behavior

#2760


gaul:
Superseded by #2765.

PR #2768, opened 2025-12-04
Micro-optimize copies, compares, and logging

gaul:
Logging should eagerly evaluate the log level to avoid unnecessary calls to parameters like calls to `STR_OBJTYPE`.  These contributed to slow Valgrind test run-times.


gaul:
This doesn't show much improvement in CI compared to local testing with callgrind, perhaps one minutes faster.  But these changes are good.

PR #2771, opened 2025-12-06
Simplify brew upgrade and install

gaul:


ggtakec:
This `brew` branch was created, so I deleted it after merging(as I thought it was unnecessary).
Please let me know if there are any problems.

Issue #2650, opened 2025-03-20
Cannot list folders when using SSE-C on Scaleway


lanormerh:
#### Version of s3fs being used (`s3fs --version`)
V1.95(commit:561ce1e

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)
fuse (2.9.9-6)

#### Kernel information (`uname -r`)
6.8.0-49-generic

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)
Debian GNU/Linux 12 (bookworm)

#### How to run s3fs, if applicable
s3fs -f -d -o curldbg -o allow_other,use_path_request_style,nocopyapi,sigv2 -o use_sse=custom:/etc/passwd-sse-c -o passwd_file=/etc/passwd-s3fs -o url=https://s3.fr-par.scw.cloud -o endpoint=fr-par bucket-name-redacted /mnt/s3fs

### Details about issue

I'm trying to run s3fs with SSE-C using Scaleway object storage.
As soon as I add an encrypted file, I can't list folders anymore (HEAD 400 Bad request)

```
2025-03-20T14:54:22.972Z [CURL DBG] < content-type: application/xml
2025-03-20T14:54:22.972Z [CURL DBG] < date: Thu, 20 Mar 2025 14:54:22 GMT
2025-03-20T14:54:22.972Z [CURL DBG] < x-amz-id-2: txgeaa2fbb4361c4a059d61-0067dc2c1e
2025-03-20T14:54:22.972Z [CURL DBG] < x-amz-request-id: txgeaa2fbb4361c4a059d61-0067dc2c1e
2025-03-20T14:54:22.972Z [CURL DBG] <
2025-03-20T14:54:22.972Z [CURL DBG] * Connection #2 to host s3.fr-par.scw.cloud left intact
2025-03-20T14:54:22.972Z [ERR] curl.cpp:RequestPerform(2629): HEAD HTTP response code 400, returning EPERM.
2025-03-20T14:54:22.972Z [WAN] curl_multi.cpp:MultiPerform(167): thread terminated with non-zero return code: -1
2025-03-20T14:54:22.972Z [ERR] s3fs.cpp:readdir_multi_head(3290): error occurred in multi request(errno=-1).
2025-03-20T14:54:22.972Z [ERR] s3fs.cpp:s3fs_readdir(3380): readdir_multi_head returns error(-1).
[...]

PR #2773, opened 2025-12-14
Fixed a bug in SSE-C md5 calculation using GnuTLS

ggtakec:
#2650 #2252

### Details
Fixed a bug in SSE-C md5 calculation when using GnuTLS.

#### [NOTE]
Currently, s3fs-fuse uses OpenSSL v3 on all operating systems, so this modified code can only be used if you build it yourself.
Note that Debian/Ubuntu packages, etc., before the transition to OpenSSL v3, use GnuTLS.
This modification does not affect OpenSSL v3 packages. (It only affects (modifies) those who build GnuTLS themselves.)

### DIFF
--- a/src/gnutls_auth.cpp
+++ b/src/gnutls_auth.cpp
@@ -233,7 +233,18 @@ bool s3fs_md5(const unsigned char* data, size_t datalen, md5_t* digest)
         S3FS_PRN_ERR("MD5 context creation failure: %s/%s", gcry_strsource(err), gcry_strerror(err));
         return false;
     }
-    gcry_md_write(ctx_md5, digest->data(), digest->size());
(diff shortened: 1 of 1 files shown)


gaul:
Thanks for fixing this!  I know this requires an unusual configuration but are any existing Debian or Ubuntu releases subject to the previously wrong code?

ggtakec:
I think this code has been affected since Release version 1.94(after merging #2252). 
But it is unlikely that it will be affected if packaging is done with OpenSSLv3.
And this only affects those who use SSE-C, and since the number of SSE-C users on S3(AWS) is likely to be decreasing, I think the impact will be small. (This is purely my personal speculation.)

Issue #2764, opened 2025-12-03
Chown and Chmod fails, except if nomultipart is set


gtz63:
#### Version of s3fs being used (`s3fs --version`)
V1.92

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)
Debian or recompiled version

#### Provider (`AWS`, `OVH`, `Hetzner`, `iDrive E2`, ...)
Openstack Swift

#### Kernel information (`uname -r`)
5.10.0-21-cloud-amd64

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)
Debian 11 / 12

#### How to run s3fs, if applicable
 sudo ./s3fs mybucket mybucket -f -o use_path_request_style,nosuid,rw,url=https://swift.private.cloud,allow_other,dbglevel=debug

#### s3fs syslog messages (`grep s3fs /var/log/syslog`, `journalctl | grep s3fs`, or `s3fs outputs`)
    s3fs.cpp:put_headers(906): [path=/afile.txt]
      curl.cpp:MultipartHeadRequest(4342): [tpath=/afile.txt]
      curl.cpp:PreMultipartPostRequest(3807): [tpath=/afile.txt]
curl_handlerpool.cpp:GetHandler(79): Get handler from pool: rest = 31
      curl_util.cpp:prepare_url(257): URL is https://swift.private.cloud/mybucket/afile.txt?uploads
      curl_util.cpp:prepare_url(290): URL changed is https://swift.private.cloud/mybucket/afile.txt?uploads
curl.cpp:RequestPerform(2484): connecting to URL https://swift.private.cloud/mybucket/afile.txt?uploads
      curl.cpp:insertV4Headers(2891): computing signature [POST] [/afile.txt] [uploads] []
      curl_util.cpp:url_to_host(334): url is https://swift.private.cloud
      curl.cpp:RequestPerform(2519): HTTP response code 200
curl_handlerpool.cpp:ReturnHandler(101): Return handler to pool
      curl.cpp:CopyMultipartPostSetup(4207): [from=/afile.txt][to=/afile.txt][part=1]
      curl_util.cpp:prepare_url(257): URL is https://swift.private.cloud/mybucket/afile.txt?partNumber=1&uploadId=2~YMOwdcgXW44b690iPVQ9gM1otvnJXy5
[...]

PR #2774, opened 2025-12-14
Added putheader_mp_threshold option

ggtakec:
##2764 

### Details
Currently, the value of multipart_threshold determines whether to upload a multipart object and its header (put_header).

Depending on the use case, there may be cases where performance is better if put_header is not executed in multipart.
For example, this is the case when using rsync as in #2764.
Executing put_header in multipart increases the number of request communications, which raises concerns about performance degradation.

To resolve this issue, we will now separate the thresholds for multipart upload and multipart header upload.
I have added `putheader_mp_threshold`, which allows user to set the threshold for executing this header upload in multipart.
If `putheader_mp_threshold` is not specified, the value of `multipart_threshold` will be used, as before.

@gaul
Although I have some concerns about adding more options, I have decided that it is necessary to provide optimal performance depending on the user (environment, execution content).
This PR is currently in Draft, but if there are no problems, I plan to change status to not set Draft.


ggtakec:
@gaul Could you please confirm the contents of this PR?
Since only users who need will set this new option, adding this option will not affect existing users.
_(Although there are concerns about increasing the number of options.)_

gaul:
I am trying to understand how to observe these changes but can't tell the difference in performance:

Default:
```
$ # default
$ dd bs=1MB count=100 if=/dev/zero of=gaulbackup/out
100+0 records in
100+0 records out
100000000 bytes (100 MB, 95 MiB) copied, 4.11383 s, 24.3 MB/s
$ time dd bs=1MB count=100 if=/dev/zero of=gaulbackup/out
100+0 records in
100+0 records out
100000000 bytes (100 MB, 95 MiB) copied, 4.06536 s, 24.6 MB/s
$ dd bs=1MB count=100 if=/dev/zero of=gaulbackup/out
100+0 records in
100+0 records out
100000000 bytes (100 MB, 95 MiB) copied, 4.3186 s, 23.2 MB/s
```
Specifying `-o putheader_mp_threshold=200`:
```
$ dd bs=1MB count=100 if=/dev/zero of=gaulbackup/out 
100+0 records in
100+0 records out
100000000 bytes (100 MB, 95 MiB) copied, 4.14951 s, 24.1 MB/s
$ dd bs=1MB count=100 if=/dev/zero of=gaulbackup/out 
100+0 records in
100+0 records out
100000000 bytes (100 MB, 95 MiB) copied, 4.12214 s, 24.3 MB/s
$ dd bs=1MB count=100 if=/dev/zero of=gaulbackup/out 
100+0 records in
100+0 records out
100000000 bytes (100 MB, 95 MiB) copied, 3.98463 s, 25.1 MB/s
```

Is there some more direct test I can run?  I am skeptical of adding more options that few people will specify.  Is there some default we can change that will benefit all users instead?  My suspicion is that `multipart_threshold` should be a much larger value than 25 MB, but on the other hand Goofys always uses multi-part uploads and uses parts as small as 5 MB for the first 500 parts.

ggtakec:
@gaul 
The reason of this issue is affected by the behavior of rsync. (#2764)
but it seems that more testing is needed.
The difference is large when the file content transfer and header update thresholds are used together, so I considered options.

As you say, I'm reluctant to add an option for cases that are unlikely to affect the majority of cases, but since this is an option that is not normally specified, it is unlikely to be used generally.

It may be possible to vary the part size depending on the target file size, as in Goofys, but the appropriate value is likely to vary depending on the user.
Then, I think these threshold vlues is difficult to decide. 
So this issue should be considered further.

gaul:
I doubt this will ever be used and it didn't resolve the symptoms in my testing but it doesn't hurt to have some configuration.

ggtakec:
@gaul Thanks

PR #2782, opened 2026-01-10
Added check for existence of uploaded test file by curl

ggtakec:
n/a

### Details
During CI testing, we encountered cases where files could not be accessed immediately after being uploaded.
The cause of the problem (the phenomenon caused by the cause) was identified from the log of the failed test.
The cause of this issue is unknown, but since it was not an issue with s3fs-fuse processing, temporary I have modified the code so that the tests can be run reliably.


ggtakec:
@gaul I'm sorry, this solution was insufficient, so I will close it and submit another PR.

PR #2788, opened 2026-01-23
Fix file descriptor leaks in error paths

CarstenGrohmann:
Close file descriptors before returning on error in three locations:
                                            
- `FdManager::MakeTempFile()`: fd leaked when `unlink()` failed after successful `mkstemp()`
- `FdEntity::Open()`: `physical_fd` leaked when `fstat()` failed after opening cache file
- `FdEntity::Open()`: `physical_fd` leaked when `OpenMirrorFile()` failed
                                          
These leaks would accumulate over time when error conditions occurred, eventually exhausting the process file descriptor limit.


ggtakec:
@CarstenGrohmann I agree with your fix.
Thank you for finding and fixing the problem.

PR #2789, opened 2026-01-23
Fix memory leaks in multipart upload error paths

CarstenGrohmann:
Delete thargs when `ThreadPoolMan::Instruct()` or `AwaitInstruct()` fails. The thread parameter struct was allocated with new but never freed when the thread pool rejected the work item, causing a memory leak on each failed multipart upload part request.


gaul:
Thank you for your contribution @CarstenGrohmann!  Ideally this would use a `std::unique_ptr` but the ownership model is more complicated due to `instruction_list`.

PR #2790, opened 2026-01-23
Use scope_guard for directory handles to prevent leaks

CarstenGrohmann:
This PR replaces manual `closedir()`  calls with the existing `scope_guard` RAII pattern in directory-traversing functions. This ensures directory handles are properly closed even when exceptions occur.

**Problem:**

The affected functions iterate over directories using `opendir()`/`readdir()`/`closedir()`. They contain:

1. `std::string` operations that can throw `std::bad_alloc`
2. Recursive calls that propagate exceptions from deeper levels

When an exception is thrown, execution bypasses the manual closedir() calls, causing directory handle leaks. In long-running processes with deep directory trees, this can eventually exhaust the file descriptor limit.

**Solution:**

Use the existing scope_guard class (defined in `s3fs_util.h`) to ensure `closedir()` is called when the scope exits, regardless of how it exits:

```
scope_guard guard([dp]() { closedir(dp); });
```

**Hint:**

The code in `s3fs.cpp` was correct and was only modified for consistency with the rest of the codebase.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -853,6 +853,7 @@ void FdManager::CleanupCacheDirInternal(const std::string &path)
         S3FS_PRN_ERR("could not open cache dir(%s) - errno(%d)", abs_path.c_str(), errno);
         return;
     }
+    scope_guard dir_guard([dp]() { closedir(dp); });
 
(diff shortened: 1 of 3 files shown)


ggtakec:
@CarstenGrohmann Thank you for addressing the leaks during exceptions.

Issue #2787, opened 2026-01-21
Sensitive Credentials Logged in Plaintext


CarstenGrohmann:
# Summary

Multiple locations in the codebase log sensitive credentials (IAM tokens, secret keys, SSE-C encryption keys) in plaintext. This exposes credentials in log files, syslog, and potentially centralized logging systems.

# Affected Code

## 1. IMDSv2 API Token logged on set (INFO3 level)

https://github.com/s3fs-fuse/s3fs-fuse/blob/4eee60ac2f45c33ab1ba58856f04e83c4e5a76a8/src/s3fs_cred.cpp#L410

## 2. Full IAM Credential Response logged (INFO3 level)

Contains AccessKeyId, SecretAccessKey, Token, and Expiration in plaintext:

https://github.com/s3fs-fuse/s3fs-fuse/blob/4eee60ac2f45c33ab1ba58856f04e83c4e5a76a8/src/s3fs_cred.cpp#L505-L507

## 3. IMDSv2 Token logged on error (ERR level - always output!)

https://github.com/s3fs-fuse/s3fs-fuse/blob/4eee60ac2f45c33ab1ba58856f04e83c4e5a76a8/src/s3fs_cred.cpp#L383

## 4. IAM Token logged in request function (INFO3 level)

https://github.com/s3fs-fuse/s3fs-fuse/blob/4eee60ac2f45c33ab1ba58856f04e83c4e5a76a8/src/s3fs_threadreqs.cpp#L1481

## 5. IAM Token AND IBM Secret Access Key logged (INFO3 level)

https://github.com/s3fs-fuse/s3fs-fuse/blob/4eee60ac2f45c33ab1ba58856f04e83c4e5a76a8/src/s3fs_threadreqs.cpp#L1497

## 6. SSE-C Encryption Key logged on error (ERR level - always output!)

https://github.com/s3fs-fuse/s3fs-fuse/blob/4eee60ac2f45c33ab1ba58856f04e83c4e5a76a8/src/curl.cpp#L651

# Risk Assessment

| Location | Log Macro | When Output |
|-----------|-----------|-------------|
| s3fs_cred.cpp:410 | `S3FS_PRN_INFO3` | `-o dbglevel=info3` or higher |
| s3fs_cred.cpp:507 | `S3FS_PRN_INFO3` | `-o dbglevel=info3` or higher |
| s3fs_cred.cpp:383 | `S3FS_PRN_ERR` | Always |
[...]

PR #2792, opened 2026-01-24
Mask sensitive authentication information to log

ggtakec:
#2787

### Details
As pointed out in #2787, this PR prevents authentication information (token and credential) from being output in plain text to logs.
Output related to authentication information has been changed so that it is output to the log by replacing it with `[SENSITIVE]`.

And the `insecure_logging` option has been added as a way to output this information as before for developer debugging purposes.
This option is provided as deprecated. (It should be avoided in production environments, but we understand that it may be used in development environments.)


ggtakec:
The source code has been updated.
- Fixed a bug that was found.
- Added masking to the message written by print_launch_message() at startup.
- Moved the added masking function to string_util.cpp(h) so that it can be unit tested in test_string_util.cpp.

Please review again.

PR #2794, opened 2026-01-26
Add error logging to scope_guard cleanup functions

CarstenGrohmann:
The scope_guard lambdas for close() and closedir() now check the return value and log an error with errno if the call fails. The error messages include the file or directory path to aid debugging.


gaul:

PR #2795, opened 2026-01-28
Bugfix：urlencode prefix when CheckBucket

fangqianan:
The prefix is not urlencoded in CheckBucket, so if I mount with a prefix with special characters inside it, such as "?prefix=%3D%3F", and I don't have the access right to the prefix, this function does not check the right prefix.


gaul:
Thank you for your contribution @fangqianan!

Issue #2798, opened 2026-02-02
Ambiguous/Missing error connecting to non-aws buckets


andrewdscott12:
#### Version of s3fs being used (`s3fs --version`)

V1.95 and up

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)

2.9.9
#### Provider (`AWS`, `OVH`, `Hetzner`, `iDrive E2`, ...)

IBM ICOS 

#### Kernel information (`uname -r`)

5.14.0-503.40.1.el9_5.s390x

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)

RHEL9.5

#### How to run s3fs, if applicable

[] command line  
[X] /etc/fstab  
rhui /var/lib/rhi fuse.s3fs _netdev,noatime,nonempty, allow_other,url=https://providerendpoint 0 0

```
```

#### s3fs syslog messages (`grep s3fs /var/log/syslog`, `journalctl | grep s3fs`, or `s3fs outputs`)

``` s3fs.cpp:s3fs_check_service(4503): Failed to check bucket and directory for mount point: Unable to connect(host=https://providerendpoint, message=)
```

### Details about issue

Ambiguous/missing errors. Checked against current version of s3fs_check_service function, and omission still appears to exist. s3fs-fuse does not provide a useful error when failing to connect due to a certificate validation issue. If Certificate validation errors from curl were passed through to the error generated by s3fs-fuse, user experience would be dramatically improved. Currently, failures due to cert validations are essentially silent. It simply won't mount the bucket and the message in the log is null.

andrewdscott12:
Don't know why log message not in post

when failure occurs, message in syslog:

s3fs.cpp:s3fs_check_service(4503): Failed toc heck bucket and directory for mount point : Unable to connect(host=https://providerendpoint, message=)

ggtakec:
@andrewdscott12 Could you please specify the `curldbg` option?
This should allow you to get the messages curl outputs.

andrewdscott12:
What prompted this issue was a support analyst beating their head against a wall for several days. In a corporate environment, such an analyst is often not allowed to make changes to the system, including adding debug switches to software invocations. They have to run them as-is and read the logs. In that kind of environment, A certificate validation error is a serious failure with a relatively simple remedy. If that could be surfaced without having to set a debug option, that would ideal.

ggtakec:
[...]

PR #2799, opened 2026-02-08
Improved message output when check bucket error occurs

ggtakec:
#2798

### Details
This PR is improved the system so that a curl error message is displayed when a curl error occurs during s3fs startup.
And revised the processing(debug message) after executing curl_easy_setopt, and added a missing error detection feature.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -43,6 +43,8 @@
 #include "addhead.h"
 #include "s3fs_xml.h"
 
+using namespace std::string_literals;
+
 //-------------------------------------------------------------------
 // Symbols
 //-------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


andrewdscott12:
Looks very useful. Thank you.

src/curl.cpp
```diff
@@ -1616,6 +1616,15 @@ bool S3fsCurl::GetResponseCode(long& responseCode, bool from_curl_handle) const
     return true;
 }
 
+bool S3fsCurl::GetCurlErrorString(std::string& strError) const
+{
+    if(CURLE_OK == curlCode){
+        return false;
+    }
+    strError = std::string("CURL ERROR(") + std::to_string(curlCode) + std::string("): ") + curl_easy_strerror(curlCode);
```
gaul: You can use string literals to improve readability:
```c++
using namespace std::string_literals;
...
strError = "CURL ERROR("s + std::to_string(curlCode) + "): "s + curl_easy_strerror(curlCode);
```

s3fs already uses this feature in `test_string_util.cpp`.

Issue #2772, opened 2025-12-07
1.97 release


gaul:
Mostly to fix #2760 caused by FUSE 3 dependency. @juliogonzalez

gaul:
I opened an issue against Ubuntu to hopefully upgrade before 26.04 releases: https://bugs.launchpad.net/ubuntu/+source/s3fs-fuse/+bug/2136274

juliogonzalez:
I am preparing everything now, and just by pure chance I discovered (for whoever is interested) that actually building on CentOS7 and clones is still possible, as FUS3E is at the `extras` repository.

The other problem is that the GCC version (4.8.5) is too old. But installing a newer version is possible:
```
yum install centos-release-scl-rh
sed -i -e '/^mirrorlist/d;/^#baseurl=/{s,^#,,;s,/mirror,/vault,;}' /etc/yum.repos.d/CentOS-*
yum install devtoolset-7
```
Gets you...
```
# /opt/rh/devtoolset-7/root/usr/bin/gcc -v
Using built-in specs.
COLLECT_GCC=/opt/rh/devtoolset-7/root/usr/bin/gcc
COLLECT_LTO_WRAPPER=/opt/rh/devtoolset-7/root/usr/libexec/gcc/x86_64-redhat-linux/7/lto-wrapper
Target: x86_64-redhat-linux
Configured with: ../configure --enable-bootstrap --enable-languages=c,c++,fortran,lto --prefix=/opt/rh/devtoolset-7/root/usr --mandir=/opt/rh/devtoolset-7/root/usr/share/man --infodir=/opt/rh/devtoolset-7/root/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-shared --enable-threads=posix --enable-checking=release --enable-multilib --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-gcc-major-version-only --enable-plugin --with-linker-hash-style=gnu --enable-initfini-array --with-default-libstdcxx-abi=gcc4-compatible --with-isl=/builddir/build/BUILD/gcc-7.3.1-20180303/obj-x86_64-redhat-linux/isl-install --enable-libmpx --enable-gnu-indirect-function --with-tune=generic --with-arch_32=i686 --build=x86_64-redhat-linux
Thread model: posix
gcc version 7.3.1 20180303 (Red Hat 7.3.1-5) (GCC) 
[...]

PR #2801, opened 2026-02-11
Changed string::assign to explicitly specify last argument

ggtakec:
https://github.com/s3fs-fuse/s3fs-fuse/issues/2772#issuecomment-3868137428

### Details
In CentOS7 + devtools-7(gcc (GCC) 7.3.1 20180303 (Red Hat 7.3.1-5)), the definition of std::string::assign is slightly different from other versions, causing the build to fail.
This is because there is no default value for `n` in `std::string& assign(const std::string& str, size_type pos, size_type n=npos)`.
This can be avoided by specifying `n` explicitly. Alternatively, or we can use std::string& assign(const char* pstr, size_type pos).
(This PR fixes by the former.)

Of course, I didn't want to make this change for suppoting C++03, but this single fix allows us to continue building on CentOS7.
There's also the question of how long we'll support CentOS7, but I would like to make this fix for now.

### DIFF
--- a/src/cache_node.cpp
+++ b/src/cache_node.cpp
@@ -1480,7 +1480,7 @@ bool DirStatCache::GetChildLeafNameHasLock(const std::string& strpath, std::stri
         return false;
     }
 
-    strLeafName.assign(strpath, GetPathHasLock().size());
+    strLeafName.assign(strpath, GetPathHasLock().size(), std::string::npos);
     if(strLeafName.empty()){
         return false;
     }


gaul (approved):
Looks like C++17 introduced the variant with a default parameter (see 8): https://en.cppreference.com/w/cpp/string/basic_string/assign.html

ggtakec:
Yes, that's true. However, CentOS 7's std::string doesn't have a default parameter, which causes a compilation error.
I would like to make this change to support building on CentOS 7.
In the s3fs-fuse source code, this assign method type is called in only one place.

juliogonzalez:
Hello!

I updated https://github.com/s3fs-fuse/s3fs-fuse/wiki/Installation-Notes#rhel7centos7etc as now I can confirm I could build `s3fs-fuse` with this patch and according to the instructions there :-)

PR #2807, opened 2026-02-23
Harden XML parsing against XXE

### LINKED ISSUE #2805: Set XML_PARSE_NONET on xmlReadMemory calls
All `xmlReadMemory()` calls in the codebase pass `options = 0`:

https://github.com/s3fs-fuse/s3fs-fuse/blob/b41e43df8a548b1d88f408b374959c39efcd4ca0/src/s3fs_xml.cpp#L467
https://github.com/s3fs-fuse/s3fs-fuse/blob/b41e43df8a548b1d88f408b374959c39efcd4ca0/src/s3fs.cpp#L3673
https://github.com/s3fs-fuse/s3fs-fuse/blob/b41e43df8a548b1d88f408b374959c39efcd4ca0/src/mpu_util.cpp#L115

S3 responses never contain DTDs or entity references. Setting `XML_PARSE_NONET` (available since libxml2 2.6) would explicitly prevent the parser from making network requests during entity resolution. On libxml2 >= 2.13, `XML_PARSE_NO_XXE` can additionally disable all external entity and DTD loading.

Currently, the code relies on libxml2 >= 2.9 defaulting to safe behavior, but doesn't declare it explicitly. A one-line flag per call would fix this.

Without the flags an attacker who can manipulate S3 responses (MITM, rogue S3-compatible endpoint) could potentially:
- read local files via XXE (libxml2 < 2.9)
- trigger SSRF, e.g. to EC2 IMDS for credential theft (libxml2 < 2.9)
- cause OOM via entity expansion / billion laughs (libxml2 < 2.11)

I'd be happy to submit a PR.

CarstenGrohmann:
All three `xmlReadMemory()` calls pass `options = 0`. S3 responses never use DTDs or entities, so the parser should reflect that.
                                                                                 
This PR:
- raises the minimum libxml2 from 2.6 to 2.9 (released 2012, all CI targets already have >= 2.9.7)
- sets `XML_PARSE_NO_XXE` on libxml2 >= 2.13 and `XML_PARSE_NONET` on 2.6–2.14 for defense in depth

Closes #2805

### DIFF
--- a/src/mpu_util.cpp
+++ b/src/mpu_util.cpp
@@ -112,7 +112,7 @@ int s3fs_utility_processing(time_t abort_time)
         // parse result(incomplete multipart upload information)
         S3FS_PRN_DBG("response body = {\n%s\n}", body.c_str());
 
-        std::unique_ptr<xmlDoc, decltype(&xmlFreeDoc)> doc(xmlReadMemory(body.c_str(), static_cast<int>(body.size()), "", nullptr, 0), xmlFreeDoc);
+        std::unique_ptr<xmlDoc, decltype(&xmlFreeDoc)> doc(xmlReadMemory(body.c_str(), static_cast<int>(body.size()), "", nullptr, S3FS_XML_PARSE_FLAGS), xmlFreeDoc);
         if(nullptr == doc){
             S3FS_PRN_DBG("xmlReadMemory exited with error.");
             result = EXIT_FAILURE;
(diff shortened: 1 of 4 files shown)


gaul (changes requested):

CarstenGrohmann:
I pushed a modified version of this change to be more focused on the change ifself.

gaul:
Thank you for your contribution @CarstenGrohmann!  I received a Claude Max subscription and will handle all of its suggested fixes.  I would rather outside contributors focus on actual issues instead of theoretical flagged ones.

PR #2835, opened 2026-04-06
Revert #2783 to call PreventStatCacheExpire again

ggtakec:
#2783 #2833

### Details
We temporarily disabled some caching functionality due to PR #2783, but the cause was as described in #2833(There was no connection between the bug and this feature), so I will restore this functionality.


ggtakec:
Please merge #2834 before this PR.

PR #2836, opened 2026-04-06
Temporarily bypass a test fails to delete directories in macOS

ggtakec:
#2833

### Details
As shown in #2833, the failure of the `test_truncate_cache` test is unlikely to be due to a bug in s3fs-fuse.
Therefore, we will modify the content of the `test_truncate_cache` test on macOS to avoid the bug in #2833 and prevent the test from failing.
This is a temporary workaround for #2833 and will be reverted once the issue is resolved.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -2124,8 +2124,20 @@ function test_truncate_cache() {
         ls "${dir}"
     done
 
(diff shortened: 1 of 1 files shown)


ggtakec:
Please merge #2834 before this PR.

PR #2838, opened 2026-04-11
Upgrade to S3Proxy 3.1.0

gaul:
Also set Content-Type for s3_cp test helper curl's --data-binary defaults Content-Type to application/x-www-form-urlencoded, which causes Jetty 12 to drain the request body as form parameters before S3Proxy 3.1.0 can hash it for x-amz-content-sha256 verification, failing the check with 400 XAmzContentSHA256Mismatch. Send application/octet-stream instead.  Release notes:

https://github.com/gaul/s3proxy/releases/tag/s3proxy-3.1.0


gaul:
For some reason older distros fail with newer S3Proxy.  Also the log lines are truncated.

ggtakec (approved):

ggtakec:
@gaul This might not be correct, but it seems like tests on macOS have been consistently failing (timeout) since this PR.
There might be other potential issues, but I wanted to share this. (I'm sorry for replying to this PR might not be ideal.)

PR #2839, opened 2026-04-11
Disable AppleDouble files in macOS integration tests

gaul:
macOS creates `._foo` sidecar files for extended attributes on filesystems that don't natively support them, which pollutes directory listings and breaks `test_file_names_longer_than_posix` (and cascades into `test_list`).  Pass `-o noappledouble` to s3fs on Darwin so the FUSE layer rejects them at mount time.

### DIFF
--- a/test/integration-test-common.sh
+++ b/test/integration-test-common.sh
@@ -290,10 +290,17 @@ function start_s3fs {
     # For macos fuse-t, we need to specify the "noattrcache" option to
     # disable NFS caching.
     #
+    # We also pass "noappledouble" so the macOS VFS does not create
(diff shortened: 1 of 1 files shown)


gaul:
This should resolve macOS CI failures observed in #2838.

ggtakec (approved):
Thank you for the noappledouble option.

ggtakec:
@gaul please rebase base-branch to master.

gaul:
This is not working as expected.  It seems that @macos-fuse-t does not support this option, only macFUSE.

ggtakec:
Understood. I hope it will be supported.

gaul:
This no longer seems necessary since macOS CI succeeds.

PR #2840, opened 2026-04-12
Set cache permissions to 700

gaul:
Claude flagged this world-writeable directory.


ggtakec (approved):
I initially thought 0777 would be better, but there are security issues with that.
Thank you.

PR #2842, opened 2026-04-12
Fix TOCTOU in `mkdirp`

gaul:
It is safer to create the directory and let it fail with `EEXIST` than to check then create.  Suggested by Claude.


ggtakec (approved):
Thanks and Thanks and for changing `stat` to `lstat`.

PR #2845, opened 2026-04-12
Change default signature type to v4-only

gaul:
Users can override this via `-o sigv2`.  Also remove the fallback logic. Suggested by Claude as a security improvement.


ggtakec (approved):
Thanks, A considerable amount of time has passed, so I think we can eliminate the fallback to sigv2.

Currently, how many backends still need to connect using sigv2?
It might be good to know that, if possible.

ggtakec:
It would probably be best to rebase to master just to be safe.

gaul:
It's hard to say if v2 is used in a meaningful way 10 years after v4 introduction.  But the v4-to-v2 fallback is potentially dangerous and users can opt-into v2 if they have a modern s3fs with an ancient object store.

PR #2851, opened 2026-04-16
Write log lines with write(2) instead of stdio

### LINKED ISSUE #2850: Readdir hangs on windows server 2022
I met some problems compiling s3fs-fuse on windows 2022, so I switched to ossfs (based on 1.91) and succeeded to mount it. But the process is stuck when I keep open some directories, and I figured out that it's the s3fs_low_logprn caused the problem. 
And the s3fs_logger in ossfs is the same as s3fs-fuse, so I think the same problem also exists for s3fs-fuse.
It will help a lot if you can help to check it out.

I compile ossfs and run on `Windows Server 2022`, mount as disk Z:\.
options are `-ologfile=xxx -ourl=xxx -oumask=000 -oallow_other -odbglevel=dbg`
After I double clicked Z:\ (about 50 sub-items inside), and opened several sub-folders, the process is blocked.

I use gdb to dump the stack:

```
#0  0x00007ff9ceec0ea4 in ntdll!ZwWaitForMultipleObjects () from /c/Windows/SYSTEM32/ntdll.dll  
#1  0x00007ff9cc82472c in WaitForMultipleObjectsEx () from /c/Windows/System32/KERNELBASE.dll  
#2  0x00007ff9cc82462e in WaitForMultipleObjects () from /c/Windows/System32/KERNELBASE.dll  
#3  0x00000001800463ae in ?? () from /c/ossfs/bin/msys-2.0.dll  
#4  0x00000001800ad923 in ?? () from /c/ossfs/bin/msys-2.0.dll  
#5  0x00000001801fc9dc in ?? () from /c/ossfs/bin/msys-2.0.dll  
#6  0x00000001801ac934 in ?? () from /c/ossfs/bin/msys-2.0.dll  
#7  0x0000000700000000 in ?? ()  
#8  0x0000000a001edf10 in ?? ()  
#9  0x00000001004bb7a0 in refptr.utility_mode ()  
#10 0x0000000a01b411a0 in ?? ()  
#11 0x00000007ff9fcc10 in ?? ()  
#12 0x000000010041d552 in s3fs_low_logprn(S3fsLog::s3fs_log_level, char const*, char const*, int, char const*, ...) ()  
#13 0x0000000a01b412a0 in ?? ()  
#14 0x0000000a000b9bd0 in ?? ()  
#15 0x00000005d244c362 in ?? () from /c/ossfs/bin/msys-curl-4.dll  
#16 0x00000005d24408c7 in ?? () from /c/ossfs/bin/msys-curl-4.dll  
#17 0x000000010042bed4 in S3fsCurl::RequestPerform(bool) ()
```

It seems to be the `s3fs_low_logprn` function that caused the problem. When multiple threads calls the `s3fs_low_logprn` (actually it's fprintf) in MSYS2, a deadlock will come out.

gaul:
fprintf on MSYS2 deadlocks under concurrent logging because the runtime's per-FILE stdio lock interacts badly with its signal-aware wait primitive, hanging s3fs during readdir on Windows (issue #2850).

Format each line into a buffer and emit it with a single write(2) call. Since logfp is opened with O_APPEND, the kernel guarantees atomic appends, so log lines also no longer interleave across threads on other platforms.

Fixes #2850.

### DIFF
--- a/src/s3fs_logger.cpp
+++ b/src/s3fs_logger.cpp
@@ -18,6 +18,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+#include <cerrno>
 #include <cstdarg>
 #include <cstdlib>
 #include <iomanip>
@@ -25,6 +26,7 @@
 #include <sstream>
 #include <string>
 #include <sys/time.h>
+#include <unistd.h>
 
 #include "common.h"
 #include "s3fs_logger.h"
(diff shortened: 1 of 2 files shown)


ggtakec (approved):
That's a precise correction. Thank you!

Issue #1944, opened 2022-05-11
Bad cache state when moving/renaming files results in the new file having all bytes set to 0


rcfox:
### Additional Information
_The following information is very important in order to help us to help you. Omission of the following details may delay your support request or receive no attention at all._
_Keep in mind that the commands we provide to retrieve information are oriented to GNU/Linux Distributions, so you could need to use others if you use s3fs on macOS or BSD_

#### Version of s3fs being used (s3fs --version)
v1.89, v1.90, v1.91

#### Version of fuse being used (pkg-config --modversion fuse, rpm -qi fuse, dpkg -s fuse)
Version: 2.9.9-3

#### Kernel information (uname -r)
5.13.0-39-generic

#### GNU/Linux Distribution, if applicable (cat /etc/os-release)
```
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

#### s3fs command line used, if applicable
```
s3fs BUCKET /mnt/BUCKET -f -o allow_other -o mp_umask=0111 -o nonempty -o use_cache=/tmp
```
#### /etc/fstab entry, if applicable
N/A

#### s3fs syslog messages (grep s3fs /var/log/syslog, journalctl | grep s3fs, or s3fs outputs)
_if you execute s3fs with dbglevel, curldbg option, you can get detail debug messages_

Log showing bad behaviour 1 below.
```
unique: 760, opcode: GETATTR (3), nodeid: 4, insize: 56, pid: 0
getattr /outgoing
 [INF] s3fs.cpp:s3fs_getattr(763): [path=/outgoing]
   unique: 760, success, outsize: 120
unique: 762, opcode: LOOKUP (1), nodeid: 1, insize: 49, pid: 0
LOOKUP /outgoing
getattr /outgoing
 [INF] s3fs.cpp:s3fs_getattr(763): [path=/outgoing]
   NODEID: 4
[...]

PR #2852, opened 2026-04-19
Add regression tests for rename content (issue #1944)

eran132:
## Summary

Issue #1944 reports that renaming a file whose local cache does not exist (created on another machine, or after cache cleanup) produces a correctly-sized but zero-filled file. That bug was fixed in e4f85c1 ("Fixed renaming bug with SSE", Jan 2023): the rename path now passes real timestamps to `FdEntity::Open`, so the freshly-created pagelist is marked *not loaded* rather than *modified*, and subsequent reads correctly fall back to S3.

The existing `test_mv_file` only compares file **length** after rename, so this class of bug had no CI coverage.

This PR:

- Extends `test_mv_file` to also verify content equality after rename.
- Adds `test_mv_file_nocache_content`, which explicitly deletes the local cache file (including `.stat` and `.mirror` entries) before the rename, to exercise the #1944 path directly. Skips cleanly on runs without `use_cache`.

## Verification

To confirm the new assertions actually catch the #1944 bug, I locally reverted the fix (`src/fdcache_entity.cpp` lines 517 and 568 — restore `pagelist.Resize(size, false, true)`) and re-ran the suite. Both new assertions fail with zero-filled content, as expected.

## Test plan

- [x] `make check -C src` — unit tests green
- [x] `make cppcheck` — clean
- [x] `make shellcheck` — clean
- [x] `make check -C test` (default `sigv4`) — 89/89
- [x] `ALL_TESTS=1 make check -C test` (10 flag variants: use_cache+update_parent_dir_stat, enable_content_md5, disable_noobj_cache, max_stat_cache_size, nocopyapi, nomultipart, sigv2, sigv4, singlepart_copy_limit, streamupload) — all pass, new tests run and skip correctly depending on `use_cache`


ggtakec:
@eran132 Thank you for this PR.

I've reviewed the suggestion from Copilot, and I don't think checking for null bytes is necessary.
If null bytes are present, the LENGTH check will result in an error, so Copilot's suggestion is excessive.

But, I would like the following line modified to handle the case where `ALT_FILE_CONTENT` is empty.
```
Your code:
    if [ "${ALT_FILE_CONTENT}" != "${TEST_TEXT}" ]

My request:
    if [ -z "${ALT_FILE_CONTENT}" ] || [ "${ALT_FILE_CONTENT}" != "${TEST_TEXT}" ]
```

eran132:
Thanks for the review! Applied the `-z` guard to both `test_mv_file` and `test_mv_file_nocache_content` (for consistency) in 62a6858. Verified locally: shellcheck clean, both tests still pass on master and still fail on a build with `pagelist.Resize(size, false, ts_times.IsOmitMTime())` reverted to `false, true`.

gaul:
Thanks for adding coverage for #1944 — and especially for verifying the new assertions actually catch the bug by reverting the fix locally. A few notes from a review pass:

**Likely bug in the cache cleanup**

```
rm -f "${CACHE_DIR}/.${TEST_BUCKET_1}.mirror/${CACHE_TESTRUN_DIR}/${TEST_TEXT_FILE}"
```

The `.mirror` directory contains hardlinks keyed by inode, not by relative path, so this `rm -f` silently no-ops. Either drop the line (the cache file + `.stat` removal is what forces the reload path) or look up the entry by inode. As written it gives a false sense of coverage.

**`CACHE_TESTRUN_DIR` derivation diverges from the established pattern**

Sibling tests (e.g. lines 2151, 2252, 2493, 2537) receive `CACHE_TESTRUN_DIR` as `$1` from the test harness. Deriving it via `basename "$(pwd)"` happens to work but is fragile — worth taking it as a parameter for consistency.

**`sync` may not be sufficient under all flag variants**

`sync(1)` flushes the kernel page cache through FUSE, but s3fs's own upload can be asynchronous depending on options. If the object hasn't reached S3 when the local cache is dropped, the reload returns empty and the test would fail for the wrong reason. The PR description says `ALL_TESTS=1` passed across 10 variants, so this is probably fine in practice — a short comment explaining why `sync` is enough would help future readers.

**Minor**

- `[ -z "${ALT_FILE_CONTENT}" ] || [ "${ALT_FILE_CONTENT}" != "${TEST_TEXT}" ]` — the second clause already covers the empty case (since `TEST_TEXT` is non-empty).
- `rm -f "${TEST_TEXT_FILE}" "${ALT_TEST_TEXT_FILE}"` at the top is unusual; most tests rely on the harness for a clean working dir.

Overall a useful regression test — would approve after the `.mirror` line is fixed or removed.

eran132:
Thanks @gaul, all good catches. Pushed 30f6767 with the changes:

**1. `.mirror` cleanup** — confirmed against `FdEntity::OpenMirrorFile` (src/fdcache_entity.cpp:280); mirror filenames are random hex via `snprintf("%x.tmp", rd())` so the path-based rm can never match. Dropped the line and added an inline comment so future readers don't try to put it back.

**2. `CACHE_TESTRUN_DIR=$1`** — switched to the harness-provided parameter, matching `test_cache_file_stat` and friends.

**3. `sync` semantics** — replaced with an explicit cat-based preflight. The reasoning (in the inline comment): if async upload hasn't reached S3 yet, the subsequent rename's S3-side copy fails with 404, which is a distinguishable error mode from the bug we're checking for (a successful rename to an all-zeros file). The preflight ensures we only enter the bug-test region when the source is genuinely ready.

I also added two more preflights guarding against the same silent-no-op class you flagged on `.mirror`: confirm the cache file is at the expected path before `rm`, and confirm it's gone after. If a future cache-layout change moves the file, the test will scream rather than pass for the wrong reason.

**4. Redundant `-z` clause** — kept it. @ggtakec specifically requested this exact form in [PR comment](https://github.com/s3fs-fuse/s3fs-fuse/pull/2852#issuecomment-4275797339). Happy to take it out if the two of you agree it's excess.

**5. Leading `rm -f`** — removed.

Verification on 30f6767:

| Gate | Result |
|---|---|
| `make shellcheck` | clean |
| `make cppcheck` | clean |
| `make check -C src` | 3/3 PASS |
| narrow run on master (test_mv_file + test_mv_file_nocache_content) | 2/2 PASS |
| narrow run on revert-fix build (`pagelist.Resize(size, false, true)` restored at fdcache_entity.cpp:517+568) | 0/2 PASS — `test_mv_file` fails on the content assertion, `test_mv_file_nocache_content` fails at the new preflight (which catches the regression *before* the rename, an even sharper signal than the original assertion). Both expected. |
| `ALL_TESTS=1 make check -C test` | 879/879 PASS across all 10 flag variants (use_cache+update_parent_dir_stat, enable_content_md5, disable_noobj_cache, max_stat_cache_size=100, nocopyapi, nomultipart, sigv2, sigv4, singlepart_copy_limit=10, streamupload). `test_mv_file_nocache_content` ran to PASS in every variant: 8 clean skips on no-use_cache flags, 2 full runs on use_cache flags. |

Local environment: WSL2 Ubuntu-24.04, s3proxy-3.0.0, gcc 13.3.0, fuse 2.9.9 + fuse3 3.14.0.

ggtakec:
@eran132 
> 4. Redundant -z clause — kept it. @ggtakec specifically requested this exact form in https://github.com/s3fs-fuse/s3fs-fuse/pull/2852#issuecomment-4275797339. Happy to take it out if the two of you agree it's excess.

My request was incorrect. Since this is not a numeric comparison, the -z conditional statement is unnecessary.
Please take it out.

eran132:
Done — bc5c5ff drops the `-z` clauses in `test_mv_file` and `test_mv_file_nocache_content` (3 sites). Re-ran shellcheck, cppcheck, and the narrow integration on both master (2/2 PASS) and a locally-reverted-fix build (still fails as expected, catching the bug at the preflight). Skipped a fresh ALL_TESTS=1 since this is a textual no-op — removing a disjunct that never short-circuited in any of the 879 cases from the prior commit.

Thanks for the careful re-review.

gaul:
Rebased to get rid of OpenSUSE 15 CI target.

PR #2853, opened 2026-04-19
Detection _FORTIFY_SOURCE value by configure.ac

ggtakec:
n/a

### Details
This is a minor change. 
Since `_FORTIFY_SOURCE=3` is not supported on macOS, we've added a check for this flag during the configure process.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -37,7 +37,47 @@ AC_CHECK_FUNCS([fallocate])
 CPP_VERSION=c++14
 AC_SUBST([CPP_VERSION])
 
-CXXFLAGS="-Wall -fno-exceptions -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=3 -std=$CPP_VERSION $CXXFLAGS"
+dnl ----------------------------------------------
+dnl Set _FORTIFY_SOURCE flag
+dnl ----------------------------------------------
+dnl Some platforms treat >2 as 2 and emit a warning from features.h.
+dnl Also, it may be built-in.
+dnl
+FORTIFY_FLAG=""
+
+AC_MSG_CHECKING([checking _FORTIFY_SOURCE])
+
+dnl
+dnl Compile a simple program and check the glibc warnings in the standard error output.
+dnl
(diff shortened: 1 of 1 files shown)


ggtakec:
The results are still insufficient, so I will change this to a draft.

ggtakec:
There are operating systems where `_FORTIFY_SOURCE=2` is required, and others defined as `built-in`, so the conditional statement has been modified.
Currently, the CI tests have passed, so the draft status has been removed.

gaul:
I think maybe we should remove `_FORTIFY_SOURCE` since distributions usually set this and s3fs can interfere with whatever hardening flags they set.  From Claude:

> ❯ Should this project set _FORTIFY_SOURCE?  Or should distributions do this?                                                           
> 
> ● Distributions should own it, not the project.
> 
>   _FORTIFY_SOURCE needs to be paired with an optimization level (-O1+) and a libc that supports it (glibc, recent musl). Hardcoding it in configure.ac breaks CFLAGS=-O0 debug builds (warning spam) and is redundant on distros like Fedora/Debian/Ubuntu/SUSE that already inject it via rpm macros or dpkg-buildflags. Homebrew/macOS and *BSD don't want it.
> 
>   The common practice for upstream C/C++ projects is to leave hardening flags (_FORTIFY_SOURCE, -fstack-protector-strong, -D_GLIBCXX_ASSERTIONS, PIE, RELRO) to the packager's CFLAGS/CXXFLAGS/LDFLAGS. If you want to help, document recommended hardening flags in INSTALL/README rather than forcing them.
> 
>   Exception: if you find users commonly building from source without distro wrappers and shipping to production, a configure-time --enable-hardening opt-in is a reasonable middle ground.

ggtakec:
Currently, I've confirmed that some distributions don't have this flag built-in, and that the level (value) varies across distributions.
This PR code will detect these and configure it so as not to interfere with the security hardening flag.
I think, looking at the current OS trends in our CI, it seems likely that this flag will eventually become unset(supports built-in).
How about merging this PR, observing the situation for a while, and then deleting it?

PR #2857, opened 2026-04-26
Treat any key ending with '/' as a directory

### LINKED ISSUE #2856: GCS "Directories" are not treated as directories
#### Version of s3fs being used (`s3fs --version`)
Amazon Simple Storage Service File System V1.93 (commit:unknown) with GnuTLS(gcrypt)

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)
Version: 3.14.0-5build1

#### Provider (`AWS`, `OVH`, `Hetzner`, `iDrive E2`, ...)
GCS

#### Kernel information (`uname -r`)
6.8.0-86-generic

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)
Ubuntu 24.04.3 LTS

#### How to run s3fs, if applicable

[x] command line  
[] /etc/fstab  

```
s3fs -o passwd_file=/etc/gcs-auth.txt -o url=https://storage.googleapis.com -o sigv2 -o nomultipart my-storage /mnt/storage
```

#### s3fs syslog messages (`grep s3fs /var/log/syslog`, `journalctl | grep s3fs`, or `s3fs outputs`)

```
2026-04-25T22:01:48.897459+00:00 vagrant s3fs[17068]: s3fs version 1.93(unknown) : s3fs -o passwd_file=/etc/gcs-auth.txt -o url=https://storage.googleapis.com -o sigv2 -o nomultipart my-storage /mnt/storage
2026-04-25T22:01:48.899265+00:00 vagrant s3fs[17068]: Loaded mime information from /etc/mime.types
2026-04-25T22:01:48.905385+00:00 vagrant s3fs[17069]: init v1.93(commit:unknown) with GnuTLS(gcrypt), credential-library(built-in)
```

### Details about issue

Google Cloud Storage mounted buckets do not treat "sub-directories" as directories

```
root@sandbox:/mnt/storage# ls -la
total 4
drwx------ 1 root root    0 Jan  1  1970 .
drwxr-xr-x 5 root root 4096 Apr 25 22:01 ..
-rwxr-x--- 1 root root    0 Mar 11 08:01 2650511c-0772-4da9-91e8-9360eed71260
-rwxr-x--- 1 root root    0 Apr 13 16:04 36910533-a3f8-43dd-96dd-0165dce0cf4e
-rwxr-x--- 1 root root    0 Mar 17 13:55 525f8137-83d0-4c14-bb02-8c8735e8c6c0
-rwxr-x--- 1 root root    0 Mar 12 17:47 5368ff7b-9ccc-4690-9dc5-72a35515f383
-rwxr-x--- 1 root root    0 Apr 16 16:32 844dd0ab-eb86-4267-8188-faca1e7e9415
-rwxr-x--- 1 root root    0 Apr 21 11:08 bab253ac-75a8-4cd1-8338-51952626c2ac
```

All UUID-like entries are "directories", but they are not treated like directories, so there is no way to cd into them and list content

gaul:
A key ending with '/' cannot be a regular object in S3/GCS, so use the trailing slash itself as a sufficient signal in get_mode() instead of requiring a directory-specific Content-Type. This fixes GCS-mounted buckets where placeholder objects use Content-Type text/plain (or other non-allowlisted types), which previously caused subdirectories to appear as -rwxr-x--- regular files and be unenterable. Fixes #2856.


betalb:
Good day @gaul, thank you for the quick fix. I've checked it and it works for me. Those folders were created by gcsfuse so they indeed have different content-type.

There is a related problem (or expected behavior) that mounting bucket with a prefix pointing to such directory fails by default. It succeeds with `compat_dir` flag, but most likely due to the fact that the check result is ignored in `remote_mountpath_exists`.

ggtakec:
@gaul 
Let's discard the `complement_stat` option and make `complement_stat=true` the default.
Could you make additional corrections for it(removing `complement_stat` option) ?

##### Memo
_s3fs was the first to configure directories using the `<dir>/` object.
Before that, other clients used `Content-Type` or the `<dir>_$folder$` object or not used dir object.
s3fs made this complex decision to accommodate all of these, but I think if you agree this `complement_stat` is no longer necessary._

src/s3fs.cpp
```diff
@@ -5549,7 +5549,7 @@ static int my_fuse_opt_proc(void* data, const char* arg, int key, struct fuse_ar
             return 0;
         }
         else if(0 == strcmp(arg, "complement_stat")){
-            complement_stat = true;
+            S3FS_PRN_WARN("complement_stat is enabled by default and a future version will remove this option.");
```
ggtakec, line 5552: Thank you for maintaining backward compatibility.

ggtakec (approved):

PR #2859, opened 2026-05-01
Remove openSUSE Leap 15 reaching EOL from CI

ggtakec:
n/a

### Details
In the OpenSUSE/Leap:15 CI environment, the installation of libxml2-devel is failing.
This is because libxml2-devel depends on readline-devel, which requires libncurses6, resulting in a version mismatch in the current package.
It seems that running dist-upgrade is necessary to resolve this issue.

@juliogonzalez
Is this solution acceptable?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,6 @@ jobs:
           - rockylinux/rockylinux:8
           - fedora:44
           - fedora:43
-          - opensuse/leap:15
           - opensuse/leap:16.0
           - alpine:3.23
 
--- a/.github/workflows/linux-ci-helper.sh
+++ b/.github/workflows/linux-ci-helper.sh
@@ -318,8 +318,7 @@ elif [ "${CONTAINER_FULLNAME}" = "fedora:44" ] ||
         procps
     )
 
-elif [ "${CONTAINER_FULLNAME}" = "opensuse/leap:15" ] ||
-     [ "${CONTAINER_FULLNAME}" = "opensuse/leap:16.0" ]; then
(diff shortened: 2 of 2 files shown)


gaul:
Alternatively, [Leap 15 is now EOL](https://news.opensuse.org/2026/04/01/leap-15-eol/) so we can remove it from the CI matrix.

ggtakec:
I agree with removing Leap:15 from CI.

@juliogonzalez
Is it okay to exclude Leap:15 from CI?
Please give us your opinion.

juliogonzalez:
> @juliogonzalez Is it okay to exclude Leap:15 from CI? Please give us your opinion.

Sorry about the delay! :-)

FMPOV, go ahead and remove.

Here's the situation, for future reference:
- openSUSE Leap 15.6 was supposed to be EoL on April 30th. But at the same time got some last minute fixes, including for [CVE-2026-31431](https://www.suse.com/security/cve/CVE-2026-31431.html) (Copy Fail)
- There should be no more updates to openSUSE Leap 15.6 after this.
- While it is true that SUSE Linux Enterprise 15 SP7 (and older service packs) are a thing, we'll still have https://build.opensuse.org/package/show/filesystems/s3fs in best-effort, making sure that at least it builds (we could even consider including some tests at some point as part of the build process).
- The SUSE Package Hub doesn't even package s3fs-fuse for SUSE Linux Enterprise 15 but rather only for SUSE Linux Enteprise 16.

ggtakec:
@juliogonzalez 
Thanks for the explanation.
Then, we will decide to remove Leap:15 from the CI.

@gaul 
I'll fix this PR and re-upload the code.

PR #2861, opened 2026-05-05
Switch macOS to FUSE 3 (libfuse3)

### LINKED ISSUE #2802: macFUSE supports FUSE 3
It appears that macFUSE supports FUSE 3: https://github.com/gromgit/homebrew-fuse/issues/170.  s3fs currently compiles against FUSE-T which only supports FUSE 2.  Moving from FUSE-T back to macFUSE would let s3fs remove FUSE 2 support.  @macos-fuse-t

gaul:
FUSE-T 1.2+ and macFUSE 5+ both ship libfuse3, so drop the FUSE 2 code path on macOS and use the same FUSE 3 API as Linux. Disable FUSE-T's Darwin-flavored operation overloads so the upstream libfuse3 signatures are used everywhere, and remove the position-aware setxattr/getxattr variants that only existed for the old macOS FUSE 2 API.  Fixes #2802.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -1050,12 +1050,19 @@ function test_update_time_xattr() {
            return 1
         fi
     else
-        # [macos] fuse-t
-        # atime/ctime are all updated.
-        # see) https://github.com/macos-fuse-t/fuse-t/issues/87
+        # [FIXME] macos fuse-t
+        # On FUSE-T 1.0.x (libfuse2) atime+ctime were both updated by setxattr
+        # (https://github.com/macos-fuse-t/fuse-t/issues/87). On FUSE-T 1.2.x
+        # (libfuse3) the kernel-cached stat returned to userspace does not
+        # reflect the ctime bump, so all three timestamps appear unchanged.
+        # Permit either observed behavior for now.
         #
-        if [ "${base_atime}" = "${atime}" ] || [ "${base_ctime}" = "${ctime}" ] || [ "${base_mtime}" != "${mtime}" ]; then
-           echo "set_xattr expected updated ctime: $base_ctime != $ctime, atime: $base_atime != $atime and same mtime: $base_mtime == $mtime"
+        if [ "${base_mtime}" != "${mtime}" ]; then
+           echo "set_xattr expected unchanged mtime: $base_mtime == $mtime"
+           return 1
+        fi
+        if [ "${base_atime}" != "${atime}" ] && [ "${base_ctime}" = "${ctime}" ]; then
+           echo "set_xattr expected updated ctime when atime is updated: $base_ctime != $ctime"
            return 1
         fi
     fi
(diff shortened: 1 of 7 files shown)


test/integration-test-main.sh
```diff
@@ -2107,37 +2114,26 @@ function test_content_type() {
 function test_truncate_cache() {
     describe "Test make cache files over max cache file size ..."
 
+    # FIXME:
+    # On macos-fuse-t (libfuse3), readdir on directories with many small files
+    # returns duplicate/missing entries (e.g. "28 28 29 29" with 26, 27 absent),
+    # which causes the subsequent rm -rf to fail. Skip until FUSE-T fixes this.
```
gaul, line 2120: @macos-fuse-t seems broken

macos-fuse-t: Could you please tell me how to run the test? When I run it directly this is what I get:
```
bash ./integration-test-main.sh
ps: Invalid process id: $>\v�\001�mG�)��\001
ps: Invalid process id: $>\v�\001
ps: Invalid process id: $>\v�\001�~O�)��\001
ps: Invalid process id: $>\v�\001�V�)��\001
ps: Invalid process id: $>\v�\001
ps: Invalid process id: $>\v�\001�R2�)��\001
TEST_BUCKET_MOUNT_POINT_1 variable not set
```

gaul: You can run `make -C test/ check` which sets up all the environment.

macos-fuse-t: the test fails with
```
s3fs: [INF] s3fs.cpp:main(5982): Free space ratio default to 10 %, ensure the available disk space is greater than 94858.416 MB
s3fs: [WAN] s3fs.cpp:main(5994): Not enough disk space for s3fs, try to clean cache dir
```
Free space seems to be determined from "df -h" command which might be misleading on macOS. 
Screenshot 2026-05-07 at 8 05 10
As you can see on the attached image my disk has 365GB of free space. 
In any case, is there a switch to disable caching so I can run the tests? 
Thank you

gaul: Something seems off with this logic.  Let me investigate but you can use `-o ensure_diskfree=0` to work around this behavior.

gaul: Can you share more complete logs for this failure?  Specifically I don't see this log message:
> There is not enough disk space for use as cache(or temporary) directory by s3fs. Requires %.3f MB, already has %.3f MB.
So I don't believe that this is related.

ggtakec: > duplicate/missing entries

Is this issue unrelated to issue #2833 that I reported previously?
(I suspect these issues likely have the same cause.)

> Not enough disk space for s3fs

I'll look into this as well. There might be an issue with the logic.

gaul: > @macos-fuse-t seems broken

CI passed this time, perhaps fixed by https://github.com/macos-fuse-t/fuse-t/releases/tag/1.2.5 ?

ggtakec (approved):

PR #2863, opened 2026-05-05
Configure Dependabot

gaul:
This is not too useful for C++ projects but will upgrade GitHub actions.


ggtakec (approved):
It may not happen very often, but I think it's useful because it can catch updates to the GitHub Actions component.

Issue #2581, opened 2024-10-30
File is truncated after Backend server being stopped and recovered


hbao0915:
#### Version of s3fs being used (`s3fs --version`)
v1.87, v1.91, v1.95

### Details about issue

The bug is in the routine: s3fs_open -> check_object_access -> get_object_attribute

Inside `get_object_attribute`, if the backend server goes south now, an HEAD with 503 response (or other 5xx resp, or timeout err) would be returned, then s3fs tries to send HEAD with filename plus "_$folder$". If the backend server recovers now, a 404 resp will be returned. Then s3fs consider the file not existing, and s3fs would call FdManager::Open with st.size=0, which will result in file being truncated.

eran132:
@hbao0915 thanks for the detailed trace. I dug through the code and confirmed the path you describe still exists on master.

The bug is in `get_object_attribute` (src/s3fs.cpp:439): the fallback to `path/`, `path_$folder$`, and `directory_empty()` runs whenever the first HEAD returns any non-zero result, including transient `-EIO` from exhausted 5xx retries (`RequestPerform`, src/curl.cpp:2210). When the backend recovers mid-cascade and answers `404` to the directory variants, the function returns `-ENOENT` and the caller fabricates a fresh empty cache — silent data corruption on the next flush.

This is the inverse of what every comparable S3 FUSE / sync tool does:

- **mountpoint-s3** (AWS official): [`mountpoint-s3-fs/src/fs/error.rs:127-143`](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-fs/src/fs/error.rs). `PrefetchReadError::GetRequestFailed` → `libc::EIO`. Only `InodeError::FileDoesNotExist` / `InodeError::InodeDoesNotExist` → `ENOENT`. No cascade that can convert transient errors to not-found.
- **rclone**: [`backend/s3/s3.go`](https://github.com/rclone/rclone/blob/master/backend/s3/s3.go). `retryErrorCodes = [429, 500, 503]` retry through `shouldRetry`; only `http.StatusNotFound` is converted to `fs.ErrorDirNotFound`.
- **goofys**: [`internal/goofys.go::mapHttpError`](https://github.com/kahing/goofys/blob/master/internal/goofys.go). 404 → `fuse.ENOENT`; 429/500 → `syscall.EAGAIN`. One-to-one, no cross-mapping.

[...]

PR #2866, opened 2026-05-06
Strictly mapping 500 range in HTTP  response to errno codes

ggtakec:
#2581

### Details
This PR strictly handles HTTP responses in the 500 range (and 429) from the S3 server, mapping them to `errno`(negative) codes.

This PR will change the return code from `S3fsCurl::RequestPerform` as follows:

| HTTP response code | return code(=errno)     |
| ------------------ | ----------------------- |
| 429                | -EAGAIN(after retrying) |
| 501                | -ENOTSUP                |
| 500                | -EIO                    |
| 502                | -ECONNRESET             |
| 503                | -EAGAIN                 |
| 504                | -ETIMEDOUT              |

This addresses the issue pointed out in #2581.

For example, if a HEAD request receives a 503 response, `get_object_attribute` will return `-EAGAIN`.
This means that, for example, when `cat`ing a file, it will look like this:

```
$ cat <file>
  cat: <file>: Resource temporarily unavailable
```

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -1661,9 +1661,6 @@ bool S3fsCurl::RemakeHandle()
     headdata.clear();
     LastResponseCode   = S3FSCURL_RESPONSECODE_NOTSET;
 
-    // retry count up
-    retry_count++;
-
     // set from backup
     postdata           = b_postdata;
     postdata_remaining = b_postdata_remaining;
@@ -1956,6 +1953,20 @@ bool S3fsCurl::RemakeHandle()
 }
 
 //
+// Sleep Method with Exponential Backoff and Jitter
+//
+// [NOTE]
+// The primary objective of this process is to prevent secondary server
+// failures caused by spikes in simultaneous retries.
+//
+void S3fsCurl::WaitBeforeRetry()
+{
+    unsigned int sleep_time = 2 << retry_count;
+    sleep(sleep_time + static_cast<unsigned int>(random()) % sleep_time);
+    retry_count++;
+}
+
+//
 // returns curl return code
 //
 int S3fsCurl::RequestPerform(bool dontAddAuthHeaders /*=false*/)
@@ -2065,22 +2076,71 @@ int S3fsCurl::RequestPerform(bool dontAddAuthHeaders /*=false*/)
                         result = -EIO;
                         break;
 
+                    case 429:
+                        if((retrycnt + 1) < S3fsCurl::retries){
+                            S3FS_PRN_INFO3("HTTP response code 429 was returned, slowing down");
+                            S3FS_PRN_DBG("Body Text: %s", bodydata.c_str());
+                            WaitBeforeRetry();
+                        }else{
+                            S3FS_PRN_INFO3("HTTP response code 429 was returned, returning EAGAIN");
+                            result = -EAGAIN;
+                        }
+                        break;
+
+                    case 500:
+                        // [NOTE]
+                        // The 500 error message occurs when the server is unable to process the request at
+                        // that time, and may be resolved by retrying the request.
+                        //
+                        if((retrycnt + 1) < S3fsCurl::retries){
+                            S3FS_PRN_INFO3("HTTP response code 500 was returned, slowing down");
+                            S3FS_PRN_DBG("Body Text: %s", bodydata.c_str());
+                            WaitBeforeRetry();
+                        }else{
+                            S3FS_PRN_INFO3("HTTP response code 500 was returned, returning EIO");
+                            result = -EIO;
+                        }
+                        break;
+
                     case 501:
                         S3FS_PRN_INFO3("HTTP response code 501 was returned, returning ENOTSUP");
                         S3FS_PRN_DBG("Body Text: %s", bodydata.c_str());
                         result = -ENOTSUP;
                         break;
 
-                    case 429:
-                    case 500:
-                    case 503: {
-                        S3FS_PRN_INFO3("HTTP response code %ld was returned, slowing down", responseCode);
-                        S3FS_PRN_DBG("Body Text: %s", bodydata.c_str());
-                        // Add jitter to avoid thundering herd.
-                        unsigned int sleep_time = 2 << retry_count;
-                        sleep(sleep_time + static_cast<unsigned int>(random()) % sleep_time);
+                    case 502:
+                        if((retrycnt + 1) < S3fsCurl::retries){
+                            S3FS_PRN_INFO3("HTTP response code 502 was returned, slowing down");
+                            S3FS_PRN_DBG("Body Text: %s", bodydata.c_str());
+                            WaitBeforeRetry();
+                        }else{
+                            S3FS_PRN_INFO3("HTTP response code 502 was returned, returning EWOULDBLOCK");
+                            result = -EWOULDBLOCK;
+                        }
                         break;
-                    }
+
+                    case 503:
+                        if((retrycnt + 1) < S3fsCurl::retries){
+                            S3FS_PRN_INFO3("HTTP response code 503 was returned, slowing down");
+                            S3FS_PRN_DBG("Body Text: %s", bodydata.c_str());
+                            WaitBeforeRetry();
+                        }else{
+                            S3FS_PRN_INFO3("HTTP response code 503 was returned, returning EAGAIN");
+                            result = -EAGAIN;
+                        }
+                        break;
+
+                    case 504:
+                        if((retrycnt + 1) < S3fsCurl::retries){
+                            S3FS_PRN_INFO3("HTTP response code 504 was returned, slowing down");
+                            S3FS_PRN_DBG("Body Text: %s", bodydata.c_str());
+                            WaitBeforeRetry();
+                        }else{
+                            S3FS_PRN_INFO3("HTTP response code 504 was returned, returning ETIMEDOUT");
+                            result = -ETIMEDOUT;
+                        }
+                        break;
+
                     default:
                         S3FS_PRN_ERR("HTTP response code %ld, returning EIO. Body Text: %s", responseCode, bodydata.c_str());
                         result = -EIO;
--- a/src/curl.h
+++ b/src/curl.h
@@ -235,6 +235,7 @@ class S3fsCurl
         // methods
         bool ResetHandle() REQUIRES(S3fsCurl::curl_handles_lock);
         bool RemakeHandle();
+        void WaitBeforeRetry();
         bool ClearInternalData();
         bool insertV4Headers(const std::string& access_key_id, const std::string& secret_access_key, const std::string& access_token);
         void insertV2Headers(const std::string& access_key_id, const std::string& secret_access_key, const std::string& access_token);
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -436,7 +436,17 @@ static int get_object_attribute(const char* path, struct stat* pstbuf, headers_t
 
         *pObjType = objtype_t::FILE;       // Since can't access the object, set as a file.
 
-    }else if(0 != result){
+    }else if(-ENOENT == result){
+        // [NOTE]
+        // Only on a -ENOENT error should the check for a compatible object(without a trailing
+        // "/" and with the _$folder$ suffix) be performed.
+        // Performing this check after receiving any other error(-EIO, -ETIMEOUT, etc.) can lead
+        // to serious problems. For example, if the system experiences a transient error, even
+        // though the target object actually exists, a compatible object will also not exist
+        // (because the object exists), resulting in the object being considered non-existent.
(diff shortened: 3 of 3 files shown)


gaul:
## Review

**Verdict: one real bug, one behavior regression worth questioning.**

### 1. The `-EAGAIN` branch for 429/503 is dead code (bug)

```cpp
for(int retrycnt = 0; ... && retrycnt < S3fsCurl::retries; ++retrycnt){
    ...
    case 429:
    case 503:
        if(retrycnt < S3fsCurl::retries){
            // sleep + retry
        }else{
            result = -EAGAIN;   // unreachable
        }
        break;
```

Inside the loop, `retrycnt < S3fsCurl::retries` is guaranteed by the loop condition, so the `else` never runs. When retries are exhausted, `result` stays `S3FSCURL_PERFORM_RESULT_NOTSET` and falls through to the post-loop block (curl.cpp:2210) which returns `-EIO` ("### giving up"), not `-EAGAIN`. So the PR description's contract ("429/503 → -EAGAIN after retrying") is not actually delivered.

Suggested fix: change the check to `if(retrycnt + 1 < S3fsCurl::retries)` (more attempts still coming) and set `result = -EAGAIN` in the `else`.

### 2. 500 no longer retried — regression vs. prior behavior

Previously 500 was bundled with 429/503 and got exponential-backoff retry. After this PR, 500 immediately returns `-EIO` with zero retries. AWS S3's own SDKs retry 500 (classed as retryable `InternalError`). If the intent was to map the *final* errno, the retry loop should still run for 500 and only set `-EIO` after retries are exhausted.

### 3. 502/504 still don't retry

Mapping 502→`ECONNRESET` and 504→`ETIMEDOUT` is fine, but these are typically transient and benefit from retry just like 503. Pre-existing behavior was no-retry (they hit `default`), so this PR doesn't make it worse — but given the PR is about "strictly handling" the 5xx range, this seems like the moment to fix it, or at least leave a comment explaining the intentional asymmetry.

### 4. `get_object_attribute` short-circuit — worth calling out

```cpp
if(-ENOTSUP == result || -ECONNRESET == result || -EAGAIN == result || -ETIMEDOUT == result){
    return result;
}else if(-EPERM == result){ ... }
```

Reasonable — hammering the server with directory-variant HEADs (`/path/`, `_$folder$`, etc.) doesn't help if it's already 5xx-ing. But it's a user-visible change: a transient 503 on HEAD now surfaces `EAGAIN` immediately rather than potentially resolving via variant probes. The PR description's `cat: file: Resource temporarily unavailable` example suggests this is intended — worth confirming.

Minor: the `}else{  // 0 == result` comment is misleading — at that branch `result` may be any unmapped non-zero errno, not strictly 0.

### Nits
- Pure whitespace cleanups (curl.cpp:1977, 3208) are unrelated to the fix; ideally a separate commit.
- A regression test that returns 503 N+1 times and asserts the syscall sees `EAGAIN` would lock in the intended behavior of issue 1.

ggtakec:
@gaul Thank you for your feedback. I've revised the code, so please review it again.

### Handling Errors 429 and 500-504

- Error 500 has been reverted to retry as before, similar to errors 429 and 503.
- Dead code(bug) has been fixed. (This was missed during verification.)
- Errors 501, 502, and 504 have also been changed to require retries, similar to errors 429, 500, and 503.

These changes unify the retry logic across errors 429, 500, 501, 502, 503, and 504, while maintaining their respective return values.

### get_object_attribute

Previously, server or communication errors(such as -ENOTSUP, -ECONNRESET, -EAGAIN, and -ETIMEDOUT) were collapsed into a single -EIO.
This caused s3fs to continue searching for the file under alternative names.
As noted in #2581, this behavior is incorrect and poses a risk of data corruption.

In this PR, I have separated these into distinct error codes.
When a server error occurs, the existence of the target object is indeterminate.
If `get_object_attribute` continues execution and fails to find an alternative name, it incorrectly concludes the file does not exist, leading s3fs to create a new, empty file.(ex. #2581)
This is a critical issue for users.

Therefore, I believe that when a server error occurs(-ENOTSUP, -ECONNRESET, -EAGAIN, -ETIMEDOUT), `get_object_attribute` should terminate with an error without even searching for files with alternative names.

I believe this change will have little impact on user programs and commands accessing files managed by s3fs.
These error codes will be caught by the calling program, and the operation will be interrupted in the same way as when receiving -EIO.
To the user, it will appear the same as when EIO previously failed (although the error message may differ slightly).

_I tried using the `cat` command (which internally calls `open`), but it output the same message every time and terminated with an error._
_If you use a function like `open()` directly, you can get its contents using `errno`, and the command calling it will catch the error and output a message. (The message content is determined by the command.)_

The only difference is that these error case no longer searches for files with different names.
But, this should be the correct behavior for users and shouldn't cause confusion.

We need to be able to identify this error state during support, but we can do it through the debug messages.

### Additional Fixes: s3fs_truncate

During the verification of this revision, I realized that there was still a problem with `s3fs_truncate`.
Currently `s3fs_truncate` created an empty file if `get_object_attribute` returned any error.
However, I realized that, it should only create an empty file when -ENOENT is returned.
Therefore, I have made additional modifications to `s3fs_truncate` as well.

### Testing

Automated tests for these scenarios require a driver to simulate specific server errors(ex. 503), which is currently difficult to implement in the test suite.
(I have verified these cases manually using gdb.)

### Miscellaneous

- Cleaned up irrelevant spacing
- Removed comment as requested
- Changed outdated comments

gaul:
## Review (revised PR)

The data-corruption fix is correct, the dead-code bug from v1 is gone, and the layered defenses (`RequestPerform` → `get_object_attribute` → `s3fs_truncate`) are well chosen. One regression concern (501 retry) and a few smaller items below.

### What this PR gets right

- **Dead-code bug from the prior review is fixed.** `(retrycnt + 1) < S3fsCurl::retries` correctly identifies "more retries available," so the `else` branch is now reachable on the final attempt.
- **500 retry regression is fixed** — 500 is back in the retry path.
- **502/504 now retry** — pre-existing gap closed.
- **The `s3fs_truncate` change is a proper defense-in-depth fix.** Previously *any* error from `get_object_attribute` triggered an empty-file create-and-flush, which is the actual data-corruption mechanism in #2581. Restricting that branch to `-ENOENT` is correct.
- **The `get_object_attribute` short-circuit correctly avoids the dangerous alt-path probes** (`/path/`, `path_$folder$`) when the server is in an indeterminate state.

### Concerns

#### 1. 501 should not be retried (regression)

`501 Not Implemented` is **not** transient — the server is telling us "I don't support this method" and another attempt will return the same answer. Folding 501 into the retry loop means every operation against a backend that lacks a feature now waits ~60s (default `retries=5` with exponential backoff: 2+4+8+16) before surfacing `-ENOTSUP`. The original immediate-return behavior was correct.

Suggestion — pull 501 back out of the shared block:

```cpp
case 501:
    S3FS_PRN_INFO3("HTTP response code 501 was returned, returning ENOTSUP");
    S3FS_PRN_DBG("Body Text: %s", bodydata.c_str());
    result = -ENOTSUP;
    break;

case 429:
case 500:
case 502:
case 503:
case 504:
    if((retrycnt + 1) < S3fsCurl::retries){
        ...
```

#### 2. Asymmetry: 500 → `EIO`, but `EIO` is not in the short-circuit list

500 is now mapped to `-EIO`, but `get_object_attribute` does not short-circuit on `-EIO` because `EIO` is also produced by other paths (`CURLE_*` failures, response-code parsing failure, the `default:` arm). The result is that a backend consistently returning 500 still triggers alt-path probes — and if the server then flips to 200/404 between probes, the #2581 corruption window reopens.

At minimum, please document the gap in the comment on the short-circuit block — "`EIO` intentionally excluded; it is overloaded with curl/transport errors that have legitimate alt-path retry semantics." Or consider a distinct errno for "server-mapped 500" that *can* be added to the short-circuit list.

#### 3. Comment in `get_object_attribute` describes WHAT, not WHY

```cpp
if(-ENOTSUP == result || -ECONNRESET == result || -EAGAIN == result || -ETIMEDOUT == result){
    // This indicates a problem on the server side, and no further testing will
    // be performed; an error will be returned.
```

The "no further testing will be performed" is obvious from the `return`. The non-obvious WHY is the corruption hazard — please reference #2581 and explain the failure mode briefly so a future contributor doesn't "fix" this by adding a fallback search:

```cpp
// Server is in an indeterminate state. Probing alternative names
// (e.g., "<path>/", "<path>_$folder$") risks one of those probes
// hitting the server after recovery and returning 404, which would
// cause us to incorrectly conclude the file does not exist and
// truncate it on the next open. See #2581.
```

#### 4. The if/else-if cascade for errno mapping is clunky

The five-way `else if(429 == responseCode ...)` block in the `else` branch is essentially a manual switch. Either an actual `switch(responseCode)` or pre-mapping the errno at the top of each `case` reads better. Style only; not blocking.

#### 5. 502 → `ECONNRESET` is a semantic stretch

`ECONNRESET` is "connection reset by peer" (a TCP-level event); 502 is "bad gateway." It does at least convey "transient, try again," so I won't insist on a change, but a one-line comment justifying the choice would help.

#### 6. `retrycnt` (local) vs `retry_count` (member) is confusing

The loop counter `retrycnt` and the member `retry_count` happen to stay synced because `RemakeHandle()` (which increments `retry_count`) is called every loop iteration where `result` is unset. Pre-existing, not introduced here, but the new `(retrycnt + 1) < S3fsCurl::retries` check sitting next to `2 << retry_count` makes the dual-counter situation more visible. A short comment explaining the relationship would help future readers.

#### 7. Test coverage

I understand fully simulating server faults is hard, but two cheap things would still pin down the contract:

- A **stub-level unit test** that exercises the response-code switch with `responseCode = 503/500/501/...` and asserts the resulting errno. Doesn't need a real network — just a way to invoke the mapping.
- An **integration test against the existing s3proxy harness** that returns 503 to a fixed number of HEADs. With `retries=2`, you can reliably exhaust retries and check the syscall sees `EAGAIN`.

The dead-code bug in v1 of this PR is exactly the kind of regression a single test would have caught.

#### 8. User-visible behavior change

Callers grepping for "Input/output error" after 502/503/504 will now see `Resource temporarily unavailable` / `Connection reset by peer` / `Connection timed out`. Worth a NEWS entry.

### Verdict

Recommend pulling 501 out of the retry loop (concern #1) before merging — that's a clear regression. The other items are improvements that can be addressed here or in follow-ups.

ggtakec:
@gaul I have revised the code again. Please review it again.

### 1. 501 should not be retried (regression)
I've modified the code so that it doesn't retry if a 501 error occurs. (It's the same as the previous code.)

### 2. Asymmetry: 500
500 errors will continue to be treated as `-EIO` errors.
While `-ENXIO` can be used as an alternative to `-EIO`, we can keep `-EIO` because `get_object_attribute` have been modified about handling return code.
Then the behavior of `get_object_attribute` has been corrected. (See below)

### 3. About get_object_attribute
First, this function was changed the search for compatible objects to only include `-ENOENT`.
(For readability, the order of the conditional statements in the result code has been changed.)
And I also added some comments.

Regarding this change, firstly errors other than `-ENOENT` and `-EPERM` means that the object's existence could not be confirmed.
So that, if a compatibility object check is performed in this state, the same thing as in #2581 will occur.

Therefore, all errors other than `-ENOENT` and `-EPERM`(including `-EIO`) should be treated as errors in this function.

I investigated the scope of the error resulting from this modification in the call to the `get_object_attribute` function and confirmed that there are no problems.
(#2888 represents the fix for unstable code found as a result. However, this is considered separate from this issue.)

During this process, a bug in `rename_directory`(skipping error handling) was found and fixed as it is related.
(But I believe this bug occurred very infrequently because the caller actually performs a check.)

### 4. In RequestPerform function(asbou if/else-if, and retry_count)
I've modified the code to handle both of these issues.
I added a `S3fsCurl::WaitBeforeRetry()` method to centralize the sleep process.
This eliminates the need to write sleep processing in multiple places; you simply call `S3fsCurl::WaitBeforeRetry()`.
Therefore, we do not need `if/else-if` steatement for each result code  in one cace block.
Also, the `retry_count` variable does not appear within the `S3fsCurl::RequestPerform` method, so there should be no confusion.
Furthermore, I've added an explanation to the `S3fsCurl::WaitBeforeRetry()` function.

### 5. 502 mapping
In making this decision, I also considered using `-EAGAIN`, but since it was an error resulting from retries, I decided that `-EAGAIN` was not appropriate.
However, since it wasn't actually a TCP level error, `-ECONNRESET` might also be inappropriate.
Therefore, I changed it to `-EWOULDBLOCK` (but this is equivalent to `-EAGAIN`).

### 6. Test coverage
Without hardcoding, clear testing seems difficult without a stub that returns errors at arbitrary times.
Do you have any ideas? and Could that be added to test cases later?
(I made a mistake with gdb, but if automation isn't possible, you can test with gdb.)

Issue #2865, opened 2026-05-06
crash on macos


macos-fuse-t:
Immediately after compilation on macOS 26:

➜  ./s3fs
s3fs: missing BUCKET argument.
Usage: s3fs BUCKET:[PATH] MOUNTPOINT [OPTION]...
libc++abi: terminating due to uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument

```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x1854415b0 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x18547b888 pthread_kill + 296
2   libsystem_c.dylib             	       0x185380850 abort + 124
3   libc++abi.dylib               	       0x18542f858 __abort_message + 132
4   libc++abi.dylib               	       0x18541e4bc demangling_terminate_handler() + 280
5   libobjc.A.dylib               	       0x185048424 _objc_terminate() + 172
6   libc++abi.dylib               	       0x18542ec2c std::__terminate(void (*)()) + 16
7   libc++abi.dylib               	       0x185432394 __cxxabiv1::failed_throw(__cxxabiv1::__cxa_exception*) + 88
8   libc++abi.dylib               	       0x18543233c __cxa_throw + 92
9   libc++.1.dylib                	       0x1853abb90 std::__1::__throw_system_error[abi:ne200100](std::__1::error_code, char const*) + 92
10  libc++.1.dylib                	       0x1853abb34 std::__1::__throw_system_error(int, char const*) + 32
11  libc++.1.dylib                	       0x1853ad8a0 std::__1::mutex::lock() + 40
12  s3fs                          	       0x104e75a74 std::__1::lock_guard<std::__1::mutex>::lock_guard[abi:nn200100](std::__1::mutex&) + 12 (lock_guard.h:33) [inlined]
13  s3fs                          	       0x104e75a74 std::__1::lock_guard<std::__1::mutex>::lock_guard[abi:nn200100](std::__1::mutex&) + 12 (lock_guard.h:32) [inlined]
14  s3fs                          	       0x104e75a74 StatCacheNode::DecrementCacheCount(objtype_t) + 12 (cache_node.cpp:125) [inlined]
15  s3fs                          	       0x104e75a74 DirStatCache::~DirStatCache() + 80 (cache_node.cpp:891)
[...]

PR #2867, opened 2026-05-06
Changed initialization for singletons of StatCache(avoid SIOF)

ggtakec:
#2865

### Details
The error reported in #2865 on macOS is an initialization issue with the StatCache singleton object.
(see, SIOF: Static Initialization Order Fiasco)

The error in #2865 indicates that s3fs attempts to terminate because no bucket was specified during s3fs startup, but a fatal error occurs during the termination process.

This fatal error occurs during the destructuring of the StatCache singleton object, caused by a failure in the initialization of the static object declared as a singleton (StatCache).

This PR only addresses the SIOF issues confirmed at this time.
We will address any other issues that may occur in the future.
No test cases related to this issue will be added. (Implementing this as a permanent test is difficult.)

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -30,26 +30,13 @@
 //-------------------------------------------------------------------
 // Static
 //-------------------------------------------------------------------
-StatCache       StatCache::singleton;
 std::mutex      StatCache::stat_cache_lock;
 
 //-------------------------------------------------------------------
 // Constructor/Destructor
 //-------------------------------------------------------------------
-StatCache::StatCache() : pMountPointDir(nullptr), CacheSize(100'000)
+StatCache::StatCache() : pMountPointDir(std::make_shared<DirStatCache>("/")), CacheSize(100'000)
 {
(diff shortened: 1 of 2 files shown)


gaul:
s3fs might have other SIOF issues according to Claude:

> Same pattern exists elsewhere — src/fdcache.h:37 (FdManager::singleton) and src/addhead.h:64 (AdditionalHeader::singleton) use the identical file-scope static idiom and are equally exposed to SIOF. The PR explicitly defers these ("only addresses the SIOF issues confirmed at this time"), which is fine, but worth filing a follow-up — if macOS reproduced this once, it can reproduce it for the others under a slightly different startup path.

ggtakec:
@gaul
There may be some SIOF-related modifications needed in areas that are not currently apparent.
I plan to review the entire code.
I believe modifications will primarily be needed in parts defined using global (including static) variables in custom classes.(not primitive type variable)
Classes handling exclusive objects, such as pthead, likely pose risks similar to those encountered in this case.

PR #2873, opened 2026-05-09
Clean up inconsistent spacing and tab-space mix

ggtakec:
n/a

### Details
- Removed unnecessary trailing spaces.
- Fixed inconsistent indentation in several files (I noticed).


gaul:
It would be good if we could automate this somehow, e.g., clang-format.

ggtakec:
It seems like we could maintain it by detecting and correcting issues with CI for each pull request.
(I don't want to make the coding rules for s3fs too strict, so I think prohibiting tabs and indentation is about right.)

It's a minor point, but when you make corrections like this, a correction history is left that only changes the format, which makes it cumbersome when trying to trace back the history with tools like Blame.:-/

Issue #2871, opened 2026-05-07
Enable clang-tidy bugprone-throwing-static-initialization


gaul:
[bugprone-throwing-static-initialization](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/throwing-static-initialization.html) can find issues with [Static Initialization Order Fiasco](https://en.cppreference.com/cpp/language/siof) seen in #2865.  @ggtakec

ggtakec:
Thanks, I believe this `bugprone-throwing-static-initialization` is also effective.

I plan to check the code regarding SIOF.
Basically, I will examine static variables in the global file of class objects.
In the StatCache class, the destructor referenced its internal variable, and because it was uninitialized, it caused a bug. (#2865)
Such variables should be defined as static variables within functions, and access functions should be prepared.
(Static variables within functions are initialized when the function is called, so they are not affected by SIOF.)

gaul:
cppcoreguidelines-avoid-non-const-global-variables better captures the SIOF issue.

PR #2874, opened 2026-05-09
Fixed the initialization of variables that could potentially cause SIOF

ggtakec:
#2867 #2871

### Details
I have fixed a potential `Static Initialization Order Fiasco`(SIOF).

I replaced some class variables with an access function to ensure safe initialization.
Those variables were pointers to class objects, wrapped in `unique_ptr`, but `unique_ptr` is no longer used because access functions are now available.
Please note that `ThreadPoolMan::singleton` still uses a `unique_ptr`, as I wanted to maintain the current class structure.

These changes are strictly limited to the initialization process; the core logic remains unchanged.

### DIFF
--- a/src/addhead.h
+++ b/src/addhead.h
@@ -61,12 +61,11 @@ using addheadlist_t = std::vector<add_header>;
 class AdditionalHeader
 {
     private:
-        static AdditionalHeader singleton;
-        bool                    is_enable;
-        addheadlist_t           addheadlist;
+        bool            is_enable = false;
+        addheadlist_t   addheadlist;
 
     protected:
-        AdditionalHeader();
+        AdditionalHeader() = default;
         ~AdditionalHeader();
 
     public:
@@ -76,7 +75,11 @@ class AdditionalHeader
         AdditionalHeader& operator=(AdditionalHeader&&) = delete;
 
         // Reference singleton
-        static AdditionalHeader* get() { return &singleton; }
+        static AdditionalHeader* get()
+        {
+            static AdditionalHeader singleton;
+            return &singleton;
+        }
 
         bool Load(const char* file);
         void Unload();
--- a/src/addhead.cpp
+++ b/src/addhead.cpp
@@ -37,29 +37,11 @@
 static constexpr char ADD_HEAD_REGEX[] = "reg:";
 
 //-------------------------------------------------------------------
-// Class AdditionalHeader
-//-------------------------------------------------------------------
-AdditionalHeader AdditionalHeader::singleton;
-
-//-------------------------------------------------------------------
 // Class AdditionalHeader method
 //-------------------------------------------------------------------
-AdditionalHeader::AdditionalHeader()
-{
-    if(this == AdditionalHeader::get()){
-        is_enable = false;
-    }else{
-        abort();
-    }
-}
-
 AdditionalHeader::~AdditionalHeader()
 {
-    if(this == AdditionalHeader::get()){
-        Unload();
-    }else{
-        abort();
-    }
+    Unload();
 }
 
 bool AdditionalHeader::Load(const char* file)
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -77,7 +77,6 @@ static constexpr char NOCACHE_PATH_PREFIX_FORM[] = " __S3FS_UNEXISTED_PATH_%lx__
 //------------------------------------------------
 // FdManager class variable
 //------------------------------------------------
-FdManager       FdManager::singleton;
 std::mutex      FdManager::fd_manager_lock;
 std::mutex      FdManager::cache_cleanup_lock;
 std::mutex      FdManager::reserved_diskspace_lock;
@@ -435,7 +434,7 @@ bool FdManager::HasOpenEntityFd(const char* path)
 
     const FdEntity* ent;
     int         fd = -1;
-    if(nullptr == (ent = FdManager::singleton.GetFdEntityHasLock(path, fd, false))){
+    if(nullptr == (ent = FdManager::get()->GetFdEntityHasLock(path, fd, false))){
         return false;
     }
     return (0 < ent->GetOpenCount());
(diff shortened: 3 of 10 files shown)


gaul:
Suggestion for simplifying `ThreadPoolMan::Singleton`: drop the enum dispatch and the `init_singleton` bool, since the `unique_ptr`'s own truthiness already encodes "is initialized." A function-local static `unique_ptr` returned by reference gives the same SIOF safety as the Meyers singletons used elsewhere in this PR, while preserving the explicit `Initialize`/`Destroy` lifetime that `ThreadPoolMan` needs (so workers are joined before `main` returns, not at static teardown).

### `threadpoolman.h`
```cpp
class ThreadPoolMan
{
    private:
        static int worker_count;

        std::atomic<bool>     is_exit;
        Semaphore             thpoolman_sem;
        std::mutex            thread_list_lock;
        std::vector<std::pair<std::thread, std::future<int>>> thread_list GUARDED_BY(thread_list_lock);
        thpoolman_params_t    instruction_list GUARDED_BY(thread_list_lock);

    private:
        // Owning slot for the singleton. Function-local static => no SIOF.
        // Returned by reference so Initialize/Destroy can mutate it.
        static std::unique_ptr<ThreadPoolMan>& Slot()
        {
            static std::unique_ptr<ThreadPoolMan> slot;
            return slot;
        }

        static void Worker(ThreadPoolMan* psingleton, std::promise<int> promise);
        // ... unchanged ...

    public:
        // ... unchanged ctors/dtors ...

        static bool Initialize(int count = -1);
        static void Destroy();
        static bool IsInitialized() { return static_cast<bool>(Slot()); }
        // ... unchanged ...
};
```

### `threadpoolman.cpp`
```cpp
int ThreadPoolMan::worker_count = 10;

bool ThreadPoolMan::Initialize(int count)
{
    auto& slot = Slot();
    if(slot){
        S3FS_PRN_CRIT("Already singleton for Thread Manager exists.");
        return false;
    }
    if(-1 != count){
        SetWorkerCount(count);
    }
    slot = std::make_unique<ThreadPoolMan>(worker_count);
    return true;
}

void ThreadPoolMan::Destroy()
{
    Slot().reset();
}

bool ThreadPoolMan::Instruct(const thpoolman_param& param)
{
    auto& slot = Slot();
    if(!slot){
        S3FS_PRN_WARN("The singleton object is not initialized yet.");
        return false;
    }
    if(!param.psem){
        S3FS_PRN_ERR("Thread parameter Semaphore is null.");
        return false;
    }
    slot->SetInstruction(param);
    return true;
}

bool ThreadPoolMan::AwaitInstruct(const thpoolman_param& param)
{
    auto& slot = Slot();
    if(!slot){
        S3FS_PRN_WARN("The singleton object is not initialized yet.");
        return false;
    }
    Semaphore await_sem(0);
    thpoolman_param local_param;
    local_param.args  = param.args;
    local_param.psem  = &await_sem;
    local_param.pfunc = param.pfunc;
    slot->SetInstruction(local_param);
    await_sem.acquire();
    return true;
}
```

The constructor's self-check `if(ThreadPoolMan::init_singleton) abort();` can be replaced with `if(Slot()) abort();` if you want to keep the belt-and-braces guard.

### Why this is cleaner
- **One source of truth** — the `unique_ptr`'s truthiness replaces `init_singleton`.
- **No enum-dispatched function** — three call sites (`Initialize`, `Destroy`, accessors) read straight, no `if(mode == ...)` ladder.
- **Same SIOF safety** — `Slot()`'s function-local static is constructed on first call; the `unique_ptr` is empty until `Initialize()` runs, so destruction of the contained `ThreadPoolMan` is still controlled by the explicit `Destroy()` call from `s3fs_destroy`. The empty `unique_ptr` left over for static teardown is trivially destructible.
- **Symmetrical with the rest of the PR** — same Meyers-singleton pattern as `FdManager::get()` etc., just owning a `unique_ptr<T>` instead of a `T` because the lifetime is bounded by FUSE init/destroy rather than by program lifetime.

ggtakec:
@gaul Thanks for your review.
I fixed codes, please review new code.

gaul (changes requested):
### Critical regression: `S3fsSignals` singleton is never constructed

This PR removes `S3fsSignals::Initialize()` from `s3fs_init`, removes `S3fsSignals::Destroy()` from `s3fs_destroy`, and deletes both static methods. Combined with `S3fsSignals::get()` remaining `protected`, nothing now triggers construction of the singleton.

Walk-through:
- `S3fsSignals::get()` is `protected` — only the class itself can call it.
- The only internal caller is `HandlerUSR1`, which only fires when the signal handler has already been registered.
- The signal handler is registered inside the `S3fsSignals` constructor.
- The only external entry point, `S3fsSignals::SetUsr1Handler`, just sets `enableUsr1 = true` without touching `get()`.

Net effect: the constructor never runs, so SIGUSR1 (cache check), **SIGUSR2 (log level bump)**, and **SIGHUP (reopen log file)** handlers are never installed, and the USR1 worker thread is never spawned. `set_check_cache_sigusr1` and runtime log control are silently broken. No tests catch this — there are no signal-handler tests in `test/`.

This also contradicts the PR description's claim that "changes are strictly limited to the initialization process; the core logic remains unchanged." Removing the signal-lifecycle calls from FUSE init/destroy is a behavior change.

**Suggested fix:** keep an explicit `Initialize()` that forces construction:
```cpp
// sighandlers.h
public:
    static bool Initialize() { (void)get(); return true; }
```
…and restore the call in `s3fs_init`. If `Destroy()` is also dropped, note that the USR1 worker thread will be joined at static teardown rather than in `s3fs_destroy`. The worker calls `FdManager::get()->CheckAllCache()`, so `FdManager` must outlive `S3fsSignals` at teardown. That happens to hold today (FdManager is touched first via the cache-dir cleanup in `s3fs_init`), but the invariant is implicit and fragile — keeping an explicit `Destroy()` in `s3fs_destroy` to join the thread early is safer.

### Minor: dead `abort()` guard in `ThreadPoolMan` constructor

`ThreadPoolMan::ThreadPoolMan` still has:

```cpp
if(ThreadPoolMan::Slot()){
    S3FS_PRN_CRIT("Already singleton for Thread Manager exists.");
    abort();
}
```

But `Initialize()` already returns `false` in that case, so this is unreachable from any current caller. Either delete it, or keep purely as defense for direct `new ThreadPoolMan(...)`.

### Nit: `GetS3fsCred()` inconsistent with other singletons

The other singletons expose a static class method (`FdManager::get()`, `AdditionalHeader::get()`, `StatCache::getStatCacheData()`). The new free function `GetS3fsCred()` in `s3fs.cpp` is the odd one out — moving it onto `S3fsCred` as `S3fsCred::get()` would match the established pattern. Purely cosmetic.

### Positive

- `ThreadPoolMan::Initialize()` returning `false` on duplicate instead of `abort()` is a nice upgrade — the caller now does `s3fs_exit_fuseloop(EXIT_FAILURE)` gracefully.
- The `Slot()`/Meyers pattern matches the existing `StatCache::getStatCacheData()` style cleanly.
- Builds clean on macOS.

src/addhead.h
```diff
@@ -61,9 +61,8 @@ using addheadlist_t = std::vector<add_header>;
 class AdditionalHeader
 {
     private:
-        static AdditionalHeader singleton;
-        bool                    is_enable;
-        addheadlist_t           addheadlist;
+        bool            is_enable;
```
gaul: In C++11 we can just have `bool is_enable = false` to avoid the initialization list in the constructor.

ggtakec: Thanks, all fixed.

ggtakec:
@gaul 
Thank you for the review. All changes have been applied. Please re-review when you have a chance.

- Fixed the missing initialization of S3fsSignals.
- Removed the unnecessary abort() call from the ThreadPoolMan constructor.
- Reimplemented GetS3fsCred() as S3fsCred::get().

PR #2877, opened 2026-05-10
Return early from s3fs_write when underlying write fails

gaul:
Previously a failed ent->Write() only logged a warning and continued, bumping mtime/ctime and possibly triggering RowFlush, which would upload the partial state and update the stat cache as if the write had succeeded.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -3158,6 +3158,7 @@ static int s3fs_write(const char* _path, const char* buf, size_t size, off_t off
 
     if(0 > (res = ent->Write(static_cast<int>(fi->fh), buf, offset, size))){
         S3FS_PRN_WARN("failed to write file(%s). result=%zd", path, res);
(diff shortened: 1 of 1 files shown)


src/s3fs.cpp
```diff
@@ -3157,6 +3157,7 @@ static int s3fs_write(const char* _path, const char* buf, size_t size, off_t off
 
     if(0 > (res = ent->Write(static_cast<int>(fi->fh), buf, offset, size))){
         S3FS_PRN_WARN("failed to write file(%s). result=%zd", path, res);
```
ggtakec, line 3160: If the program stops here due to an error, it is better to call S3FS_PRN_ERR.

ggtakec: I think S3FS_PRN_ERR would be better, but I'll merge this first.

ggtakec (approved):

PR #2878, opened 2026-05-10
Fix infinite loop in DirStatCache::TruncateCacheHasLock

gaul:
The for-loop has no increment expression and relies on each branch to advance the iterator. The early-continue path when GetChildLeafNameHasLock returns false skipped the advancement, causing an infinite loop on any malformed child entry whose path is shorter than this directory's path or whose leaf name is empty.


ggtakec (approved):
Thanks!

PR #2880, opened 2026-05-10
Replace fixed sleeps in test startup with short polls

gaul:
Three integration-test startup paths each waited 1s per iteration:
  - unconditional sleep after launching s3fs in the background, used to give the subshell time to write its pid file
  - macOS mount-poll loop waiting for the s3fs mount to appear in df
  - wait_for_port polling for S3Proxy's listener

Replace the unconditional sleep with a poll on the pid file and shrink the per-iteration sleep on the two existing poll loops from 1s to 0.1s. All three loops now use a 0.1s tick. Total wait budgets are unchanged (5s, 20s, 30s), but the loops exit promptly once the resource is ready.

Local benchmark on macos-fuse-t with S3Proxy, single-test mode: ~3.13s -> ~1.53s per run (~51% reduction); ~17s saved across the ALL_TESTS=1 11-flag matrix.

### DIFF
--- a/test/integration-test-common.sh
+++ b/test/integration-test-common.sh
@@ -129,7 +129,7 @@ else
 fi
 
 # This function execute the function parameters $1 times
-# before giving up, with 1 second delays.
+# before giving up, with 0.1 second delays.
 function retry {
     local N="$1"
     shift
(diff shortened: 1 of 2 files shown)


ggtakec (approved):
That's good.
I overlooked the fact that `sleep` can use decimal points.

PR #2882, opened 2026-05-11
Re-apply S3Proxy 3.1.0 upgrade

gaul:
Also set Content-Type for s3_cp test helper curl's --data-binary defaults Content-Type to application/x-www-form-urlencoded, which causes Jetty 12 to drain the request body as form parameters before S3Proxy 3.1.0 can hash it for x-amz-content-sha256 verification, failing the check with 400 XAmzContentSHA256Mismatch. Send application/octet-stream instead.

Release notes:
https://github.com/gaul/s3proxy/releases/tag/s3proxy-3.1.0


gaul:
I don't think that S3Proxy 3.1.0 is responsible for macOS CI hangs but rather upgrading triggers latent issues.  I will keep experimenting.

ggtakec:
@gaul I agree. It seems the real issue is independent of the s3proxy upgrade, and the timing of the increased frequency was purely coincidental. Since we haven't identified the cause yet, let's investigate further.

ggtakec (approved):

PR #2886, opened 2026-05-12
Add fastreaddir to skip per-object HEAD in readdir/getattr (#1482)

gaul:
Reduces RPC count on large directories by populating the stat cache directly from ListObjects results rather than issuing one HEAD per entry. On a getattr cache miss, a single HEAD is issued; if it 404s, retry once with a trailing slash to find directory markers stored as "foo/". The retry only adopts the slash form when the response looks like a directory marker (size 0 or Content-Type=application/x-directory) to avoid false positives on backends that return 200 for HEAD "foo/" when only "foo" exists. The _$folder$ probe and directory_empty LIST fallback remain disabled.

Synthesized stat uses Content-Length / Last-Modified plus 0750/0640 mode defaults and global uid/gid. x-amz-meta-* headers (mode, owner, times, xattr) are discarded both on the read-miss path and on every cache write via a new StatCache::Sanitizer hook, so writes from chmod/chown/utimens/setxattr cannot temporarily surface stale metadata inside the TTL window. Trade POSIX metadata fidelity for substantially fewer S3 requests on large directories.

Skip the metadata-roundtrip integration tests (chmod, chown, utimens, xattr, ACL, mtime) when fastreaddir is set; those features are inert by design in this mode.


ggtakec:
Since this is still a draft, I just wanted to ask two quick questions after looking through the code:

1. Regarding the `Sanitizer` ( `sanitize_for_fastreaddir` ), is there a specific future expansion planned that led you to implement it as a function pointer instead of a `StatCache` method?

2. While the object types are currently only `objtype_t::DIR_NORMAL` and `objtype_t::FILE`, it seems like we will need some verification for cases when other object types are passed.

PR #2895, opened 2026-06-06
Return std::optional from accessors instead of bool out-parameter

gaul:
Bump the C++ standard to C++17 and convert single-value "bool f(..., T& out)" accessors to return std::optional<T>. This removes uninitialized out-parameter hazards and lets callers use the if(auto v = f(...)){ ... *v ... } idiom.

Converted: get_keyword_value, convert_unixtime_from_option_arg, get_unixtime_from_iso8601, StatCache::GetSymlink, StatCacheNode::GetExtra, FdEntity::GetXattr, FdEntity::GetSize, PseudoFdInfo::GetUploadId, S3ObjList::GetLastName, simple_parse_xml,
S3fsCred::ParseIAMRoleFromMetaDataResponse, S3fsCurl::GetCurlErrorString and S3fsCurl::GetIAMRoleFromMetaData.

C++17 activates clang-tidy's modernize-use-scoped-lock against the existing std::lock_guard usage, so disable that check to keep the lint gate green.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -651,15 +651,15 @@ bool S3fsCurl::PushbackSseKeys(const std::string& input)
     }
 
     // make MD5
-    std::string strMd5;
-    if(!make_md5_from_binary(raw_key.c_str(), raw_key.length(), strMd5)){
+    auto strMd5 = make_md5_from_binary(raw_key.c_str(), raw_key.length());
+    if(!strMd5){
         S3FS_PRN_ERR("Could not make MD5 from SSE-C keys(%s).", mask_sensitive_string(raw_key.c_str()));
         return false;
     }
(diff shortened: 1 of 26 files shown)


gaul:
Depends on #2892.

src/curl.cpp
```diff
@@ -2021,16 +2014,15 @@ int S3fsCurl::RequestPerform(bool dontAddAuthHeaders /*=false*/)
 
                 {
```
ggtakec, line 2015: Wouldn't this scope become unnecessary?

ggtakec: I will remove the redundant scope definition (the outer `{...}`) later. For now, I'll merge this.

ggtakec (approved):

PR #2899, opened 2026-06-14
Inline IsS3fsLogLevel and make Level ordinal to shrink .text

gaul:
Two related fixes to the log-level guard that fronts every S3FS_PRN_* call site (~1,000 of them):

1. IsS3fsLogLevel was defined out-of-line in s3fs_logger.cpp, so without LTO every guard compiled to a function call -- even though its siblings (GetSyslogLevel, IsSetLogFile, the IsS3fsLog* helpers) are all inline in the header. Move it into the header. Each guard now folds to a direct test of debug_level instead of a call: -1,602 bytes .text, ~1,055 calls removed.

2. Level used a nested-bitmask encoding {CRIT=0,ERR=1,WARN=3,INFO=7,DBG=15} with a subset test, level == (debug_level & level). Since debug_level is only ever assigned a canonical level (SetLogLevel from the dbglevel= string, or BumpupLogLevel cycling through them) and levels are never OR'd or ordered-compared anywhere, that subset test is exactly equivalent to an ordinal threshold. Switch to {0,1,2,3,4} with level <= debug_level and drop the now-defunct "& DBG" masking in GetSyslogLevel/GetLevelString. Every inlined guard collapses from movzbl+and+cmp to a single cmpb (and frees a scratch register): -5,061 bytes more.

Total -6,663 bytes .text (-0.92%). No behavioral change: the threshold and the [CRT]/[ERR]/[WAN]/[INF]/[DBG] tags were verified at runtime across dbglevel=info/dbg, the syslog mapping stays monotonic, and CRIT remains the always-logged floor. Unit tests pass.

### DIFF
--- a/src/s3fs_logger.h
+++ b/src/s3fs_logger.h
@@ -41,12 +41,15 @@
 class S3fsLog
 {
     public:
+        // Ordinal severity, ascending verbosity: a message at `level` is shown
+        // when level <= debug_level (see IsS3fsLogLevel). CRIT is the floor and
+        // is therefore always logged.
         enum class Level : uint8_t {
             CRIT = 0,          // LEVEL_CRIT
             ERR  = 1,          // LEVEL_ERR
-            WARN = 3,          // LEVEL_WARNING
-            INFO = 7,          // LEVEL_INFO
-            DBG  = 15          // LEVEL_DEBUG
+            WARN = 2,          // LEVEL_WARNING
+            INFO = 3,          // LEVEL_INFO
+            DBG  = 4           // LEVEL_DEBUG
         };
 
     protected:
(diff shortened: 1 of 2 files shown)


src/s3fs_logger.h
```diff
@@ -41,12 +41,15 @@
 class S3fsLog
 {
     public:
+        // Ordinal severity, ascending verbosity: a message at `level` is shown
+        // when level <= debug_level (see IsS3fsLogLevel). CRIT is the floor and
+        // is therefore always logged.
         enum class Level : uint8_t {
             CRIT = 0,          // LEVEL_CRIT
             ERR  = 1,          // LEVEL_ERR
```
ggtakec, line 49: If changing to sequential numbering, shouldn't we omit the values ​​after ERR?

ggtakec: The result is the same whether this is specified or omitted, so I will merge it.

ggtakec (approved):

PR #2901, opened 2026-06-14
Support heterogeneous lookup for std::map

gaul:
This avoids temporary constructors during lookup.


ggtakec (approved):
It might be good to have `operator()(const char* a, const char* b)` as well, but there are currently no calls to it, so it's not a problem.

PR #2902, opened 2026-06-16
Avoid crash on empty parent access path

### LINKED ISSUE #2903: s3fs_flush can crash when FUSE passes a null path
## Summary

`s3fs_flush()` can crash with `SIGSEGV` when the FUSE callback receives a null path. The observed crash happened in `check_parent_object_access()` after `s3fs_flush()` passed the null path into the parent permission check.

## Environment

```text
Amazon Simple Storage Service File System V1.94 (commit:70a30d6) with OpenSSL
```

The same code path is still present in the current v1.97/master implementation.

## Observed Stack

```text
Program terminated with signal SIGSEGV, Segmentation fault.
#0  check_parent_object_access(path=0x0, mask=1) at s3fs.cpp:859
#1  s3fs_flush(_path=<optimized out>, fi=0x...) at s3fs.cpp:3007
```

## Root Cause

`check_parent_object_access()` assumes `path` is always non-null and immediately evaluates:

```cpp
strcmp(path, "/")
```

When `s3fs_flush()` passes `path == nullptr`, this becomes a null pointer dereference.

## Expected Behavior

A malformed or unexpected null/empty path should return an error such as `-EINVAL` instead of terminating the s3fs process.

## Related PR

A minimal fix is proposed in #2902.

jiangshantao-dbg:
Fixes #2903

## Summary
- return `-EINVAL` when `check_parent_object_access()` receives a null or empty path
- prevent `strcmp(path, ...)` from dereferencing a null path passed through `s3fs_flush()`

## Testing
- `git diff --check`
- full build not run locally: autotools/configure generation is unavailable in this environment (`autoupdate: command not found`)

## Notes
This is intentionally scoped as a minimal crash fix based on the v1.97 code path.


src/s3fs.cpp
```diff
@@ -755,8 +755,10 @@ static int check_parent_object_access(const char* path, int mask)
     std::string parent;
     int result;
 
+    if(!path || '\0' == path[0]){
+        return -EINVAL;
```
gaul, line 759: We check a similar condition in `get_object_attribute`, `create_directory_object`, and `get_meta_xattr_value`.  The return values differ: `-ENOENT`, `-EINVAL`, and `false`.  I'm not sure which of the two is the best choice.  `man 2 stat` suggests:

* EINVAL (fstatat()) Invalid flag specified in flags.
* ENOENT A component of pathname does not exist or is a dangling symbolic link.
* ENOENT pathname is an empty string and AT_EMPTY_PATH was not specified in flags.

The third bullet seems the most plausible to me.  @ggtakec what do you think?

ggtakec: I posted #2903 by mistake, so I have closed it.

Rather than modifying this utility function, I think it would be better overall to check the `path` variable when it is called from the FUSE handler function.

@jiangshantao-dbg
If possible, could you please revise this PR using #2903 as a reference?

ggtakec:
I have created a new PR#2907 that encompasses this fix. In that PR, the arguments for each `s3fs_~` handler are safely converted to `std::string` to ensure robust and secure handling. Additionally, the arguments of the `check_parent_object_access` function revised in this PR have also been changed to `std::string` to permanently prevent this issue from occurring.

@gaul 
Regarding the error code, the `s3fs_~` handlers will now detect invalid parameters. I believe returning `EINVAL` is the most appropriate approach here. One of the reasons is that it allows us to use a consistent error code for invalid parameter detection across all handlers.

@jiangshantao-dbg @gaul 
Personally, I would like to move forward with the new PR#2907 to ensure s3fs is comprehensively hardened across the entire codebase. With that in mind, would it be okay to close this PR and switch over to reviewing the new one?

src/s3fs.cpp
ggtakec: Please see [comment](https://github.com/s3fs-fuse/s3fs-fuse/pull/2902#issuecomment-4890458338) about this PR.

jiangshantao-dbg:
> I have created a new PR#2907 that encompasses this fix. In that PR, the arguments for each `s3fs_~` handler are safely converted to `std::string` to ensure robust and secure handling. Additionally, the arguments of the `check_parent_object_access` function revised in this PR have also been changed to `std::string` to permanently prevent this issue from occurring.
> 
> @gaul Regarding the error code, the `s3fs_~` handlers will now detect invalid parameters. I believe returning `EINVAL` is the most appropriate approach here. One of the reasons is that it allows us to use a consistent error code for invalid parameter detection across all handlers.
> 
> @jiangshantao-dbg @gaul Personally, I would like to move forward with the new PR#2907 to ensure s3fs is comprehensively hardened across the entire codebase. With that in mind, would it be okay to close this PR and switch over to reviewing the new one?

OK, i will close this PR, thanks.

PR #2904, opened 2026-06-29
Check for the handler function argument more robust

ggtakec:
### Relevant Issue (if applicable)
#2903

### Details
This PR is strengthening the validation of the `path` variable in each handler function.
This is an interim measure in response to the report in #2903.
The checks have been enhanced to prevent critical segmentation faults.

PR #2907, opened 2026-07-06
Use std::string converted from const char* argument in handler

ggtakec:
#2902 #2903 #2904

### Details
As a response to #2903, #2902 and #2904 were previously proposed. This PR extends those approaches and builds upon them.

The root cause of why the `path` argument became null in #2903 remains unknown; it might have been passed as null directly from FUSE. To eliminate this potential issue, I have updated the code to pre-check all handler function arguments (e.g., `const char* path`) and convert them to `std::string`. 

Additionally, the arguments of the utility functions defined in `s3fs.cpp` (such as the `check_parent_object_access` function proposed in #2902) have been changed from `const char*` to `const std::string&`. This completely mitigates the risk raised in #2903 and improves the overall robustness of the codebase.

---

#### Changes & Trade-offs
* **Macro Replacement:** The `WTF8_ENCODE` macro has been deprecated and replaced with the `CVT_SRTPTR_TO_STRING` macro.
* **Performance Impact:** Previously, `WTF8_ENCODE` was designed to handle `const char*` directly to operate at minimum cost when `use_wtf8=0` (as discussed in past PRs). With this new change, converting `const char*` to `std::string` introduces a slight overhead. However, I believe this performance cost is well within the acceptable range for the stability gained.

#### Additional Bug Fixes
I have also introduced `common_rmdir` and `common_unlink` to resolve a nesting bug. Previously, the `rename` process called `s3fs_rmdir` and `s3fs_unlink` directly, which caused the `WTF8_ENCODE` macro to be triggered redundant times. To fix this, I extracted the core logic into these new `common_~` functions to prevent duplicate WTF-8 re-conversions. 
*(While this fix could technically be isolated into a separate PR, I decided to include it here as it is highly relevant to the `std::string` refactoring.)*

#### Next Steps
Following this PR, I plan to sequentially refactor other functions that handle paths (such as `fdcache`) to use `std::string` as well.

### DIFF
--- a/src/curl_util.h
+++ b/src/curl_util.h
@@ -38,8 +38,8 @@ std::string get_canonical_headers(const struct curl_slist* list, bool only_amz =
 std::string get_header_value(const struct curl_slist* list, const std::string &key);
 bool MakeUrlResource(const char* realpath, std::string& resourcepath, std::string& url);
 std::string prepare_url(const char* url);
-bool get_object_sse_type(const char* path, sse_type_t& ssetype, std::string& ssevalue);   // implement in s3fs.cpp
-int put_headers(const char* path, const headers_t& meta, bool is_copy, bool use_st_size = true);    // implement in s3fs.cpp
+bool get_object_sse_type(const std::string& path, sse_type_t& ssetype, std::string& ssevalue);          // implement in s3fs.cpp
+int put_headers(const std::string& path, const headers_t& meta, bool is_copy, bool use_st_size = true); // implement in s3fs.cpp
 
 bool make_md5_from_binary(const char* pstr, size_t length, std::string& md5);
 std::string url_to_host(const std::string &url);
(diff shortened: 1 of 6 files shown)


gaul:
I'm not sure this is the correct fix.  FUSE 3 has a behavior change where it can give a NULL file name for a delete file with `-o hard_remove`.  So I think we need to use `fi->fh` instead.

ggtakec:
Thanks for sharing information about the `hard_remove` option.
Since the `hard_remove` option results in a `null` path, the approach in #2908 is preferable to the one in this PR; therefore, I am closing this PR.

Note that in s3fs, even if a file is deleted while open, the actual file data remains intact due to the mirror file (or temp file), so a crash is unlikely to occur.
However, it does not return an `ESTALE` error.
In any case, since this does not align with FUSE3 behavior, #2908 is the recommended approach over this PR.

I will also bundle the task of replacing `const char* path` with `std::string` into a separate PR.

PR #2909, opened 2026-07-14
Fix inverted HTTP proxy scheme check in SetProxy

gaul:
Commit eba1bf2 converted the previous find() != 0 comparison to is_prefix() without negating it, so proxy_http became false exactly when the proxy URL uses the http:// scheme.  As a result CURLOPT_PROXYUSERPWD was skipped for HTTP proxies (credentials from proxy_cred_file were silently dropped, causing 407 responses) and applied to non-HTTP proxies instead.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -979,7 +979,7 @@ bool S3fsCurl::SetProxy(const char* url)
         pos += strlen("://");
 
         // Check if it is other than "http://"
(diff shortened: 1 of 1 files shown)


ggtakec (approved):

ggtakec (approved):
Sorry, I missed that too.

PR #2911, opened 2026-07-14
Fix rename target in RowFlushStreamMultipart

gaul:
The stream upload flush ignored its tpath parameter and always uploaded to the entity's current path, unlike the other RowFlush variants which use tpath ? tpath : path.  rename_object_nocopy() calls RowFlush(fd, to, true) expecting the data to be uploaded to the destination key, so with streamupload and nocopyapi/norenameapi a rename uploaded the contents back to the source key, which s3fs_unlink() then deleted: the destination was never created and the file contents were lost.

Upload to tpath when it is specified.  If a stream upload is already in flight, its parts belong to a multipart upload for the current path, which cannot be completed under tpath; finish it for the current path first and then upload the whole file to tpath.  Also do not use copy multipart upload when retargeting, since the copy source would be the not-yet-existing target object.

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -1154,8 +1154,10 @@ int FdEntity::NoCacheLoadAndPost(PseudoFdInfo* pseudo_obj, off_t start, off_t si
 
 // [NOTE]
 // If the request is successful, initialize upload_id.
+// If tpath is specified, the multipart upload is started for tpath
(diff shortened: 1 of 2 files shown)


ggtakec (approved):
Thank you.
I have reviewed the issue regarding `rename_object_nocopy`, and 
this fix is ​​correct.

PR #2912, opened 2026-07-15
Rebuild untreated list when a stream upload flush fails

gaul:
The stream upload flush error paths cleared untreated_list and the upload info but left the modified pages dirty.  A subsequent flush -- s3fs_release retries Flush() while IsModified() -- derives the areas to upload only from the untreated list and the upload list(see ExtractUploadPartsFromAllArea), so it found nothing to upload, copied the stale object over itself via copy multipart upload, marked the modified pages clean, and returned success: the written data was silently lost after a single transient upload failure.

On failure, abort the in-flight multipart upload(its parts are discarded and were partly removed from the untreated list already) and rebuild the untreated list from the modified pages so that a retried flush uploads all locally modified data again.  Also stop clearing the untreated list at the end of the flush when the result is an error.

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -1717,6 +1717,43 @@ int FdEntity::RowFlushMixMultipart(PseudoFdInfo* pseudo_obj, const char* tpath)
     return result;
 }
 
+// [NOTE]
+// This method is called when a stream upload fails.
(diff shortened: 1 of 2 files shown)


ggtakec (approved):
The changes look fine.
Could you perform a rebase to resolve the file conflicts?

ggtakec (approved):

PR #2915, opened 2026-07-17
Skip sudo mdutil when Spotlight indexing is already disabled

gaul:
This allows make check to run non-interactively on macOS, where sudo cannot prompt for a password.  mdutil -a -i off only needs to run when some volume still has indexing enabled.


ggtakec (approved):
Thanks,

PR #2916, opened 2026-07-17
Remove HEAD recheck for backends conflating "dir" with "dir/"

gaul:
get_object_attribute() re-probed "dir/" whenever a HEAD request for "dir" (no trailing slash) returned a directory-format response, because some backends (s3proxy nio2 providers before 3.3.0) incorrectly answered 200 for "dir" when only the "dir/" object existed.  Real S3 treats the two keys as distinct, so on AWS the recheck always returned 404 and cost one extra request for every cache-miss stat of an old style "dir" object.

s3proxy fixed the conflation in 3.3.0 (gaul/s3proxy@f461ef75), which the integration tests already use, so drop the recheck.  A backend that still conflates the keys now classifies such a path as DIR_NOT_TERMINATE_SLASH, which s3fs already handles gracefully as an old style (pre-1.63) directory object.

Also add an integration test covering no-slash directory objects.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -477,35 +477,19 @@ static int get_object_attribute(const char* path, struct stat* pstbuf, headers_t
         }else if(is_symlink_fmt(*pmeta)){
             *pObjType = objtype_t::SYMLINK;
         }else if(is_dir_fmt(*pmeta)){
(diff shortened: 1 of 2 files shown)


ggtakec (approved):
Thanks, this will result in a slight decrease in the number of requests.

PR #2917, opened 2026-07-20
Apply miscellaneous C++17 cleanups

gaul:
* Replace &str[0] with non-const std::string::data and remove the C++14 workaround comments and NOLINT suppressions
* Remove out-of-line definitions of static constexpr members, which are implicitly inline in C++17
* Use inline constexpr for header-scope constants
* Convert MAX_STAT_CACHE_COUNTER from #define to inline constexpr
* Replace last NULL with nullptr
* Update COMPILATION.md C++17 compiler requirements


ggtakec (approved):
Thanks, it looks good.

PR #2918, opened 2026-07-20
Fix inverted AddStat check in readdir compat dir path

gaul:
Commit 5a2a7ca dropped the negation when rewriting this block, so a successful stat cache addition logged a spurious error and filled the readdir entry with an empty stat, while a failed addition silently took the success path.  Every other AddStat caller tests the negated result.


ggtakec (approved):
Thanks for fixing this which is my fault.

PR #2921, opened 2026-07-20
Make test_ensurespace_move_file exercise cache eviction

gaul:
The disk fill loop compared free space in MB from df against BIG_FILE_LENGTH in bytes, so it never copied a single filler file and the test was a plain mv with a misleading name.  Fixing the units alone would not help: the loop read df once outside the loop and would never terminate, and genuinely filling the disk would also starve s3proxy's own object store on the same file system.

Rebuild the test on the fake_diskfree budget instead, and move it to the configuration that has one: cache a closed 25MB file, consume the budget with a filler so less than the moved file size remains, and mv a 25MB file onto the mount.  ReserveDiskSpace now genuinely fails during the upload and must evict the unused cache file via CleanupCacheDir to proceed, which the test asserts, along with re-reading the evicted file. Previously the gate excluded the fake_diskfree configurations, i.e. the only ones where the low-disk condition can be simulated safely.

### DIFF
--- a/test/integration-test-common.sh
+++ b/test/integration-test-common.sh
@@ -321,6 +321,17 @@ function start_s3fs {
     #     Keep s3fs in foreground instead of daemonizing
     #
 
+    # [NOTE]
+    # Record the free space(MB) of the cache file system just before
+    # mounting: s3fs snapshots the same value at startup to compute the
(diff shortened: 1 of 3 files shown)


ggtakec (approved):
I realized there was an issue with the original test code.
Thank you.

PR #2922, opened 2026-07-20
Exercise max_dirty_data mid-write flush in tests

gaul:
The forced flush in s3fs_write when dirty data reaches max_dirty_data never ran in the test suite: the default threshold is 5GB and no test config lowered it.  Set max_dirty_data=50 (its minimum) in the use_cache test configuration and add a test writing 64MB through one descriptor, which now flushes a complete multipart upload mid-write and then performs a second multipart upload at close that combines copying the already-uploaded head with uploading the modified tail.  The existing 64MB test_truncate_shrink_file also crosses the threshold in this configuration and gains the same mid-write flush coverage.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -861,6 +861,25 @@ function test_multipart_mix {
     fi
 }
 
+function test_max_dirty_data_flush {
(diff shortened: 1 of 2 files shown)


ggtakec (approved):
Thanks for adding this new test case.

PR #2923, opened 2026-07-20
Fix data corruption in multipart upload without cache

gaul:
FdEntity::NoCacheLoadAndPost had three defects that corrupted scattered writes whenever an upload engine ran out of local disk space and switched to uploading without cache:

- It ignored its size argument inside the upload loop and uploaded to the end of whatever page spanned the requested area (usually EOF), while the page bookkeeping afterwards assumed it stopped at the requested size.  Later writes then landed behind the already uploaded position and the flush appended their parts out of order: CompleteMultipartUpload failed with EIO, the object kept its stale content, and the multipart upload leaked.
- The flush drain uploaded only the most recently updated untreated part, silently dropping any other data written after the switch.
- Parts were split on page boundaries, so a small loaded page between unloaded pages became its own part far below the 5MB S3 minimum, which only works against permissive backends.

Rework NoCacheLoadAndPost to upload exactly the requested range in multipart size chunks independent of page boundaries: a chunk that is not fully loaded is assembled in the temporary file from the loaded areas of the cache file, downloads of the unloaded areas, and zeros beyond the original size.  The flush now uploads everything between the already uploaded position (new PseudoFdInfo::GetNextUploadPos) and EOF before completing, mid-write part uploads only run when they directly follow the uploaded position, and a write before that position fails with ENOSPC instead of corrupting the upload, since its bytes can no longer be patched into the ordered parts.  Uploaded bytes are released by shrinking the file when nothing after them holds data, or by punching a hole otherwise, instead of an unconditional truncation that destroyed not yet uploaded bytes.

Add test_diskfull_fallback_upload reproducing the corruption: modify 16 bytes every 4MB of an externally created 26MB object through one descriptor while a filler file consumes the fake_diskfree budget.  The sub-5MB gaps between the writes cannot become server-side copy parts, so the write path must download them, runs out of space, and switches to the upload without cache mid-write; the flush at close then uploads the remainder.  Before this fix the test failed with EIO at close and stale object content.

### DIFF
--- a/src/fdcache_entity.cpp
+++ b/src/fdcache_entity.cpp
@@ -88,6 +88,34 @@ int FdEntity::FillFile(int fd, unsigned char byte, off_t size, off_t start)
 }
 
 // [NOTE]
+// Copies the data in the range [start, start + size) from srcfd to dstfd
+// at the same offsets.  If srcfd ends before the range does, the rest is
+// skipped(the caller prepares the destination so that it reads as zeros).
+//
+int FdEntity::CopyFdBytes(int srcfd, int dstfd, off_t start, off_t size)
+{
+    char bytes[1024 * 32];         // 32kb
+
+    for(off_t total = 0, onread = 0; total < size; total += onread){
+        if(-1 == (onread = pread(srcfd, bytes, std::min(static_cast<off_t>(sizeof(bytes)), size - total), start + total))){
+            S3FS_PRN_ERR("pread failed. errno(%d)", errno);
+            return -errno;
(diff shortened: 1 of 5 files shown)


gaul:
Requires #2924.

ggtakec:
#2924 was merged, so please resolve conflicts.

ggtakec (approved):
This looks good.
Handling cases where disk space runs out does get complicated, but I think this is simpler than the previous code and also fixes the issue.
Thanks.

PR #2924, opened 2026-07-20
Fix lock-order inversion in FdManager::ChangeEntityToTempPath

gaul:
Commit 1a50b9a removed fd_manager_lock from ChangeEntityToTempPath because it is called from FdEntity::NoCacheLoadAndPost while the entity locks (fdent_lock, fdent_data_lock) are held, whereas every other FdManager method acquires fd_manager_lock *before* the entity locks (e.g. Close, Open, Rename, GetPseudoFdCount).  Commit 3d6975b then re-added an fd_manager_lock guarded fent.find(path) pre-check, which reintroduced the inversion: a thread flushing under the entity locks blocks on fd_manager_lock while another thread holding fd_manager_lock blocks on that entity's lock, hanging the whole mount.

The pre-check only served to avoid re-keying an entity that is no longer mapped under path in fent (cache directory not specified, or NoCacheLoadAndPost already ran).  Move that decision into UpdateEntityToTempPath, which already runs under fd_manager_lock: when the staged path no longer maps to the entity, drop the staging entry instead of creating a spurious recovery mapping.  This keeps 3d6975b's behavior, and performs the check atomically with the move (the old pre-check had a gap between checking and moving), while ChangeEntityToTempPath again touches only except_fent under except_entmap_lock.

### DIFF
--- a/src/fdcache.cpp
+++ b/src/fdcache.cpp
@@ -790,21 +790,17 @@ bool FdManager::Close(FdEntity* ent, int fd)
 bool FdManager::ChangeEntityToTempPath(std::shared_ptr<FdEntity> ent, const char* path)
 {
     // [NOTE]
-    // If the path element does not exist in fent, it may be because a cache directory
-    // has not been specified, or FdEntity::NoCacheLoadAndPost has already been called.
-    // In these cases, the path element(=ent) has already been registered as a TempPath
-    // element from fent, so there is no need to register ent in the except_fent map.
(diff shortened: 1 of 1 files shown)


gaul:
Fixes ThreadSanitizer warning observed in https://github.com/s3fs-fuse/s3fs-fuse/actions/runs/29769323764/job/88443228379?pr=2923 .

ggtakec (approved):
Thanks for fix a bug,

PR #2929, opened 2026-07-22
Do not share the readdir listing cache with incompatible listings

gaul:
The per-directory S3ObjList cache added in 052b0f9 stored three different listing shapes in one slot: s3fs_readdir's complete one-level listing, directory_empty's max-keys=2 first page, and rename_directory's recursive no-delimiter listing.  Consumers cannot tell them apart:

- After rmdir or rename failed on a non-empty directory, s3fs_readdir served the truncated two-key page, so ls showed only the first entry.  This is visible when the children are not in the stat cache, for example on a fresh mount, because readdir merges stat cache entries into its result.
- After ls, rename_directory consumed the one-level listing as if it were the recursive object list, so objects under subdirectories were never renamed and mv failed half way with ENOTEMPTY, leaving the tree split between the old and new names.

Keep s3fs_readdir as the only producer of the cached listing. directory_empty still reads it, because a complete one-level listing answers the emptiness check exactly, but no longer stores its truncated page.  rename_directory always requests a fresh recursive listing and does not store it.

The new integration tests create the objects with s3_cp so that the stat cache cannot mask the listing cache behavior.

### DIFF
--- a/src/cache.h
+++ b/src/cache.h
@@ -109,6 +109,14 @@ class StatCache
         bool AddStat(const std::string& key, const struct stat& stbuf, const headers_t& meta, objtype_t type, bool notruncate = false);
         bool AddStat(const std::string& key, const struct stat& stbuf, objtype_t type, bool notruncate = false);
         bool AddNegativeStat(const std::string& key);
+
+        // [NOTE]
+        // The cached S3ObjList must always be the complete one-level
+        // listing(delimiter=/) of the directory, as stored by
+        // s3fs_readdir.  Never store truncated or recursive(no
+        // delimiter) listings, because every caller of GetS3ObjList
+        // assumes this shape.
+        //
         bool AddS3ObjList(std::string key, const S3ObjList& list);
 
         // Update meta stats
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1347,27 +1347,28 @@ static int s3fs_unlink(const char* _path)
 
 static int directory_empty(const char* path)
 {
-    int       result = 0;
     S3ObjList head;
 
     // check s3objlist in cache
+    //
+    // [NOTE]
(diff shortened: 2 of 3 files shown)


ggtakec:
I believe it would be better to omit the `check_content_only` argument (effectively setting it to `false`) for the `list_bucket` function called by `directory_empty`.
(This implies deprecating the `check_content_only` argument and standardizing the `list_bucket` operation across all calls.)

Even if `directory_empty` stops caching, the most common use case involves immediately calling `list_bucket` to retrieve the full list after calling `directory_empty`; therefore, I think it is acceptable for `directory_empty` to read and cache everything.

To provide some background, the previous stat cache implementation did not store `S3ObjList`.
Consequently, multiple `list_bucket` calls would occur during the `directory_empty` call and the surrounding operations.
To reduce this overhead, `check_content_only` was set to `true` for `directory_empty` calls, limiting `max-keys` to 2.

In contrast, the current stat cache stores `S3ObjList`, so reading all objects via the `list_bucket` call within `directory_empty` would not result in wasted effort.

ggtakec:
@gaul I have opened PR #2964 as a Draft. Please feel free to take a look when you get a chance.

In that PR, I took a different approach regarding the `S3ObjList` cache compared to this PR:
* **`directory_empty`:** When utilizing the cache, it now retrieves all entries instead of just the first one and caches the `S3Object`s as well. (This is the opposite of this PR, which updated it to fetch only the first entry.) When the cache is disabled, it retains the conventional behavior of retrieving only the first entry.
* **`rename_directory`:** By handling this process reentrantly, subdirectories are now properly supported as well. As a result, `S3Object` caching continues to work as before.

Although the test cases added in this PR are not included in PR #2964 yet, I have verified the functionality on my local test environment.

ggtakec:

PR #2930, opened 2026-07-22
Fix trailing-slash name handling in S3ObjList HasName and Remove

gaul:
Both methods built the slash-less probe with substr(size - 1), which yields the last character "/" instead of the name without its terminating slash.  As a result, a slash-terminated query missed entries stored under the plain name: HasName("x/") did not find a file "x", and Remove("x/") erased the "x/" entry but left the normalized "x" alias behind.  All current callers pass leaf names without a slash, so this was latent.  Also reject empty names, which previously invoked undefined behavior via back().

Add a test_s3objlist unit test program covering the slash-insensitive lookup and removal semantics, which fails on the unfixed code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,6 +83,7 @@ src/s3fs
 src/test_auth
 src/test_curl_util
 src/test_page_list
+src/test_s3objlist
 src/test_string_util
 
 test/chaos-http-proxy-*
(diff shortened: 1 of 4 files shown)


ggtakec (approved):

ggtakec (commented):
Please resolve the conflicts and rebase.

ggtakec (approved):

PR #2932, opened 2026-07-26
Fix directory sub-type poisoning by readdir listings

gaul:
S3ObjList::insert typed slash-less CommonPrefixes entries as DIR_NOT_TERMINATE_SLASH, asserting a legacy "dir" object that the listing gives no evidence for.  readdir_multi_head forwards this type as a hint and the HEAD worker stores it verbatim into the stat cache exactly when a HEAD request for "dir/" succeeds, i.e. when the truth is DIR_NORMAL.  After any readdir, every chmod/chown/utimens/setxattr on a listed subdirectory then took the replace-directory path: DELETE of the "dir" key (which destroys a coexisting "dir" file object on backends that allow both) plus a full re-PUT of "dir/" from a fresh header set that drops metadata set by other clients, instead of a metadata copy of the existing object.

The wrong sub-type could also never heal: DirStatCache::AddHasLock updates stat and meta for an existing directory child but not the sub-type (isSameObjectTypeHasLock treats all directory sub-types as equal), and directory nodes survive DelStat, so even the truthful AddStat(objtype_t::DIR_NORMAL) issued after rebuilding the directory object was ignored and the next metadata update rebuilt it again.

Type CommonPrefixes entries as DIR_NORMAL (legacy no-slash directory objects appear as Contents keys and keep getting their type from the HEAD worker's header inference), and make DirStatCache::AddHasLock re-enter the child node when both the cached and the added type are directories with different sub-types, so that its own-path branch updates the sub-type and the data depending on it.

Add a cache_node unit test covering the CommonPrefixes type and the sub-type update, and an integration test that chmods a listed externally-created directory and asserts that the "dir/" object is copied, keeping the other client's metadata.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,6 +81,7 @@ doc/man/s3fs.1
 default_commit_hash
 src/s3fs
 src/test_auth
+src/test_cache_node
 src/test_curl_util
 src/test_page_list
 src/test_s3objlist
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -73,6 +73,7 @@ s3fs_LDADD = $(DEPS_LIBS)
 
 noinst_PROGRAMS = \
     test_auth \
+    test_cache_node \
     test_curl_util \
     test_page_list \
     test_s3objlist \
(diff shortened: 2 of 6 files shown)


ggtakec:
It seems good, but please rebase onto master to resolve the conflicts.

ggtakec (approved):

PR #2939, opened 2026-07-28
Add [[nodiscard]] to error-returning functions

gaul:
Annotate the S3fsCurl request methods, the FdEntity flush/upload/load chain, SyncFiller, CacheFileStat::Release, and utility error returns so that the compiler rejects silently ignored errors, and address the discards this surfaced:

* Handle SyncFiller::Fill and SufficiencyFill results in readdir; the non-seekable filler only fails when it cannot allocate memory, so abort the listing with -ENOMEM instead of silently truncating it
* Convert PageList::Compress, Resize, ClearAllModified and UntreatedParts::Duplicate to void since they always returned true, and remove a dead error branch in FdEntity::Open
* Make the intentional Release discards explicit in ~CacheFileStat and FdManager::RawCheckAllCache

merge_headers and the char* forms of s3fs_wtf8_encode/decode are left unannotated because their bool means "modified something", not an error, and callers legitimately ignore it.


ggtakec:
Please resolve the conflicts and rebase.

ggtakec:
Apologize for the inconvenience, but please resolve the conflict and rebase again.

ggtakec (approved):

PR #2940, opened 2026-08-03
Fix constexpr linkage and extend constexpr coverage

gaul:
compare_timespec(timespec, timespec) was declared constexpr in filetimes.h but defined in filetimes.cpp.  constexpr implies inline, so any translation unit other than filetimes.cpp that called this overload would fail to link.  It only linked because the sole external caller in fdcache_entity.cpp resolves to the non-constexpr (struct stat&, stat_time_type, ...) overload instead.  Move the definition into the header.  This also allows removing CompareStatCacheTime from cache_node.cpp, which was a byte-for-byte duplicate.

Also make a few pure leaf functions constexpr:

- to_acl now takes std::string_view and compares with ==, replacing an eight-way strcmp chain, and matches the already-constexpr str(acl_t) beside it.
- getCurlDebugHead moves to the header as a switch returning literals directly, dropping five pointless local variables.
- mask_sensitive_header's SensitiveHeaders becomes a constexpr std::array<std::string_view>, dropping the nullptr sentinel, the sentinel break, and three strlen calls.  The authorization: entry must remain first; see isAuthHeader.

Finally, switch namespace-scope constexpr in headers from static to inline.  static gives internal linkage, so each translation unit got its own copy at a distinct address.

### DIFF
--- a/src/string_util.cpp
+++ b/src/string_util.cpp
@@ -19,6 +19,7 @@
  */
 
 #include <algorithm>
+#include <array>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
@@ -34,6 +35,8 @@
 
 #include "s3fs_logger.h"
 #include "string_util.h"
+
+using namespace std::string_view_literals;
 
 //-------------------------------------------------------------------
 // Functions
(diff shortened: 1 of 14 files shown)


src/string_util.cpp
```diff
@@ -679,16 +680,15 @@ const char* mask_sensitive_string_with_flag(const char* sensitive, bool nomask)
 
 std::string mask_sensitive_header(const char* pheader, size_t length)
 {
-    static const char* SensitiveHeaders[] = {
+    static constexpr std::array<std::string_view, 8> SensitiveHeaders = {
```
ggtakec: Omitting the array size in std::array<std::string_view, 8>, it helps prevent mistakes when updating the list later.

How about using class template argument deduction(CTAD) so it’s inferred automatically, like this?

```
static constexpr std::array SensitiveHeaders = {
    "authorization:"sv,                                   // do not change this position(see. isAuthHeader)
    "x-amz-security-token:"sv,
    "x-amz-credential:"sv,
    "x-amz-signature:"sv,
    "x-amz-server-side-encryption-customer-key-md5:"sv,
    "x-amz-server-side-encryption-aws-kms-key-id:"sv,
    "x-amz-copy-source-server-side-encryption-customer-key:"sv,
    "x-amz-copy-source-server-side-encryption-customer-key-md5:"sv
};
```

ggtakec (commented):
Please check if it is possible to make a correction.

src/string_util.cpp
gaul: Unfortunately not:
```
string_util.cpp: In function ‘std::string mask_sensitive_header(const char*, size_t)’:
string_util.cpp:683:49: error: wrong number of template arguments (1, should be 2)
  683 |     static constexpr std::array<std::string_view> SensitiveHeaders = {
      |                                                 ^
In file included from /usr/include/c++/16/bits/stl_algobase.h:63,
                 from /usr/include/c++/16/algorithm:62,
                 from string_util.cpp:21:
/usr/include/c++/16/bits/stl_pair.h:99:12: note: provided for ‘template<class _Tp, long unsigned int _Nm> struct std::array’
   99 |     struct array;
      |            ^~~~~
string_util.cpp:683:51: error: scalar object ‘SensitiveHeaders’ requires one element in initializer
  683 |     static constexpr std::array<std::string_view> SensitiveHeaders = {
      |                                                   ^~~~~~~~~~~~~~~~
```
Providing a too-small size generates a warning but a too-large size adds empty strings.

ggtakec: I didn't convey it well.
Adding `std::string_view_literals` as shown below should resolve the error.
(Please do not specify the `<>` after `std::array`. `sv` indicates that the string is defined as a `std::string_view` rather than a `char*`.)
```
using namespace std::string_view_literals;
...

static constexpr std::array SensitiveHeaders = {
    ...
};
```

gaul: Sorry I misunderstood -- this is C++17 CTAD.  Done as suggested.

ggtakec (approved):

PR #2944, opened 2026-08-04
Remove stdbuf from the integration tests

gaul:
Make the awk filter that prefixes s3fs log output flush each line via fflush() instead. stdbuf only affects libc stdio buffering via LD_PRELOAD, so two of its three uses never had an effect: s3fs emits each log line with raw write(2) (S3fsLog::Printf) and the JVM writes System.out through FileOutputStream. The only command that actually buffered was awk itself, which fflush() addresses directly. fflush() is supported by gawk, mawk, busybox awk, and BSD awk, and was standardized in POSIX.1-2024.

This also removes the macos-14 workaround for coreutils' gstdbuf lacking an arm64e slice.


ggtakec (approved):
Thank you for it is excellent that the stdbuf limitation has been removed.

PR #2945, opened 2026-08-04
Remove GNU sed from the Alpine CI packages

gaul:
Nothing needs it: busybox sed handles every sed invocation in the build and the test suite, including the case-insensitive s///i flag used to parse Content-Type headers.  Verified by building s3fs and running the unit tests in an alpine:3.24 container with only busybox sed installed.  Also drop the stale claim that the macos job uses gsed; brew never installed gnu-sed and nothing references it.


ggtakec:
Please rebase master to resolve the conflict.

ggtakec (approved):

PR #2946, opened 2026-08-04
Add a timeout to the macOS CI job

gaul:
This job usually completes in 10 minutes but sometimes hangs until the default 360 minute limit.


ggtakec (approved):
Thanks!

PR #2947, opened 2026-08-04
Upload test logs as CI artifacts on failure

gaul:
Only a filtered excerpt of test-suite.log reaches the console, and a run that hangs until the job timeout prints nothing at all because automake writes test-suite.log only after the whole run finishes. Attach the full logs, including the incrementally written per-test logs that capture hung runs, as artifacts.


ggtakec (approved):
This allows us to check the logs if a process is interrupted, which I think is a good idea.
However, since there is a storage limit for artifacts, we should make sure to delete them manually when they are no longer needed.
(I am actually a bit concerned about how large the log size might become if the execution time limit is reached.)

PR #2951, opened 2026-08-07
Update outdated fuse-t timestamp granularity assumptions in tests

gaul:
fuse-t 1.2.7 passes nanosecond timestamps through its NFS bridge, so the wait_ostype comment claiming second-granularity attribute values no longer holds; verified by round-tripping a nanosecond mtime through a live mount.

Remove the Darwin wait in test_extended_attributes: the xattr sequence passed 62 consecutive runs without it.

Keep the Darwin wait in test_external_modification: with use_cache and only the base 1 second wait, about 1 in 3 runs still read stale data because post-release getattrs extend stat_cache_interval_expire; 2 seconds was reliable.  Document that this is a caching effect, not timestamp granularity.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -558,6 +558,10 @@ function test_external_modification {
     # This extends the expiration date of the target file(1 sec by
     # stat_cache_interval_expire option).
     # Therefore, on MacOS, you need to add an additional 1 sec.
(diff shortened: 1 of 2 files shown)


ggtakec (approved):
Thank you. I think that is the right decision.

PR #2952, opened 2026-08-10
Send SSE-C copy source headers on multipart copy parts

gaul:
The mix multipart upload and the stream upload build their copy parts through S3fsCurl::MultipartUploadPartSetup, whose synthetic meta carries only x-amz-copy-source and the range.
MultipartUploadCopyPartSetup adds the copy source SSE-C headers only when that meta carries the source key md5, so every UploadPartCopy went out with the destination triple but no source key, and AWS and other enforcing servers reject reading the source with 400 InvalidRequest.  Regular put_headers copies do not have this problem because their meta comes from a HEAD response, which echoes the key md5.

Thread the source key md5 from the entity's metadata down to the part setup, and drop the copy headers mix_multipart_upload_request still wrote into its meta copy: they have been dead since the thread pool refactor moved request building into the workers, which is how the key went missing.

With use_sse=custom against s3proxy, the integration suite improves from 79 to 86 of 93 with plain flushes and passes all multipart mix and stream upload copy tests with -o streamupload; the remaining failures are unrelated to part copies.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -3470,7 +3470,7 @@ int S3fsCurl::PreMultipartUploadRequest(const char* tpath, const headers_t& meta
     return 0;
 }
 
-int S3fsCurl::MultipartUploadPartSetup(const char* tpath, int upload_fd, off_t start, off_t size, int part_num, const std::string& upload_id, etagpair* petag, bool is_copy)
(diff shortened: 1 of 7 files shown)


ggtakec:
The changes themselves look fine. Please rebase and resolve the conflicts.

I will re-check after the rebase that I am concerned about the following two code(could this be due to the conflict?):
- sse-c-part-copy: s3fs_threadreqs.cpp#L94, #L126
Although it doesn't appear in the diff for this PR, it looks like the return value check for the `pthparam->psyncfiller->Fill()` call has been removed.
(The check remains in the code on the master branch.)

PR #2953, opened 2026-08-10
Set the SSE-C headers in the meta of newly created objects

gaul:
When SSE-C is used, reading an object requires presenting its key, including when the object is the source of a copy request. put_headers() and the multipart copy functions resolve that key from the key md5 in their meta, which normally comes from a HEAD response echoing it.  But the meta built locally for a newly created file, directory, or symbolic link carried no SSE-C headers, and it is stored in the stat cache and the entity org meta, where it stands in for a HEAD response.  Until those entries expired, any meta updating copy on a fresh object, such as utimens or chmod, sent no copy source SSE-C headers and failed with 400 InvalidRequest against AWS and other enforcing servers, and get_object_sse_type() reported the object as unencrypted.

Seed the algorithm and the md5 of the current key, which new uploads are encrypted with, into the locally built meta at the creation sites.  With use_sse=custom against s3proxy, test_pjdfstest_utimensat now passes; the remaining failures are a separate multipart copy part bug and test harness limitations.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1183,6 +1183,15 @@ static int s3fs_create(const char* _path, mode_t mode, struct fuse_file_info* fi
         meta["x-amz-meta-xattr"] = *xattrvalue;
     }
 
+    // [NOTE]
+    // This meta is cached and used in place of a HEAD response, so it
+    // must carry the SSE headers describing how the object is stored.
+    //
+    if(!S3fsCurl::AddSseResponseHead(meta)){
+        S3FS_PRN_ERR("failed to set the SSE headers in the meta of %s", strpath.c_str());
+        return -EIO;
+    }
+
     // Set stat structure
     struct stat stbuf = {};
     if(!convert_header_to_stat(strpath, meta, stbuf, false)){
@@ -1502,6 +1511,15 @@ static int s3fs_symlink(const char* _from, const char* _to)
     headers["x-amz-meta-mtime"] = strnow;
     headers["x-amz-meta-uid"]   = std::to_string(pcxt->uid);
     headers["x-amz-meta-gid"]   = std::to_string(pcxt->gid);
+
+    // [NOTE]
+    // These headers are cached and used in place of a HEAD response, so
+    // they must describe how the object is stored.
+    //
+    if(!S3fsCurl::AddSseResponseHead(headers)){
+        S3FS_PRN_ERR("failed to set the SSE headers in the meta of %s", strTo.c_str());
+        return -EIO;
+    }
 
     // [NOTE]
     // Symbolic links do not set xattrs.
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -2814,6 +2814,42 @@ bool S3fsCurl::AddSseRequestHead(sse_type_t ssetype, std::string ssevalue, bool 
 }
 
 //
+// Set the SSE headers that a HEAD response would carry into meta
+//
+// This is the counterpart of AddSseRequestHead() for the meta which is
+// built locally for a newly created object and then cached in place of
+// a HEAD response.  The object is stored with the current SSE setting,
+// so the meta must describe it the same way the server would, without
+// the SSE-C key itself which a response never echoes.  New objects are
+// encrypted with the latest SSE-C key(position 0).
+//
+bool S3fsCurl::AddSseResponseHead(headers_t& meta)
+{
+    switch(S3fsCurl::GetSseType()){
+        case sse_type_t::SSE_DISABLE:
+            return true;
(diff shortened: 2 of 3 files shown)


src/s3fs.cpp
```diff
@@ -1073,6 +1073,28 @@ static int s3fs_readlink(const char* _path, char* buf, size_t size)
...
+//
+// When SSE-C is used, the correct key must be sent to read an object,
+// including as the source of a copy request.  put_headers() and the
+// other copy functions find that key by the md5 in their meta, which
+// normally comes from a HEAD response echoing it.  Meta made locally
+// for a newly created object is cached and used in place of a HEAD
+// response, so it must carry the same headers.  New objects are
+// encrypted with the first loaded key(position 0).
+//
+static void set_ssec_object_meta(headers_t& meta)
```
ggtakec: The headers added here should perhaps be made equivalent to those in `S3fsCurl::AddSseRequestHead()`.
It might also be better to prepare a static method that takes `headers_t` as an argument(equivalent to `S3fsCurl::AddSseRequestHead()`), and call that instead of the current function.

I think this approach seems preferable because it encapsulates the handling of these headers within the `S3fsCurl` class.

src/s3fs.cpp
```diff
@@ -1087,6 +1109,8 @@ static int create_file_object(const char* path, mode_t mode, uid_t uid, gid_t gi
     meta["x-amz-meta-ctime"] = strnow;
     meta["x-amz-meta-mtime"] = strnow;
 
+    set_ssec_object_meta(meta);
```
ggtakec: Adding the `x-amz-server-side-encryption-customer*` header to `meta` here appears to be a wasted operation, as it is not stored in the stats cache.
Furthermore, during the `put_header`(->`S3fsCurl::PutRequest`) process, any `x-amz-server-side-encryption-customer*` headers within the `meta` variable are ignored, and the header is re-set when the request is actually sent.
Therefore, if the meta is not stored to stats caching, I believe the call to `set_ssec_object_meta` here serves no purpose.

src/s3fs.cpp
```diff
@@ -1244,6 +1270,8 @@ static int create_directory_object(const char* path, mode_t mode, const struct t
         meta["x-amz-meta-xattr"] = pxattrvalue;
     }
 
+    set_ssec_object_meta(meta);
```
ggtakec: Same as the comment on line 1112.

ggtakec (commented):
I have a few questions, so I've left a review comment.

And, although not visible in the diff, the method calls for SyncFiller have changed. (L3699, L3737, L3764–L3776)

src/s3fs.cpp
gaul: Good idea, done.  Replaced the file-local helper with `S3fsCurl::AddSseResponseHead(headers_t& meta)`, placed next to `AddSseRequestHead()` and written as its response-side counterpart: the same `switch` over `sse_type_t` and the same error reporting, now covering every type rather than only SSE-C -- `x-amz-server-side-encryption: AES256` for SSE-S3, `aws:kms` plus the kms key id for SSE-KMS, and the algorithm plus key md5 for SSE-C.  The SSE-C key itself is left out, since a HEAD response never echoes it.

src/s3fs.cpp
gaul: You are right, removed.  The meta here only feeds `put_request()`, which sets the SSE headers itself when it sends the request, and `convert_header_to_stat()`; `s3fs_mknod()` then caches only the `struct stat`, not the headers.  So nothing observed the headers this added.

src/s3fs.cpp
gaul: Removed for the same reason.  The remaining two call sites are `s3fs_create()` and `s3fs_symlink()`, where the meta really does reach the stat cache (`AddStat()`/`AddSymlink()` with headers) and the `FdEntity` org meta, and so stands in for a HEAD response.

gaul:
Thanks for the review!

Regarding SyncFiller: those were not intended changes.  The branch was based on master from before the `[[nodiscard]]` commit, so the newly added `syncfiller.Fill()` result checks showed up as removals.  I have rebased onto current master and that noise is gone.

src/s3fs.cpp
ggtakec: Thanks for the update. I've reviewed it.

src/s3fs.cpp
ggtakec: I've reviewed it.

src/s3fs.cpp
ggtakec: Thanks for the update. I've reviewed it.

ggtakec (approved):

PR #2954, opened 2026-08-10
Send SSE-C headers from the test helpers under use_sse=custom

gaul:
The s3_head and s3_cp helpers make raw signed requests, but sent no SSE-C headers, so heading an object that the mount encrypted failed with 400 in test_create_empty_file, test_mv_file,
test_multipart_copy, test_content_type, and
test_chmod_listed_directory_keeps_metadata, and the files those failed tests left behind broke test_list.

While S3FS_SSE_CUSTOM_KEYFILE is exported, the helpers now send the SSE-C headers for the first key in that file, which is the key s3fs encrypts new objects with.  s3_cp sends them too, so every object in an SSE-C lane is encrypted with the same key whether it was written through the mount or by a helper, and either side can read the other's.  The flag loop exports the key file parsed from the flag.

Also reword the TODO on the disabled SSE combos: S3Proxy now supports SSE on a branch, but the pinned release the harness downloads does not yet.

### DIFF
--- a/test/small-integration-test.sh
+++ b/test/small-integration-test.sh
@@ -58,8 +58,8 @@ if [ -n "${ALL_TESTS}" ]; then
         sigv2
         sigv4
         "singlepart_copy_limit=10 -o multipart_copy_size=10"  # limit sizes to exercise multipart copy code paths
-        #use_sse  # TODO: S3Proxy does not support SSE
-        #use_sse=custom:/tmp/ssekey  # TODO: S3Proxy does not support SSE
+        #use_sse  # TODO: needs an S3Proxy release with SSE support
+        #use_sse=custom:/tmp/ssekey  # TODO: needs an S3Proxy release with SSE support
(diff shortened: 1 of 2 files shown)


test/small-integration-test.sh
```diff
@@ -58,8 +58,8 @@ if [ -n "${ALL_TESTS}" ]; then
         sigv2
         sigv4
         "singlepart_copy_limit=10 -o multipart_copy_size=10"  # limit sizes to exercise multipart copy code paths
-        #use_sse  # TODO: S3Proxy does not support SSE
-        #use_sse=custom:/tmp/ssekey  # TODO: S3Proxy does not support SSE
+        #use_sse  # TODO: needs an S3Proxy release with SSE support
+        #use_sse=custom:/tmp/ssekey  # TODO: needs an S3Proxy release with SSE support
```
gaul, line 62: SSE and SSE-C support are coming in S3Proxy 4.1.0.

ggtakec (approved):

PR #2957, opened 2026-08-10
Skip test_concurrent_directory_updates on macOS

gaul:
A macOS NFS client kernel bug can permanently wedge processes that race open/unlink/recreate of the same names.  FUSE-T mounts are NFS, so the test intermittently hangs the macos-14 CI job until its 20 minute timeout: the racing syscalls are never delivered to s3fs while the rest of the mount stays serviceable.  The FUSE-T maintainer confirms the bug reproduces on a native NFS mount with the same script and that Apple has known about it for years:

https://github.com/macos-fuse-t/fuse-t/issues/112

Linux continues to run the test.

### DIFF
--- a/test/integration-test-main.sh
+++ b/test/integration-test-main.sh
@@ -3300,7 +3300,15 @@ function add_all_tests {
     add_tests test_copy_file
     add_tests test_write_after_seek_ahead
     add_tests test_overwrite_existing_file_range
-    add_tests test_concurrent_directory_updates
+    # [NOTE]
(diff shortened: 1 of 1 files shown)


gaul:
Let's see if #2955 improves CI reliability first.

ggtakec:
For reference: While reviewing the test failure logs on macOS-14, I noticed a case where requests were logged for a test file used in test_copy_file, even though the log indicated that test_concurrent_directory_updates was running at that time.

While log entries can sometimes be slightly out of order, this still seems a bit odd.
(It might also be related to stat caching.)

PR #2958, opened 2026-08-12
Remove IBM-specific knowledge from the curl transport layer

gaul:
S3fsCurl reached into S3fsCred to ask whether IBM IAM authentication was in use, and built the IBM OAuth token request itself.  Neither belongs in the transport layer.

GetIAMCredentials now takes an opaque post body and Authorization header instead of an IBM apikey; S3fsCred::LoadIAMCredentials builds both.  The IAMCRED case of ResetHandle keys off postdata rather than IsIBMIAMAuth -- the block is a copy of the unconditional COMPLETEMULTIPOST case and only takes effect on the RemakeHandle retry path, where postdata has just been restored from b_postdata.  GetIAMCredentials sets the same four options itself on the initial pass.

No functional change.

### DIFF
--- a/src/curl.cpp
+++ b/src/curl.cpp
@@ -1915,7 +1915,13 @@ bool S3fsCurl::RemakeHandle(bool keepAuthHeader)
             if(CURLE_OK != curl_easy_setopt(hCurl, CURLOPT_WRITEFUNCTION, WriteMemoryCallback)){
                 return false;
             }
-            if(S3fsCurl::ps3fscred->IsIBMIAMAuth()){
+            // [NOTE]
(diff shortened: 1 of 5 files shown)


gaul:
@Bhagyashreek8 FYI.

ggtakec:
The changes look good.
However, (as with the previous PR,) there are unintended changes from master in the following part(related to filler).
`gaul:ibm-code-extraction` s3fs_threadreqs.cpp L94, L126
Could you please revert them to match master?

gaul:
> Could you please revert them to match master?

Rebased.

ggtakec (approved):

PR #2959, opened 2026-08-14
Update timestamps on truncate, removexattr and before the epoch

gaul:
Three unrelated timestamp defects found by xfstests generic/313, 728 and 258.

truncate(2) only updated ctime and mtime when the file was not already open, because the update sat inside the "need_flush" branch that runs when there is no open descriptor.  ftruncate(2) on an open file, and truncate(2) on a file another descriptor holds open, therefore left both times untouched.  Set the times whichever path is taken, and when the file is open mark the metadata dirty so the flush at close uploads them and refresh the stat cache so it stops serving the pre-truncate values.

removexattr(2) did not update ctime, though setxattr(2) already did.

Times before the epoch were discarded.  get_mtime(), get_ctime() and get_atime() reported a missing header as a negative sentinel and then treated every negative tv_sec as invalid, so a real pre-1970 timestamp fell back to Last-Modified; convert_header_to_stat() clamped what was left to the epoch, and valid_timespec() rejected it a third time. Report absence separately from the parsed value and key the remaining checks off the existing UTIME_OMIT marker, which already means "no value", so a negative number of seconds survives the round trip.

test_metaheader covers the parsing; it fails against the previous code. The xfstests generic/quick group goes from 26 to 23 failures with no new ones, and the ALL_TESTS integration matrix passes all 11 legs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,6 +83,7 @@ src/s3fs
 src/test_auth
 src/test_cache_node
 src/test_curl_util
+src/test_metaheader
 src/test_page_list
 src/test_s3objlist
 src/test_string_util
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -75,6 +75,7 @@ noinst_PROGRAMS = \
     test_auth \
     test_cache_node \
     test_curl_util \
+    test_metaheader \
     test_page_list \
     test_s3objlist \
     test_string_util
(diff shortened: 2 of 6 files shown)


ggtakec:
Thanks for the bug fix, your changes look good.
However, since this includes unnecessary changes related to `filler` similar to the other PRs listed below, please revert them to match the master branch.

`gaul:fix-timestamps` s3fs.cpp L3689, L3727, L3755-3766

gaul:
Rebased.

ggtakec (approved):

PR #2960, opened 2026-08-14
Stop a truncated listing from standing in for a directory's contents

gaul:
directory_empty() only needs to know whether a directory has any child, so it lists with max-keys=2 and stops after the first page.  It then stored that deliberately truncated listing in the child listing cache that readdir() reads as the complete list of a directory's children.  A failed rmdir(2) therefore made the directory appear to hold at most two entries:

    $ mkdir d && for i in $(seq 10); do : > d/f$i; done   # cold cache
    $ rmdir d
    rmdir: failed to remove 'd': Directory not empty
    $ ls d | wc -l
    1

The same cache is consulted by rename_directory(), so renaming such a directory afterwards moved only the two cached objects and left the rest stranded under the old key prefix, while reporting success:

    $ mv d e                    # 11 objects under d/
    $ # 9 objects stranded under d/, 2 moved to e/

Do not cache the truncated listing.  A complete listing that is already cached still answers the question, so it is still read.

rename_directory() also must not use that cache at all: it needs a recursive listing, whereas readdir() fills the cache with a "/" delimited one holding only direct children, and writing its recursive listing back would in turn make readdir() report nested paths as direct entries.  It now always lists directly.

xfstests generic/013 failed 5 of 5 runs before this change and 2 of 5 after, so a further problem remains under parallel fsstress; the generic/quick group goes from 26 failures to 25 with no new ones, and the ALL_TESTS integration matrix passes all 11 legs.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1352,20 +1352,20 @@ static int directory_empty(const char* path)
 
     // check s3objlist in cache
     if(!StatCache::getStatCacheData()->GetS3ObjList(path, head)){
+        // [NOTE]
+        // Only the presence of a child matters here, so this listing is
+        // deliberately truncated(max-keys=2 and no paging).  It must not be
+        // stored in the shared child listing cache: readdir() reads that cache
+        // as the complete list of a directory's children, so caching this would
+        // make a directory with many entries appear to hold at most two.
+        //
         if((result = list_bucket(path, head, "/", true)) != 0){
             S3FS_PRN_ERR("list_bucket returns error.");
             return result;
         }
-        if(!head.IsEmpty()){
-            if(!StatCache::getStatCacheData()->AddS3ObjList(path, head)){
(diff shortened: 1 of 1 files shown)


ggtakec:
@gaul I have opened PR #2964 as a Draft. Please take a look when you have a chance.
(I have left the same comment on #2929 as well.)

In that Draft PR, I took a different approach regarding the `S3ObjList` cache:

* **`directory_empty`:** When using the cache, it now retrieves all entries rather than just the first one and caches the `S3Object`s as well. (This is the opposite of this PR, which updated it to fetch only the first entry.) When the cache is not enabled, it maintains the traditional behavior of fetching only the first entry.
* **`rename_directory`:** By handling this process reentrantly, subdirectories are now properly supported as well. As a result, `S3Object` caching continues to work as before.

ggtakec:

ggtakec (approved):
Could you please resolve the conflict?

PR #2961, opened 2026-08-14
Include not yet uploaded files in a cached directory listing

gaul:
A file exists as soon as it is created, but its object is only uploaded when it is flushed.  A listing taken in between does not mention it, and readdir() cached that listing as-is, so the file stayed missing from the directory until the entry expired.

Because rmdir(2) decides emptiness from the same cached listing, this also loses the file: a create racing a listing of its parent hid the file in 34 of 200 attempts, and in every one of those the directory was then removed and the object left unreachable under a prefix nothing points to.

The Stat Cache does know about the file, since its entry is kept with NoTruncate until the upload happens.  Merge the children it holds into the listing before caching it, which is what rename_directory() already does for the same reason.  The same 200 attempts then lose nothing.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -3781,6 +3781,28 @@ static int s3fs_readdir(const char* _path, void* buf, fuse_fill_dir_t filler, of
             return result;
         }
 
+        // [NOTE]
+        // A file exists as soon as it is created, but its object is only
+        // uploaded when it is flushed, so a listing taken in between does not
+        // mention it.  The Stat Cache does know about it(the entry is kept
+        // with NoTruncate until the upload happens), so merge in the children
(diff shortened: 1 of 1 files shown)


ggtakec:
@gaul I have opened PR #2964 as a Draft. Please feel free to take a look when you get a chance.
(I have left the same comment on #2929 as well.)

In that Draft PR, I modified `StatCache` so that if a child object with the `notruncate` flag is cached, its parent directory object will not be cached out(evicted). 
While this might cause the cache to exceed its normal expiration time, it should not be a major issue. Adding this eviction restriction directly within `StatCache` eliminates the need to handle this workaround outside of the cache system.

Please refer to the PR above for details.

ggtakec:

PR #2963, opened 2026-08-17
Use C++17 structured bindings and if-with-initializer

gaul:
Convert map iterator loops to range-for with structured bindings, destructure existing range-for loops that used pair.first/second, and scope map.find results into their if statements where the iterator is not used afterwards.  Also replace the remaining (*iter).member spellings with iter->member and constify the PutRequest meta parameter, which the loop conversion exposed as read-only.

Loops that erase while iterating or reuse the iterator after the loop are left unchanged.


gaul:
This is a big diff but there are no functional changes.

ggtakec (approved):

ggtakec:
@gaul Please rebase to resolve the conflicts.

gaul:
Merging due to previous approval.

PR #2964, opened 2026-08-23
Prevent expiration of Directory StatCaches with no_truncate child nodes

ggtakec:
#2929, #2960, #2961

### Issue
When a newly created file has not yet been flushed, its metadata exists solely within the StatCache because the file has not yet been uploaded to the server.
If the parent directory's StatCache is cached out (evicted) in this state, rebuilding the directory's file list from the server causes an inconsistency, as the un-uploaded file will not be included.

### Solution
Objects that have not yet been uploaded to the server have a `no_truncate` flag set in their StatCache.
Therefore, when eviction of a parent directory's cache is requested, eviction is now prevented if the directory contains any child objects with the `no_truncate` flag.

Additionally, when utilizing the StatCache during `directory_empty` checks, the behavior has been updated to list all files(S3Objects) in the directory and cache them.
(Previously, it did not perform a full listing.)
In typical use cases, checking whether a directory is empty is often followed by operations that retrieve the file list, so performing a full listing during `directory_empty` should be fine.

### Context / Notes
This PR demonstrates an alternative approach to #2929, #2960, and #2961.
While the approach in the original set of PRs also resolves the issue, I believe leveraging the StatCache effectively(especially for newly created files) is a better direction.

This work originally started as an effort to fix slow performance when renaming directories containing a large number of files.
I decided to create this PR thinking it would be best to address the issues raised in #2929, #2960, and #2961 at the same time.
Thanks to many hints from those PRs, I arrived at the conclusion that updating `StatCache` itself might be the best solution.

### Future Work (Backlog)
I discovered a separate issue in `StatCache` unrelated to this PR or #2929, #2960, and #2961.
I plan to work on a fix for that issue soon.

@gaul 
Could you please compare this PR with #2929, #2960, and #2961 and let me know your thoughts?

### DIFF
--- a/src/cache_node.cpp
+++ b/src/cache_node.cpp
@@ -200,7 +200,7 @@ bool StatCacheNode::ResumeExpireCheck()
 //-------------------------------------------------------------------
 // Methods
 //-------------------------------------------------------------------
-StatCacheNode::StatCacheNode(const char* path, objtype_t type) : cache_type(type), fullpath(path ? path: "")
+StatCacheNode::StatCacheNode(const char* path, const std::shared_ptr<StatCacheNode>& parentdir, objtype_t type) : cache_type(type), fullpath(path ? path: ""), parent(parentdir)
 {
     if(IS_DIR_OBJ(cache_type)){
         // directory type must end with '/'.
@@ -222,6 +222,19 @@ StatCacheNode::StatCacheNode(const char* path, objtype_t type) : cache_type(type
 
 StatCacheNode::~StatCacheNode()
 {
+    if(notruncate){
+        // [NOTE]
+        // No locking is performed when calling GetParentHasLock()
+        // within the destructor.
+        // Similarly, the call to pParent->DecrementNoTruncateChildHasLock()
+        // here does not involve locking. This is because the destructor for
+        // the StatCacheNode class (and its derived classes) is invoked with
+        // pParent already locked.
+        //
+        if(auto pParent = GetParentHasLock()){
+            pParent->DecrementNoTruncateChildHasLock();
+        }
+    }
     StatCacheNode::DecrementCacheCount(objtype_t::UNKNOWN);
 }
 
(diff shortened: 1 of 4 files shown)


ggtakec:
I figured out why the CI is failing only on macOS.

### Detail
In the `test_external_directory_creation` test:
This test creates a file (`/testrun-XXXX/directory/test-s3fs-001.txt`) externally via AWS CLI and then deletes it.
However, the intermediate directory (`/testrun-XXXX/directory`) created externally does not exist as an actual object.

During test execution, after the file deletion(`s3fs_unlink: /testrun-XXXX/directory/test-s3fs-001.txt`), a request to fetch attributes is made for its parent directory(`s3fs_getattr: /testrun-XXXX/directory`).

* **Before this PR:**
  The Stat cache for the directory(`/testrun-XXXX/directory`) **remained** and returned the directory attributes, allowing the test to complete successfully.
* **With this PR:**
  After deleting the file(`/testrun-XXXX/directory/test-s3fs-001.txt`), the parent directory's Stat cache is checked. Since the directory object(`/testrun-XXXX/directory`) does not exist, it is removed from the Stat cache. Then, `-ENOENT(-2)` is returned for the subsequent attribute request(`s3fs_getattr: /testrun-XXXX/directory`).

It seems that once this response is returned to FUSE(fuse-t), fuse-t stops sending any further responses or commands(making it appear as if the file deletion never completed).
When tested on a physical machine, a system dialog appears for NFS stating like `Server connection interrupted`, prompting a forced unmount.

The test hangs at this point and eventually times out.

---

### Summary

* **Before PR:** After deleting `/testrun-XXXX/directory/test-s3fs-001.txt`, directory attributes for `/testrun-XXXX/directory` could be fetched.
* **With PR:** Directory attributes cannot be fetched, causing fuse-t to stall (or appear to stall).

---

### Going Forward

I am considering if any workaround can be implemented on our end in this PR, but we first need to check with the fuse-t side.
This issue seems to have surfaced because previously, the valid cache masked requests for non-existent directories (where neither a directory object nor any object under that path exists).

ggtakec:
Addressed an issue on macOS(see: https://github.com/s3fs-fuse/s3fs-fuse/pull/2964#issuecomment-5650686149).

In FUSE-T, when deleting an S3 file object(ex. `/path1/file`) whose intermediate directory object(ex. `/path1`) does not exist, the intermediate directory object was also being removed from StatCache.
While this is technically the expected behavior, if `getattr()` is called after `unlink()` completes and returns that the intermediate directory does not exist, FUSE-T hangs up.

To work around this issue, StatCache has been modified so that such intermediate directory objects(type `DIR_NOT_EXIST_OBJECT`) are not deleted within their expiration period, even if they lack metadata or S3 object data.
This prevents the hang-up during `unlink()` on macOS(FUSE-T).

As a result, the intermediate directory object will appear to exist within its expiration time after the file object is deleted, but this causes no issues on Linux systems(it simply looks like an empty directory exists).

PR #2966, opened 2026-08-23
Extract the IBM IAM code into its own class and file

gaul:
IBM Cloud Object Storage support was spread across S3fsCred and S3fsCurl as a boolean flag consulted in thirteen places.  Collect it in a new IbmIamCred class: the two options, the token endpoint URL, the OAuth POST body, the token response parser, and the bearer token headers.  The class touches no other s3fs class, so it comes with a unit test -- the first coverage of any of this code.

Splitting the response parser lets IAM_token_field and IAM_expiry_field go away.  They existed only so that "ibm_iam_auth" could retune the shared parser, which is why the IBM token field was spelled with literal quote characters inside the field name.  The AWS parser now names its fields as constants beside IAMCRED_ACCESSKEYID, and no longer has to ask whether it is parsing an integer or a quoted string.  Likewise the IBM path no longer needs IAM_field_count or the forced IMDSv1, since it never reaches the code that reads them.

Two option ordering bugs fall out of the split:

- "ibm_iam_endpoint" was applied by overwriting IAM_cred_url, which "ibm_iam_auth" then unconditionally reset.  Given in that order the endpoint was silently discarded and the public endpoint used instead. IbmIamCred stores the endpoint and derives the URL when it is read.

- "ibm_iam_auth" is rejected together with ecs and iam_role, but only when it comes first.  Check the combination again in CheckAllParams, where the order no longer matters.


gaul:
@Bhagyashreek8 FYI.

Bhagyashreek8:
@gaul Thank you for the update.

ggtakec (approved):

Issue #2809, opened 2026-02-25
Mounting Bucket Subdirectory Fails in v1.97


Benedikt0x4C:
We use `s3fs-fuse` to mount subdirectories in S3 buckets. These subdirectories were properly created through the file system before (**not** using the AWS console).

We've been using `s3fs-fuse` v1.95 for quite some time without any issues; it mounts subdirectories fine. As an example:
```
s3fs my-s3-bucket:/assets /var/www/assets -o iam_role
```

That said, after the update to v1.97, it fails to mount with the exact same commands, if it contains an S3 bucket subdirectory. It does not show any error message or logs any error that we could find. It doesn't matter if you use the command line or `/etc/fstab` to mount a subdirectory of an S3 bucket, it will refuse to mount in both cases. What still worked in v1.97 is if you simply mount the root directory of an S3 bucket, e.g., `s3fs my-s3-bucket /var/www/assets -o iam_role`.

After downgrading `s3fs-fuse` to v1.95 again, mounting a subdirectory of an S3 bucket worked flawlessly again.

We're using official packages that are distributed with Rocky Linux 9.

#### Version of s3fs being used (`s3fs --version`)
Amazon Simple Storage Service File System V1.97 with OpenSSL

#### Version of fuse being used (`pkg-config --modversion fuse`, `rpm -qi fuse` or `dpkg -s fuse`)
Name        : fuse
Version     : 2.9.9
Release     : 16.el9
Architecture: x86_64

#### Provider (`AWS`, `OVH`, `Hetzner`, `iDrive E2`, ...)
AWS

#### Kernel information (`uname -r`)
5.14.0-611.30.1.el9_7.x86_64

#### GNU/Linux Distribution, if applicable (`cat /etc/os-release`)
NAME="Rocky Linux"
VERSION="9.5 (Blue Onyx)"

#### How to run s3fs, if applicable
[x] command line
[x] /etc/fstab

Command line:
```
s3fs my-s3-bucket:/assets /var/www/assets -o iam_role
```

/etc/fstab:
```
my-s3-bucket:/assets /var/www/assets fuse.s3fs _netdev,allow_other,iam_role,umask=002,uid=1000,gid=48 0 0
```

#### s3fs syslog messages (`grep s3fs /var/log/syslog`, `journalctl | grep s3fs`, or `s3fs outputs`)
(No relevant logs in `journalctl`.)

ggtakec:
[...]

PR #2967, opened 2026-08-26
Fix assertion when mounting an implicit bucket subdirectory

n3crosis:
## Summary

This fixes a crash when mounting a bucket subdirectory with `-o compat_dir` when that directory exists only as an implicit S3 prefix.

For example, `prefix/file.txt` exists, but neither the `prefix/` directory-marker object nor the `prefix_$folder$` compatibility object exists. This is a normal S3 layout: prefixes do not require directory objects.

In this situation s3fs discovers the prefix by listing it during mount initialization. The mount root becomes an empty relative path, and that empty key is passed to `StatCache::AddS3ObjList()`. The function calls `std::string::back()` before checking whether the key is empty, causing an assertion abort:

```text
std::string::back(): Assertion '!empty()' failed
```

The visible result is a failed mount, often followed by `Transport endpoint is not connected`.

Related to #2809.

## Fix

The stat-cache tree already represents its root as `/`. Normalize the empty mount-root key to that value before adding the trailing slash used for all cached directory listings.

```diff
diff --git a/src/cache.cpp b/src/cache.cpp
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@
 bool StatCache::AddS3ObjList(std::string key, const S3ObjList& list)
 {
     const std::lock_guard<std::mutex> lock(StatCache::stat_cache_lock);
 
+    if(key.empty()){
+        key = "/";
+    }
     if('/' != key.back()){
         key += '/';
     }
```

This only changes the empty-key case; all existing non-root cache keys keep their current behavior.

## Reproduction

The following can be run on a normal Linux FUSE host. It uses the AWS CLI to create a child object without creating a directory marker, then mounts that implicit prefix.

Set the endpoint and credentials first. `BUCKET` must already exist, and the AWS CLI credentials must be valid for the same endpoint as `PASSWD_FILE`.

```sh
export S3_URL=http://127.0.0.1:8081 # or another S3-compatible endpoint
export BUCKET=my-test-bucket
export PASSWD_FILE=/path/to/passwd-s3fs
export S3FS_BIN=/path/to/s3fs # defaults to s3fs when unset
```

Save and run this script:

```sh
#!/usr/bin/env bash
set -euo pipefail

: "${S3_URL:?set the S3-compatible endpoint URL}"
: "${BUCKET:?set an existing test bucket}"
: "${PASSWD_FILE:?set the s3fs credentials file}"
s3fs_bin="${S3FS_BIN:-s3fs}"

prefix="implicit-prefix-$$"
mountpoint="$(mktemp -d)"
trap 'fusermount3 -u "$mountpoint" 2>/dev/null || true; rmdir "$mountpoint" 2>/dev/null || true' EXIT

# Create a child object only. Do not create prefix/ or prefix_$folder$.
printf 'data\n' | aws --endpoint-url "$S3_URL" s3 cp - "s3://$BUCKET/$prefix/file.txt"
! aws --endpoint-url "$S3_URL" s3api head-object --bucket "$BUCKET" --key "$prefix/" >/dev/null 2>&1

"$s3fs_bin" "$BUCKET:/$prefix" "$mountpoint" \
  -o passwd_file="$PASSWD_FILE" \
  -o use_path_request_style \
  -o url="$S3_URL" \
  -o compat_dir

test -f "$mountpoint/file.txt"
echo "implicit subdirectory mount succeeded"
```

## Expected results

- v1.95: the mount succeeds and `file.txt` is visible.
- v1.97: s3fs aborts with the empty-string assertion above.
- This patch: the mount succeeds and `file.txt` is visible.

The regression was introduced after v1.95 by `052b0f9d0535b5ef6d035431305e385cf9f09617` ("Improved stat cache to include ListBucket results").


ggtakec (approved):
I attempted to reproduce the issue by using a directory path without a corresponding directory object as the mount point and referencing the file objects underneath it, but I could not confirm any errors in either v1.97 or the current `master` branch.
However, the behavior of the `back()` method when the key is empty is indeed problematic, so it makes sense to include this fix as a fail-safe.
Therefore, I approve this PR.

@n3crosis Thank you for your kindness.

n3crosis:
@ggtakec I reproduced the issue with the following Docker/S3Proxy script. It creates an implicit prefix (`prefix/file.txt`) without `prefix/` or `prefix_$folder$`, verifies both corresponding `HEAD` requests return `404`, and then tests the affected and fixed revisions. v1.97 is the default version installed via apt, but it is affected by this issue. The current solution is to pin s3fs to v1.95 until a release that includes the merged fix is available.

Expected results:

- v1.95 mounts successfully.
- v1.97 aborts with `std::string::back(): Assertion '!empty()' failed`. 
- Merged commit `0b95042f65e86db42653ff633b20db7ac5fd89ed` mounts successfully.

```sh
#!/usr/bin/env bash

set -euo pipefail

readonly RUN_ID="s3fs-implicit-prefix-${RANDOM}${RANDOM}"
readonly NETWORK="${RUN_ID}-net"
readonly S3PROXY_CONTAINER="${RUN_ID}-s3proxy"
readonly TEST_BUCKET='testbucket'
readonly TEST_PREFIX='prefix'
readonly S3PROXY_USER='reproaccess'
readonly S3PROXY_PASSWORD='reprosecret'
S3PROXY_PORT=''

cleanup() {
    docker rm -f "${S3PROXY_CONTAINER}" >/dev/null 2>&1 || true
    docker network rm "${NETWORK}" >/dev/null 2>&1 || true
}
trap cleanup EXIT

require_docker() {
    command -v docker >/dev/null
    command -v curl >/dev/null
    test -e /dev/fuse
}

# Send an S3 request directly to the disposable unauthenticated S3Proxy.
s3_curl() {
    curl \
        --fail-with-body --silent --show-error \
        "$@"
}

s3_status() {
    curl \
        --silent --output /dev/null --write-out '%{http_code}' \
        "$@"
}

start_s3proxy() {
    docker network create "${NETWORK}" >/dev/null
    docker run --detach --rm \
        --name "${S3PROXY_CONTAINER}" \
        --network "${NETWORK}" \
        --publish '127.0.0.1::80' \
        -e S3PROXY_AUTHORIZATION=none \
        -e S3PROXY_IDENTITY="${S3PROXY_USER}" \
        -e S3PROXY_CREDENTIAL="${S3PROXY_PASSWORD}" \
        andrewgaul/s3proxy:latest >/dev/null

    S3PROXY_PORT="$(docker port "${S3PROXY_CONTAINER}" 80/tcp | awk -F: '{print $NF}')"

    for _ in $(seq 1 30); do
        if curl --fail --silent --show-error \
            "http://127.0.0.1:${S3PROXY_PORT}/healthz" >/dev/null 2>&1; then
            return 0
        fi
        sleep 1
    done

    docker logs "${S3PROXY_CONTAINER}"
    return 1
}

create_implicit_prefix() {
    # This creates prefix/file.txt only. It deliberately does not create
    # prefix/, which is the condition that triggered the assertion.
    s3_curl --request PUT --header 'Content-Length: 0' --data-binary '' \
        "http://127.0.0.1:${S3PROXY_PORT}/${TEST_BUCKET}"
    s3_curl --request PUT --data-binary $'fixture\n' \
        "http://127.0.0.1:${S3PROXY_PORT}/${TEST_BUCKET}/${TEST_PREFIX}/file.txt"
}

assert_implicit_prefix() {
    local prefix_url="http://127.0.0.1:${S3PROXY_PORT}/${TEST_BUCKET}/${TEST_PREFIX}"
    local status

    status="$(s3_status --head "${prefix_url}/")"
    test "${status}" = 404 || {
        echo "Expected HEAD ${TEST_PREFIX}/ to return 404, got ${status}" >&2
        return 1
    }
    status="$(s3_status --head "${prefix_url}_\$folder\$")"
    test "${status}" = 404 || {
        echo "Expected HEAD ${TEST_PREFIX}_\$folder\$ to return 404, got ${status}" >&2
        return 1
    }
}

run_variant() {
    local variant="$1"
    local ref="$2"

    echo "==> Testing ${variant}"

    docker run --rm --privileged --device /dev/fuse --network "${NETWORK}" \
        -e TEST_BUCKET="${TEST_BUCKET}" \
        -e TEST_PREFIX="${TEST_PREFIX}" \
        -e S3PROXY_HOST="${S3PROXY_CONTAINER}" \
        -e S3PROXY_USER="${S3PROXY_USER}" \
        -e S3PROXY_PASSWORD="${S3PROXY_PASSWORD}" \
        -e S3FS_REF="${ref}" \
        alpine:3.21 sh -ceu '
            echo "    installing build dependencies"
            apk add --no-cache \
                autoconf automake build-base ca-certificates curl-dev fuse-dev fuse3-dev \
                git libtool libxml2-dev linux-headers openssl-dev pkgconf >/dev/null

            echo "    fetching ${S3FS_REF}"
            git init -q /src
            cd /src
            git remote add origin https://github.com/s3fs-fuse/s3fs-fuse.git
            git fetch --depth 1 origin "${S3FS_REF}" >/dev/null 2>&1
            git checkout --detach -q FETCH_HEAD

            echo "    configuring"
            # The reported RHEL build detected the undefined empty-key access.
            # Enable the same libstdc++ bounds check for a deterministic test.
            export CXXFLAGS="-U_FORTIFY_SOURCE -D_GLIBCXX_ASSERTIONS"
            ./autogen.sh >/dev/null
            ./configure >/dev/null
            echo "    building"
            make -j2 >/dev/null

            echo "    mounting implicit prefix"
            printf "%s:%s\n" "${S3PROXY_USER}" "${S3PROXY_PASSWORD}" >/tmp/passwd-s3fs
            chmod 600 /tmp/passwd-s3fs
            mkdir -p /mnt/test

            ./src/s3fs "${TEST_BUCKET}:/${TEST_PREFIX}" /mnt/test -f \
                -o passwd_file=/tmp/passwd-s3fs \
                -o url="http://${S3PROXY_HOST}:80" \
                -o use_path_request_style \
                -o free_space_ratio=0 \
                -o compat_dir &
            s3fs_pid=$!

            for _ in $(seq 1 20); do
                if ! kill -0 "${s3fs_pid}" 2>/dev/null; then
                    wait "${s3fs_pid}" || true
                    exit 10
                fi
                if test -f /mnt/test/file.txt; then
                    (fusermount3 -uz /mnt/test 2>/dev/null || \
                        fusermount -uz /mnt/test 2>/dev/null || \
                        umount -l /mnt/test) || true
                    wait "${s3fs_pid}" || true
                    exit 0
                fi
                sleep 1
            done

            (fusermount3 -uz /mnt/test 2>/dev/null || \
                fusermount -uz /mnt/test 2>/dev/null || \
                umount -l /mnt/test) || true
            wait "${s3fs_pid}" || true
            exit 11
        '
}

expect_success() {
    local variant="$1"
    shift
    if "$@"; then
        echo "PASS: ${variant}"
    else
        echo "FAIL: ${variant} was expected to mount successfully" >&2
        return 1
    fi
}

expect_failure() {
    local variant="$1"
    shift
    if "$@"; then
        echo "FAIL: ${variant} was expected to reproduce the assertion" >&2
        return 1
    fi
    echo "EXPECTED FAILURE: ${variant}"
}

require_docker
start_s3proxy
create_implicit_prefix
assert_implicit_prefix

expect_success 'v1.95' run_variant 'v1.95' 'v1.95'
expect_failure 'v1.97' run_variant 'v1.97' 'v1.97'
expect_success 'merged empty-key fix (0b95042)' run_variant \
    '0b95042f65e86db42653ff633b20db7ac5fd89ed' \
    '0b95042f65e86db42653ff633b20db7ac5fd89ed'

echo 'All expected outcomes observed.'
```

PR #2968, opened 2026-08-26
Upgrade to S3Proxy 4.1.0 and enable the SSE tests

gaul:
S3Proxy 4.1.0 supports server-side encryption, so run the previously disabled use_sse and use_sse=custom flag configurations.


ggtakec (approved):

ggtakec:
@gaul Thanks, this helps a lot with testing SSE.

PR #2970, opened 2026-09-08
Update ChangeLog for 1.98

gaul:


gaul:
We haven't run a release sine December -- suggest we just cut from master with this PR.

gaul:
CI failure is spurious -- see #2971.

ggtakec:
@gaul 
Should we release v1.98 soon?
I think we could also wait to merge the following four PRs before the release.

- #2938 is almost ready to merge with just a minor fix.
- #2929, #2960, and #2961  
we could merge them first and discuss #2964 later.  
(Note: #2964 still has issues on macOS and remains in Draft, so it will take more time.)

ggtakec:
Since the macOS issue in #2964 cannot be addressed immediately, I have merged everything except #2960 (which is waiting for a conflict resolution).

PR #2975, opened 2026-09-23
Refactored test_chmod_listed_directory_keeps_metadata

ggtakec:
n/a

### Details
Fixed flaky test failures caused by the HEADER response assertion.
The raw response string was being used directly, so it needed to strip CRLF or CR(\r) before validation.


test/integration-test-main.sh
```diff
@@ -577,29 +577,48 @@ function test_chmod_listed_directory_keeps_metadata {
         --header "Content-Type: application/x-directory" \
         --header "x-amz-meta-mode: 16877" \
         --header "x-amz-meta-foreign: keepme" < /dev/null
-    local HEADERS; HEADERS=$(s3_head "${TEST_BUCKET_1}/${OBJECT_NAME}")
-    echo "${HEADERS}" | grep -qi "x-amz-meta-foreign: keepme"
```
gaul: I don't understand why the change helps.  Either way the `grep` should match.  What am I missing?

If some special handling is required we should add a function that does this.

